PR #2, opened 2020-06-16
173108170 activity home page

mklewandowski:
Add LARA activity intro/home page based on activity JSON.  This PR includes:
- add SVG icons used in intro page
- add mock activity data used to simulate displaying activities 
- add structure and logo images for project types (logos are used in intro header)
- add React components used to display intro page content (headers, main content, page nav, etc.)
- add basic page navigation with stub content shown for activity pages


src/components/activity-header/header.tsx
```diff
@@ -0,0 +1,40 @@
...
+import './header.scss';
+
+interface IProps {
+  projectId: number | null;
+}
+
+export class Header extends React.PureComponent <IProps> {
+  render() {
+    const projectType = ProjectTypes.find(pt => pt.id === this.props.projectId);
+    const logo = projectType !== undefined ? projectType.logo : undefined;
```
kswenson, line 14:
```suggestion
    const logo = projectType?.logo;
```

src/components/activity-header/header.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const projectURL = projectType !== undefined ? projectType.url : "";
```
kswenson, line 15:
```suggestion
    const projectURL = projectType?.url || "";
```

src/components/activity-header/header.tsx
mklewandowski: Fixed.

src/components/activity-header/header.tsx
mklewandowski: Fixed.

mklewandowski:
@kswenson I'm not opposed to either of those general ideas.  
- A few of the components should probably be converted from class to functional components just to simplify them.  
- I modeled my component unit tests after an example in the starter projects.  If React testing library is what we want to go with, maybe we should consider removing the enzyme-based unit test from the starter projects and using React testing library (testing-library/react-hooks is included in the starter projects, but testing-library/react is not)?  

However, in hopes of getting this PR in, I would also advocate doing the above conversions in smaller ongoing PRs when time permits.

PR #8, opened 2020-07-02
Add basic template for theme colors

mklewandowski:
This PR adds CSS vars which can be used to store and access theme colors and utility functions for setting the theme colors (presumably from an actual activity JSON file).  It also includes a ThemeButtons component that can be added for demoing (but should not be used in production and can be removed once we integrate the color themes into the actual activities).

There were a handful of potential approaches for the color themes.  CSS vars seemed straightforward and easy to integrate into our existing CSS files.  I explored some other mechanisms, and I'm open to alternative ideas if we don't think this approach is adequate of if it is too clunky.

### DIFF
--- a/src/components/theme-buttons.tsx
+++ b/src/components/theme-buttons.tsx
@@ -13,7 +13,7 @@ export const ThemeButtons: React.FC = () => {
   const handleOrangeTheme = () => {
     setThemeColors("#ff8415", "#0592af");
   };
-  const handleCbioTheme = () => {
+  const handleCBioTheme = () => {
     setThemeColors("#008fd7", "#f15a24");
   };
   const handleWatersTheme = () => {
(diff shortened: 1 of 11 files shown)


src/utilities/theme-utils.ts
```diff
@@ -0,0 +1,28 @@
+// example from:
+// https://stackoverflow.com/questions/5560248/programmatically-lighten-or-darken-a-hex-color-or-rgb-and-blend-colors
```
kswenson, line 2:
```suggestion
// https://stackoverflow.com/a/62640342
```
Share link to actual answer that appears to be the source of the code in question. It's a long thread...

mklewandowski: Fixed.

src/components/theme-buttons.tsx
```diff
@@ -0,0 +1,34 @@
...
+import "./theme-buttons.scss";
+
+export const ThemeButtons: React.FC = () => {
+  const handleTealTheme = () => {
+    setThemeColors("#0592af", "#ff8415");
+  };
+  const handleOrangeTheme = () => {
+    setThemeColors("#ff8415", "#0592af");
+  };
+  const handleCbioTheme = () => {
```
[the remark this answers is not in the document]
mklewandowski: Fixed.

PR #12, opened 2020-07-10
Add font theme colors and getContrastYIQColor function 

mklewandowski:
This PR adds primary and secondary font theme colors and a function, getContrastYIQColor, which will compute a light or dark font color based on a background color (or can be used generally for computing a light/dark contrast color from a base color).  Note, failed Travis test is fixed in a separate PR (https://github.com/concord-consortium/activity-player/pull/11).

Contrast color example from this article:
https://24ways.org/2010/calculating-color-contrast

image

image

### DIFF
--- a/src/utilities/theme-utils.ts
+++ b/src/utilities/theme-utils.ts
@@ -1,10 +1,10 @@
 // example from:
 // https://stackoverflow.com/a/62640342
-export const colorShade = (col: string, amt: number) => {
-  col = col.replace(/^#/, "");
-  if (col.length === 3) col = col[0] + col[0] + col[1] + col[1] + col[2] + col[2];
+export const colorShade = (color: string, amt: number) => {
+  color = color.replace(/^#/, "");
+  if (color.length === 3) color = color[0] + color[0] + color[1] + color[1] + color[2] + color[2];
 
-  let [r, g, b]: any = col.match(/.{2}/g);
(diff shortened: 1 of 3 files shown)


src/utilities/theme-utils.ts
```diff
@@ -22,7 +22,21 @@ export const setThemeColors = (primary: string, secondary: string) => {
...
+// example from:
+// https://24ways.org/2010/calculating-color-contrast
+const getContrastYIQColor = (color: string) => {
+  color = color.replace(/^#/, "");
+  if (color.length === 3) color = color[0] + color[0] + color[1] + color[1] + color[2] + color[2];
+  const r = parseInt(color.substr(0,2),16);
+  const g = parseInt(color.substr(2,2),16);
+  const b = parseInt(color.substr(4,2),16);
+  const yiq = ((r * 299) + (g * 587) + (b * 114)) / 1000;
+  return (yiq >= 128) ? "#222222" : "#eeeeee";
```
kswenson, line 41: I'd be inclined to use named constants for these rather than magic numbers. Ideally, they'd be shared with the SCSS, but I'd settle for named constants in the short term.

mklewandowski: Good call.  I added them to the .scss file and imported them from there using the same method you came across for use in CLUE (and I think Chris used in CBio).

PR #13, opened 2020-07-13
Load activities using url parameters

sfentress:
This adds url parameters to allow loading either a sampler activity, or an exported activity from LARA.

* `[activity_player_url]/`: Load the default sample activity, "sample-activity-multiple-layout-types". (We may want to eventually remove this default, but this way going to the url with no parameters continues to work as before.)
* `[activity_player_url]/?activity=sample-activity-cbio`: Load "sample-activity-cbio"
* `[activity_player_url]/?activity=1&baseUrl=http%3A%2F%2Fapp.lara.docker`: Loads activity "1" from http://app.lara.docker

The last option depends on PR https://github.com/concord-consortium/lara/pull/590.

To test the LARA connection:

1. Check out the LARA PR above
2. Modify `ACTIVITY_PLAYER_URL` in `docker-compose.yml` to `http://localhost:8080`, or the appropriate url (note: for devs this should probably be part of a separate docker-compose file brought in using .env, will update when I see comments on other PR)
3. Run `docker compose up` to load the changes
4. Click on the Preview in Activity Player link

Note that LARA will show preview links for individual pages as well, but those don't work yet.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -21,7 +21,6 @@ export const getActivityDefinition = (activity: string, baseUrl?: string | null)
   });
 };
 
-// http://app.lara.docker/activities/5/pages/10/export
 const getActivityDefinitionFromLara = (activity: string, baseUrl: string): Promise<ActivityDefinition> => {
   return new Promise((resolve, reject) => {
     const exportUrl = `${baseUrl.replace(/\/$/, "")}/api/v1/activities/${activity}.json`;
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,0 +1,16 @@
+import sampleActivities from "./data";
+
+export type ActivityDefinition = any;
+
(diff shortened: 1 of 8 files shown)


src/api.ts
```diff
@@ -0,0 +1,43 @@
...
+      if (baseUrl.match(urlRegex)) {
+        getActivityDefinitionFromLara(activity, baseUrl).then(resolve);
+      } else {
+        reject(`${baseUrl} must be a valid url, e.g. baseUrl=https%3A%2F%2Fauthoring.concord.org`);
+      }
+    }
+  });
+};
+
+// http://app.lara.docker/activities/5/pages/10/export
```
kswenson, line 24: Is this an example URL? A little context would be useful.

sfentress: ...nope, that's a comment to myself while I was working... Begone!

PR #14, opened 2020-07-13
173594834 collapse columns

mklewandowski:
Add collapsible column feature for question column on each activity page.  Column can be shown or collapsed using a header button.
collapse


mklewandowski:
@eireland I'll merge this, and if you need help merging or sorting out code clashes, let me know.  I could even do a test of rebasing your branch onto master once this goes in and see if I have issues.

PR #17, opened 2020-07-16
173359665 managed interactives

mklewandowski:
This PR adds the following changes in support of adding managed interactives to each internal activity page:
- Add managed interactives as iframes with authored state passed in to display content such as multiple choice, open response, etc.
- Convert page introduction to new header block format that allows for text boxes and managed/library interactives
- Restructure question numbering to take into consideration the embeddable type and content in the header block
- Restructure sample JSON to use managed/library interactives and new header blocks
- Add text block header

The following stories are addressed in this PR:
173359665 Managed Library Interactive Questions Display
173816551 Question Numbering
173831321 Update Header Block (FKA Page Introduction)


src/components/activity-page/activity-page-content.tsx
```diff
@@ -48,14 +47,21 @@ export class ActivityPageContent extends React.PureComponent <IProps, IState> {
     const vertical = page.layout === PageLayouts.FullWidth;
     const primaryFirst = page.layout === PageLayouts.FullWidth || page.layout === PageLayouts.FortySixty;
 
+    const introEmbeddables = page.embeddables.filter((e: any) => e.section === EmbeddableSections.Introduction);
     const primaryEmbeddables = page.embeddables.filter((e: any) => e.section === EmbeddableSections.Interactive);
     const secondaryEmbeddables = page.embeddables.filter((e: any) => !e.section);
```
sfentress, line 52: I'm guessing this is correct, but just wondering if it would be safer to filter all those whose section is not `Introduction` or `Interactive`, in case we ever start naming secondary sections.

src/components/activity-page/iframe-interactive/iframe-interactive.tsx
```diff
@@ -9,11 +9,12 @@ interface IProps {
...
-    const { embeddable, questionNumber } =this.props;
+    const { embeddable, questionNumber } = this.props;
+    const questionName = embeddable.name ? `: ${embeddable.name}` : "";
     return(
-      <div className="question" data-cy="iframe-interactive-question">
-        <div className="header">Question #{questionNumber}</div>
-        <iframe className="content" src={embeddable.url}></iframe>
+      <div className="iframe-interactive" data-cy="iframe-interactive-question">
+        { questionNumber && <div className="iframe-interactive-header">Question #{questionNumber}{questionName}</div> }
+        <iframe className={`iframe-interactive-content ${questionNumber ? "vertical-offset" : ""}`} src={embeddable.url}></iframe>
```
sfentress, line 17: Don't plain iframe interactives (MwInteractives) also support authored and student state? Couldn't we have `IframeRuntime` here? Or is that just not in the scope of this PR?

src/components/activity-page/activity-page-content.tsx
mklewandowski: Ah, that's a good idea.  I'm not entirely comfortable with the section being null to represent that an embeddable goes in the info/assessment section.  Your idea seems much safer.

src/components/activity-page/iframe-interactive/iframe-interactive.tsx
mklewandowski: Scott mentioned this as well.  I wasn't really thinking about it in this PR, but it does make sense.  I think I will try to make that change and use `IFrameRuntime`.

PR #23, opened 2020-07-24
Improve pinning and sizing of interactives, update sample JSON

mklewandowski:
This PR makes improvements to the pinning and sizing of the interactives (some of the sizing still needs improvements, but this should get us far enough to work with and test pinning, scrolling, and interactive content).  Changes include:
- add new examples pages to see pinned content when scrolling	
- reset scroll when navigating to a new page
- get default aspect ratio of interactive
- better handling of container size using useResizeObserver React hook
- add supportedFeatures listener to get potential interactive dimensions
- better handling of interactive size based on height or aspect ratio from iFrame, container size, and embeddable settings

This PR has changes that address the following stories:
173594992 Interactive Pinning
173913078 Managed/Library Interactive Sizing


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -74,15 +83,19 @@ export const IframeRuntime: React.FC<IProps> =
...
-  }, [initialHeight]);
+  const heightFromSupportedFeatures = ARFromSupportedFeatures && containerWidth ? containerWidth / ARFromSupportedFeatures : 0;
+  // There are several options for specifying the iframe height. Check if we have height specified by interactive (from IframePhone
+  // "height" listener), height based on aspect ratio specified by interactive (from IframePhone "supportedFeatures" listener),
+  // or height from container dimensions and embeddable specifications.
+  const height = heightFromInteractive
+                 ? heightFromInteractive
+                 : heightFromSupportedFeatures
+                    ? heightFromSupportedFeatures
+                    : proposedHeight ? proposedHeight : kDefaultHeight;
```
kswenson, line 94:
```suggestion
  const height = heightFromInteractive || heightFromSupportedFeatures || proposedHeight || kDefaultHeight;
```

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -1,40 +1,59 @@
...
-    }, [divRef]);
-
     const { embeddable, questionNumber } = props;
     const questionName = embeddable.name ? `: ${embeddable.name}` : "";
-    const url = embeddable.url
-                ? embeddable.url
-                : (embeddable.library_interactive.data.base_url ? embeddable.library_interactive.data.base_url : "");
+    // in older iframe interactive embeddables, we get url, native_width, native_height, etc. directly off
+    // of the embeddable object. On newer managed/library interactives, this data is in library_interactive.data.
+    const embeddableData = embeddable.library_interactive?.data ? embeddable.library_interactive.data : embeddable;
```
kswenson, line 22:
```suggestion
    const embeddableData = embeddable.library_interactive?.data || embeddable;
```

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -1,40 +1,59 @@
...
+    const url = embeddableData.base_url ? embeddableData.base_url : embeddableData.url ? embeddableData.url : "";
```
kswenson, line 23:
```suggestion
    const url = embeddableData.base_url || embeddableData.url || "";
```

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -1,40 +1,59 @@
...
+    // TODO: handle different aspect ration methods
+    // const aspectRatioMethod = data.aspect_ratio_method ? data.aspect_ratio_method : "";
+    const nativeHeight = embeddableData.native_height ? embeddableData.native_height : 0;
```
kswenson, line 26:
```suggestion
    const nativeHeight = embeddableData.native_height || 0;
```

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -1,40 +1,59 @@
...
+    const nativeWidth = embeddableData.native_width ? embeddableData.native_width : 0;
```
kswenson, line 27:
```suggestion
    const nativeWidth = embeddableData.native_width || 0;
```

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -74,15 +83,19 @@ export const IframeRuntime: React.FC<IProps> =
     };
   }, [url, authoredState, report]);
 
-  useEffect(() => {
-    if (initialHeight) {
-      setIframeHeight(initialHeight);
-    }
-  }, [initialHeight]);
+  const heightFromSupportedFeatures = ARFromSupportedFeatures && containerWidth ? containerWidth / ARFromSupportedFeatures : 0;
```
scytacki, line 86: No need to fix this now, but the other behavior of heightFromSupportedFeatures is that it should be limited to the available visible area. This also means means that the width has to be updated in order to preserve the aspect ratio. 

If the height is sent by the interactive then the width is always `100%` regardless of the aspectRatio either from supported features or the the nativeWidth,nativeHeight from LARA.

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -1,40 +1,59 @@
...
+    const aspectRatio = nativeHeight && nativeWidth ? nativeWidth / nativeHeight : kDefaultAspectRatio;
```
scytacki, line 28: Regarding the TODO above:
- if the method is 'DEFAULT' then the nativeHeight and nativeWidth are ignored and the `4/3` is used. Unless the interactive provides its own aspectRatio in supported features or its own height with the height message.
- if the method is 'MANUAL' then the nativeHeight and nativeWidth are required, and the supported features aspectRatio is ignored. The height message is still allowed though and that will override the nativeHeight and nativeWidth aspect ratio.
- if the method is 'MAX' then the nativeHeight and nativeWidth are ignored. The supported features aspectRatio is ignored. The height message is still allowed.  If the height message is not sent, then the width is '100%' and the height is 100 percent of the available visible area. If there is no header or footer needed then in css this would be `100vh`.

mklewandowski:
@scytacki yes, most of the changes here are in service of the sizing (and some sample updates needed to test the pinning).  It was difficult to confirm if the pinning was working because the sizing often resulted in odd/confusing container heights that affected the scrolling/pinning.  Once I improved the sizing, it appeared that the pinning was already in good shape.  I didn't expect this to be an exhaustive set of fixes to the sizing (especially since I was instructed in grooming to put interactive sizing on the back burner) so I'll add your notes to the PT story about sizing (#173913078), and we can continue this work when time permits.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
mklewandowski: Fixed.

src/components/activity-page/managed-interactive/managed-interactive.tsx
mklewandowski: Fixed.

src/components/activity-page/managed-interactive/managed-interactive.tsx
mklewandowski: Fixed.

src/components/activity-page/managed-interactive/managed-interactive.tsx
mklewandowski: Fixed.

src/components/activity-page/managed-interactive/managed-interactive.tsx
mklewandowski: Fixed.

mklewandowski:
If no one has any final requests, I'm going to rebase and merge this PR.

PR #24, opened 2020-07-24
Add sample data and handling of hidden pages and questions

mklewandowski:
This PR includes changes to hide pages, page sections, and embeddables that are marked as hidden in the activity JSON.  Question numbering, page numbering, and page navigation is updated to reflect hidden content.  An additional sample JSON file is included that contains hidden pages, hidden sections (the interactive box section has content but is hidden) and questions (use `?activity=sample-activity-hidden-content` to access).


mklewandowski:
> This all looks good, AFAICT. It would be nice to see a spec test for this, though. I assume `sample-activity-hidden-content` is a real activity? Then the spec test could be based on what is really hidden in that activity (which is presumably the same, but hard for me to check).

@eireland Any thoughts on testing hidden content?  Would this make more sense as something we build into the Cypress tests?

mklewandowski:
@sfentress I added tests for `activity-utils.ts` that uses 2 of the sample activities (which also revealed a bug that I fixed).  Should be a good starting point for testing functions in that module.

PR #26, opened 2020-07-24
Add warning banner

mklewandowski:
Add in-progress warning banner to top of application.

Demo here:
https://activity-player.concord.org/branch/warning-banner/


mklewandowski:
> This is totally fine if that is what is wanted. I wonder if it will make testing layout a little trickier, since it pushes down all the content, so the screen cuts off at a different point.
> 
> Alternatives would be to absolutely position this so it was laid on top of the content, and/or to have a close button that removed it. But again, if those aren't wanted that's fine.

I'm fine with either approach.  @lbondaryk wrote the original story (which this PR follows) so she might have an opinion about leaving it as a static banner or having it be something that is positioned on top of the content and can be closed.  One issue with placing it over content is that it would hide the header (until closed) which could make people think that the header is missing.  It would look something like this:
image

PR #28, opened 2020-07-27
Adds partner logos for activity footers

eireland:
Adds project/theme specific text for footers
Changed theme names in the sample docs to show a variety of themes


src/utilities/project-utils.ts
```diff
@@ -10,109 +10,257 @@ import pcLogo from "../assets/project-images/pc-logo.png";
...
+import pennStateLogo from "../assets/partner-logos/PennStateLogo.png";
+import argonneLogo from "../assets/partner-logos/argonne.png";
+import millersvilleLogo from "../assets/partner-logos/millersville.png";
+import uicLogo from "../assets/partner-logos/uic.png";
 
 export interface ProjectType {
   id: number;
   name: string;
-  logo: any;
+  projectLogo: any;
```
mklewandowski, line 31: We already have "project" specified in the name of the interface (`ProjectType`).  So instead of `projectLogo`, I would call this `headerLogo` (if it is ONLY used in the header) to be more in line with the new `footerLogo` and to be less redundant.

src/utilities/project-utils.ts
```diff
@@ -10,109 +10,257 @@ import pcLogo from "../assets/project-images/pc-logo.png";
...
   url: string;
+  footer: string;
+  footerLogo: Array<any>;
+  defaultTheme: string;
```
mklewandowski, line 35: Is this theme parameter needed now?  I think moving forward themes will just be a set of colors and the activity player won't use the old LARA themes that are set in authoring.

src/utilities/project-utils.ts
```diff
@@ -10,109 +10,231 @@ import pcLogo from "../assets/project-images/pc-logo.png";
...
+    footerLogo: [ concordLogo, natgeoLogo, ucscLogo, nsfLogo ],
   },
   {
     id: 3,
     name: "RITES",
-    logo: ritesLogo,
+    projectLogo: ritesLogo,
     url: "http://investigate.ritesproject.net/",
+    footer: "",
+    footerLogo:[],
```
mklewandowski, line 62: Super minor complaint, but we should add a space after the colon for consistency: `footerLogo: [],`.

src/utilities/project-utils.ts
```diff
@@ -10,109 +10,231 @@ import pcLogo from "../assets/project-images/pc-logo.png";
...
+    footerLogo:[ concordLogo, nsfLogo ],
+  },
+  {
+    id: 25,
+    name: "CodeR4MATH",
+    projectLogo: undefined,
+    url: "",
+    footer: "<p class=\"footer-txt\" style=\"color: #555;\">Copyright &copy; 2020 <a href=\"https://concord.org/\">The Concord Consortium</a>. All rights reserved. This activity is licensed under a <a href=\"https://creativecommons.org/licenses/by/4.0/\"> Creative Commons Attribution 4.0 License </a>. The software is licensed under <a href=\"https://opensource.org/licenses/BSD-2-Clause\"> Simplified BSD </a>, <a href=\"https://opensource.org/licenses/MIT\">MIT</a> or <a href=\"https://opensource.org/licenses/Apache-2.0\">Apache 2.0</a> licenses. Please provide attribution to the Concord Consortium and the URL <a href=\"https://concord.org\">https://concord.org</a>.</p>",
+    footerLogo:[],
```
mklewandowski, line 238: I didn't see these additional projects when I was making the header.  Where did you find them?  Production only?

eireland: I took from here https://authoring.concord.org/projects
Not sure how many of them are actually being used, but I thought to add them just because they exist.

mklewandowski: Ah, good thinking!  We'll probably need them eventually.

src/components/activity-introduction/footer.tsx
```diff
@@ -1,39 +1,36 @@
...
+        <div className={`inner ${fullWidth ? "full" : ""}`}>
+          <div className={`footerText`}>
+            {renderHTML(footer)}
+          </div>
+          <div className="footer-logo-container">
+            {footerLogos.length > 0 ? footerLogos.map((logo: any, i: number) => 
+              <img key={i} src={logo} className="footer-logo" data-cy="partner-logo" />
+              ) 
+              : ""
+            }
```
mklewandowski, line 30: You should be able to write this simply as:
```
            {footerLogos.map((logo: any, i: number) =>
              <img key={i} src={logo} className="footer-logo" data-cy="partner-logo" />
              )
            }
```

src/components/activity-introduction/footer.tsx
```diff
@@ -1,39 +1,36 @@
...
-  );
+    return (
+      <div className="footer" data-cy="footer">
+        <div className={`inner ${fullWidth ? "full" : ""}`}>
+          <div className={`footerText`}>
+            {renderHTML(footer)}
+          </div>
+          <div className="footer-logo-container">
+            {footerLogos.length > 0 ? footerLogos.map((logo: any, i: number) => 
+              <img key={i} src={logo} className="footer-logo" data-cy="partner-logo" />
```
mklewandowski, line 27: You probably want to add some margins on the footer-logo class so that the logos son't get jammed together.  It looks like this on LARA:
image

mklewandowski: Which also brings up the question of if there should be a background color on the footer to deal with images that have transparency looking odd.  Check out these HAS logos.  The transparency doesn't look great without a background:
https://activity-player.concord.org/branch/update-header-footer/?activity=sample-activity-has

eireland: yea I didn't work on styling because the icons where appearing all over the footer (float left, right, bottom, inline) and of varying sizes(some projects had them bigger) without asking somebody. But after chatting with Leslie, I think I'll just style them to appear on the bottom and all the same height.

mklewandowski: Sounds good.  I would stick a margin around them to so they're not all jammed together,

eireland:
Some notes on footer:
When no project or footer text is specified, default text will be shown.
Because of the logo transparencies, I made the logo section have a white background, but this should probably be fixed when themes are introduced.
Most footer logos are stored as a an array in ProjectTypes, except for NGSS Assessments. The logos for that project were embedded in many different parts of the footer.

PR #29, opened 2020-07-28
Authenticate and fetch data for question interactives

sfentress:
This adds the ability to

1. Launch the Activity Player from the portal with a token, and use it to fetch data from the portal
2. Use the Firebase JWT from the portal to authenticate with Firebase
3. Find answers for the student and load them into the embeddables.

For the moment, the third point is only concerned with question interactives, and loads the state contained in `answer.report_state.interactiveState`. It's not clear to me whether this is where the data will be for all managed interactives, or just the questions interactives. @pjanik perhaps you can clarify.

Also, Piotr, I had a question about associating the answers with the questions. When I load the activity structure, the embeddables have `ref_ids` such as "404-ManagedInteractive", but when I load the answers they have `question_ids` such as "managed_interactive_404". Im unsure if I'm doing something wrong, or if there's something inconsistent about how the JSON is being created in LARA.

@mklewandowski I added the answer state directly to the embeddables in the activity state. Let me know if you think the answers should instead be kept separate and passed down.

It's tricky to test this. To test it from scratch, you need to 

1. Create a LARA activity in authoring.staging.concord.org
2. Assign it to a class and run it as a student
3. Edit the interactive and give it the url `https://activity-player.concord.org/branch/authenticate-and-fetch-data/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F{activity-number}.json&firebase-app=report-service-dev&report-source=authoring.staging.concord.org`, no launch url, and check off the "short-lived authentication token" box
4. Launch it as a student again.

However, you can also just log into learn.staging as "sstudent1/password" and run "Sam's Data Test Learn.Staging (now Activity Runner)" to see the data I previously added loading.


src/components/app.tsx
```diff
@@ -130,4 +139,42 @@ export class App extends React.PureComponent<IProps, IState> {
...
+    newActivityState.pages.forEach((page: any) => {
+      page.embeddables.forEach((embeddable: any) => {
+        const refId = embeddable.embeddable.ref_id;
+        if (questionAnswers[refId]) {
+          embeddable.embeddable.interactiveState = questionAnswers[refId];
+        }
+      });
+    });
+
+    this.setState({activity: newActivityState});
```
mklewandowski, line 177: This seems fine to add the answers to the activity JSON.  The way things are built, at present, we tend to assume that relevant activity-specific info lives in the JSON that gets passed from component to component.

pjanik:
I'll try to look at it more precisely tomorrow (but I'm only half-day).

> Also, Piotr, I had a question about associating the answers with the questions. When I load the activity structure, the embeddables have ref_ids such as "404-ManagedInteractive", but when I load the answers they have question_ids such as "managed_interactive_404". Im unsure if I'm doing something wrong, or if there's something inconsistent about how the JSON is being created in LARA.

I'm not exactly sure what you mean by `ref_ids`. Where is that coming from? But I'm 99% sure that question_id in the answer should/could match question id. E.g. PortalReport code does that and I don't remember any transformation necessary. But I'm not entirely sure what's source of the Activity JSON.

pjanik:
So, I've looked at it a bit more today. I'm not really following this project, so I don't know planned architecture and so on, but a few general comments (they might not be related directly to this PR actually):

- Activity JSON is downloaded from `https://authoring.staging.concord.org/api/v1/activities/<ID>.json`. I bet it's some new API, right? I've checked it only defines `ref_id` without `id` that matches question ID saved in answer. Note that the teacher report also reconstructs the whole activity structure (in some sense it's similar to this project) and it gets it from a different source - Firestore, the same db that you're already using for answers. The path is: `source/.../resources/activity_<id>`. I think it'd be good to switch to use this one. We'll use the same format for reporting and activity player, and won't have to maintain two separate JSON formats. LARA automatically pushes activity JSON there during publishing. Also, it'd let you automatically update Activity when it gets updated by the author without having to poll LARA API. But I might not be aware of some plans, so @scytacki could confirm if it's a good idea or not.

- I think it might be reasonable to think about state management in this project. It seems that you're downloading activity JSON and then updating the answer directly in it. This seems okay now, but later when you save new user answers, each time an answer is updated, the whole activity JSON will have to be reconstructed. It means that the whole component tree will be re-rendered. Since it's SPA and activity can have lots of pages with lots of questions, it might get problematic (slow). The teacher report already had (or still has?) this problem. I don't have a ready solution here, but just warning. Probably extracting answers to some separate structure could help. This also might get easier (or more difficult :wink:) because of iframes - we can't treat iframes as "controlled inputs" (React terminology) due to async way of communication and limitations of interactive-api (that accepts new interactive state only on init). So, they're more like "uncontrolled inputs" with the initial state only - this might actually simplify these performance issues, but keeping code clean might be more challenging. ;) 

- Not really related to this PR, but I'd put some effort into typing `ActivityDefinition` interface, especially now when it gets extended dynamically. 

- Again only half-related to this PR, but lara-api and portal-api code seem familiar. I guess I've seen it in quite a few projects already. It'd be nice to extract it one day and create some npm library.

sfentress:
@pjanik This is all great feedback.

> [Switching to downloading activities from Firestore]

This makes sense. I think we added the API so that we could immediately create the activity structure without needing to first authenticate with the portal to get the Firestore JWT. But I'm interested in @scytacki 's thoughts.

> I think it might be reasonable to think about state management in this project.

I'm thinking that too. I think @mklewandowski wanted to hold off on that for as long as possible, but it's possible we're reaching that time.

> Again only half-related to this PR, but lara-api and portal-api code seem familiar. I guess I've seen it in quite a few projects already. It'd be nice to extract it one day and create some npm library.

Yes, this was the longer-term goal. The discussion was basically "let's copy&paste one last time...", with the idea that it takes three versions to actually generalize something. This version is 90% from the portal report, and 10% from clue, and is currently student-only. I think once we add the other user-types to this version, we could think about spinning it off to a library.

scytacki:
Regarding the activity format.
The current format used by the activity-player is the export format from LARA. So it isn't a new format.

The activity structure in the report-service only includes things the report code needs to know about.  So it doesn't include text boxes, page layouts, themes, projects, some embeddable properties. (and probably more I'm not thinking of).

I think it makes sense to stick with this export format, and have the report-service stick with its stripped down format. But it would be good to unify the ids of the two if we could.  Originally the ids of the export format had no meaning other than being used for references inside of the export itself.  

I made a new story for this where can discuss further. But I think it will be kind of a big story due to migrations. So I think for the time being the activity player needs to deal with the two formats of ids.

mklewandowski:
Two quick comments:
- Yes, early on it didn't seem necessary to force state management other than React props/state, but that was when we had static sample activity JSON.  Piotr's comments about performance are fair ones, and we'll have to keep an eye on this and see if we need to add more sophisticated state management.
- Typing the activity structure at some point also seems like a good idea.  It's relatively known so we could certainly do this.

sfentress:
So it sounds like this could come in now:

1. State management: Both Matt and I think this should come in, but I think that's a PR in its own right, to avoid mixing concerns. And at the same time we should type the activity structure (though if we use something like MST, this will come basically for free)
2. We'll continue getting the activity JSON from the LARA API
3. Question IDs should be unified somewhere, but for now we should support both versions.

Any objections to merging this in?

mklewandowski:
@sfentress I would vote to merge it.

scytacki:
I'd also vote to merge as is.

PR #30, opened 2020-07-29
Add codecov

eireland:
Creating a draft PR to see if codecov will appear on github when PR is done


eireland:
I think the cypress tests are getting included in the report because the text box shows it is covered and it only has cypress test associated with it.

PR #31, opened 2020-07-30
Add type definitions to activities

sfentress:
Mostly just a refactoring. There are very few functional changes, just a couple places where components may not have been handling the possibility of a prop being undefined.

### DIFF
--- a/src/components/activity-page/managed-interactive/managed-interactive.tsx
+++ b/src/components/activity-page/managed-interactive/managed-interactive.tsx
@@ -1,10 +1,19 @@
 import React from "react";
 import { IframeRuntime } from "./iframe-runtime";
 import useResizeObserver from "@react-hook/resize-observer";
+import { Embeddable } from "../../../types";
 
 interface IProps {
-  embeddable: any;
+  embeddable: Embeddable;
   questionNumber?: number;
+}
+
+interface x {
+  foo: string;
+}
+
+interface y {
+  bar: string;
 }
 
 const kDefaultAspectRatio = 4 / 3;
--- a/src/components/activity-page/managed-interactive/managed-interactive.tsx
+++ b/src/components/activity-page/managed-interactive/managed-interactive.tsx
@@ -8,14 +8,6 @@ interface IProps {
   questionNumber?: number;
 }
 
-interface x {
-  foo: string;
-}
-
-interface y {
-  bar: string;
-}
-
 const kDefaultAspectRatio = 4 / 3;
 
 export const ManagedInteractive: React.FC<IProps> = (props) => {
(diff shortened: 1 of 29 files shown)


src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -1,12 +1,21 @@
...
 }
 
+interface x {
+  foo: string;
+}
+
+interface y {
+  bar: string;
+}
+
```
mklewandowski, line 18: Probably meant to come out?

sfentress: Ha, good catch.

That was just me confirming to my own satisfaction that we can't do

```
interface x {
  foo: string;
}

interface y {
  bar: string;
}

const getFooBar = (xOrY: x | y) {
  if (xOrY.foo) {
    ....
  }
}
```

which is really annoying...

sfentress: I mean... that was a test to find out if anyone would really read through the whole commit.

mklewandowski:
@sfentress are we also missing a definition for EmbeddableXhtmlInteractive?

sfentress:
> Are we also missing a definition for EmbeddableXhtmlInteractive?

Is that something beyond

```
export interface Embeddable {
  type: "ManagedInteractive" | "MwInteractive" | "Embeddable::Xhtml";
  content?: string;
}
```

That's already a part of `Embeddable`. I was considering splitting it into

```
interface BaseEmbeddable { ... }

interface ManagedEmbeddable extends BaseEmbeddable { ...}
interface MwEmbeddable extends BaseEmbeddable { ...}
interface XhtmlEmbeddable extends BaseEmbeddable { ...}

type Embeddable = ManagedEmbeddable | MwEmbeddable | XhtmlEmbeddable
```
but it made the code in `embeddable.tsx` much more complicated. I could go back to it, though, if it's clearer.

If you mean something else, then I may just be missing it.

mklewandowski:
I was looking specifically at text-box.tsx which is expecting EmbeddableXhtmlInteractive to be exported from the new types file.

mklewandowski:
Ah, I'm assuming then that you meant for text-box.tsx to use `Embeddable` as the type on the prop.

sfentress:
D'oh, mistaken commit from when I was testing that Embeddable type split.

src/components/activity-page/text-box/text-box.tsx
```diff
@@ -2,9 +2,10 @@ import React from "react";
 import { renderHTML } from "../../../utilities/render-html";
 
 import "./text-box.scss";
+import { EmbeddableXhtmlInteractive } from "../../../types";
 
 interface IProps {
-  embeddable: any;
+  embeddable: EmbeddableXhtmlInteractive;
```
mklewandowski, line 8: Meant to just be `Embeddable` as the type?

PR #39, opened 2020-08-05
Save interactive state

sfentress:
This adds the ability to save back interactive state, either from interactives with existing state or creating new state.

This can be tested by assigning the activity Activity Player Saving to a class and running it.

As part of this work, the [report-service-dev rules](https://console.firebase.google.com/project/report-service-dev/database/firestore/rules) were updated to allow learners to create and update answers. We will need to copy these rules over to report-service-pro if we bring this in.

I have a few questions about some of the metadata that gets sent up, these may need to be addressed before this is brought in. @scytacki perhaps you could take a look.

```
Answers saved to the Firestore database:

{
      ...
      // if we don't have an old id, I create a uuid. Old answers have ids like "open_response_answer_1234". I
      //  can't tell if there's any reason to have a readable id
      id: "...",

      // these are coming from the jwt
      platform_id: "https://learn.staging.concord.org",
      platform_user_id: "67",

      // db source. If we need old data this is set in a url parameter, otherwise we use our own url
      source_key: "activity-player.concord.org" | "authoring.staging.concord.org"

      // unsure where this is coming from. I added "https://" to the source-key, since old answers
      // show "https://authoring.staging.concord.org". Possibly this should be the current url regardless of
      // the source, however.
      tool_id: "https://" + sourceKey,

      // In the JWT claims there is a `platform_user_id`, used above, and a `user_id`, which I use below. This
      // is a url to the user in the portal.
      // However, in old answers this looks like "2399" instead of this url, which is presumably the user number
      // in LARA itself. I don't see this value for this user in any of the information received from the portal
      tool_user_id: "https://learn.staging.concord.org/users/67",

      // this used to be a nice clean url like "https://authoring.staging.concord.org/activities/20648", but is
      // now a great big activity url with parameters 
      resource_url: "https://activity-player.concord.org/branch/save-interactive-state/?activity=https%3A%2F%2Fauthoring.staging.concord .....",

      // unsure where this comes from, or if it is needed before we get to anonymous users
      run_key: "",
}
```

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -14,7 +14,7 @@ interface ILogRequest {
 interface IProps {
   url: string;
   authoredState: any;
-  interactiveState: any;
+  initialInteractiveState: any;
   setInteractiveState: (state: any) => void;
   report?: boolean;
   proposedHeight?: number;
@@ -23,7 +23,7 @@ interface IProps {
 }
 
 export const IframeRuntime: React.FC<IProps> =
-  ({ url, authoredState, interactiveState, setInteractiveState, report, proposedHeight, containerWidth, setNewHint }) => {
+  ({ url, authoredState, initialInteractiveState, setInteractiveState, report, proposedHeight, containerWidth, setNewHint }) => {
   const [ heightFromInteractive, setHeightFromInteractive ] = useState(0);
   const [ ARFromSupportedFeatures, setARFromSupportedFeatures ] = useState(0);
   const iframeRef = useRef<HTMLIFrameElement>(null);
@@ -33,9 +33,9 @@ export const IframeRuntime: React.FC<IProps> =
   // it reloads the iframe each time it's called, it's not a great experience for user when that happens while he is
   // interacting with the iframe (e.g. typing in textarea). And interactiveState is being updated very often,
   // as well as setInteractiveState that is generated during each render of the parent component.
-  const interactiveStateRef = useRef<any>(interactiveState);
+  const interactiveStateRef = useRef<any>(initialInteractiveState);
   const setInteractiveStateRef = useRef<((state: any) => void)>(setInteractiveState);
-  interactiveStateRef.current = interactiveState;
+  interactiveStateRef.current = initialInteractiveState;
   setInteractiveStateRef.current = setInteractiveState;
 
   useEffect(() => {
@@ -78,7 +78,7 @@ export const IframeRuntime: React.FC<IProps> =
         phoneRef.current.disconnect();
       }
     };
-  }, [url, authoredState, report, setNewHint]);
+  }, [url, authoredState, report, initialInteractiveState, setNewHint]);
 
   const heightFromSupportedFeatures = ARFromSupportedFeatures && containerWidth ? containerWidth / ARFromSupportedFeatures : 0;
   // There are several options for specifying the iframe height. Check if we have height specified by interactive (from IframePhone
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -28,15 +28,6 @@ export const IframeRuntime: React.FC<IProps> =
   const [ ARFromSupportedFeatures, setARFromSupportedFeatures ] = useState(0);
   const iframeRef = useRef<HTMLIFrameElement>(null);
   const phoneRef = useRef<IframePhone>();
-  // Why is interativeState and setInteractiveState kept in refs? So it's not necessary to declare these variables as
-  // useEffect's dependencies. Theoretically this useEffect callback is perfectly fine either way, but since
-  // it reloads the iframe each time it's called, it's not a great experience for user when that happens while he is
-  // interacting with the iframe (e.g. typing in textarea). And interactiveState is being updated very often,
-  // as well as setInteractiveState that is generated during each render of the parent component.
-  const interactiveStateRef = useRef<any>(initialInteractiveState);
-  const setInteractiveStateRef = useRef<((state: any) => void)>(setInteractiveState);
-  interactiveStateRef.current = initialInteractiveState;
-  setInteractiveStateRef.current = setInteractiveState;
 
   useEffect(() => {
     const initInteractive = () => {
@@ -45,7 +36,7 @@ export const IframeRuntime: React.FC<IProps> =
         return;
       }
       phone.addListener("interactiveState", (newInteractiveState: any) => {
-        setInteractiveStateRef.current?.(newInteractiveState);
+        setInteractiveState(newInteractiveState);
       });
       phone.addListener("height", (newHeight: number) => {
         setHeightFromInteractive(newHeight);
@@ -61,8 +52,7 @@ export const IframeRuntime: React.FC<IProps> =
       phone.post("initInteractive", {
         mode: report ? "report" : "runtime",
         authoredState,
-        // This is a trick not to depend on interactiveState.
-        interactiveState: interactiveStateRef.current
+        interactiveState: initialInteractiveState
       });
     };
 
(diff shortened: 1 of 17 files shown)


sfentress:
Note, this PR includes #32, which probably fits better here, which also adds some tests.

scytacki:
@sfentress the report-service rules should be updated via a PR here: 
https://github.com/concord-consortium/report-service/blob/master/firestore.rules

scytacki:
Regarding the answer ids:
They don't need to be readable. However they should be shared with the logging system (not implemented in the AP yet). 
That way the logs can be matched up with the answers.
The logging story is here: https://www.pivotaltracker.com/story/show/174201695
If you can add a comment to that story describing the the answer id used in the report-service so whoever implements it knows what to do.

Regarding the tool properties:
I don't think they are currently used. They are LTI concepts. The intention is for the report-service to support more than one tool. In this case making the tool id be the current url (just the domain part) regardless of the source seems like the best approach. That will help us differentiate between old and new answer answers for a converted activity.
You are correct that the tool_user_id is the LARA user id. Since the activity-player is not maintaining its own set of users, I think it is best to just not include the tool_user_id.

Regarding the resource_url:
The long url might cause problems for some of the researcher report system work that @dougmartin is doing. Because this url is sanitized and then used as a folder name in S3. However I think this resource_url needs to match what is in the portal as the url for the resource (external activity).  Otherwise we can't match up these answers with the portal resource.  So I'd leave the long url for now and we'll figure out how to deal with it in the researcher report system. 

Regarding the run_key:
I think you can leave it off of answers when running from the portal. 

I'm going to look at the code now...

src/firebase-db.ts
```diff
@@ -74,7 +110,85 @@ const watchCollection = (path: string, portalData: IPortalData, listener: IDocum
...
+  return unsubscribe;
+};
+
+const notifyListeners = (path: string, value: any) => {
+  listeners[path]?.forEach(listener => listener(value));
+};
+
+// returns value immediately if it exists, or gets the value once on the first update.
+// equivalent to `firebase.once`
+export const getCurrentDBValue = (path: string, listener: DBChangeListener) => {
```
scytacki, line 142: This would probably be better if it returned a promise that resolved to the value.  Then someone who just wants to get the current value doesn't need to create a DBChangeListener. Just a suggestion though, no need to change.

src/utilities/embeddable-utils.ts
```diff
@@ -0,0 +1,89 @@
...
+    } as IExportableMultipleChoiceAnswerMetadata;
+  } else if (interactiveState.answerType === "open_response_answer") {
+    exportableAnswer = {
+      ...exportableAnswerBase,
+      answer: interactiveState.answerText
+    } as IExportableOpenResponseAnswerMetadata;
+  } else {
+    exportableAnswer = {
+      ...exportableAnswerBase,
+      answer: reportStateJSON
```
scytacki, line 52: It would be good to add a comment about how it isn't handling the has_report_url property of interactives. It is correct that this code shouldn't handle it because interactives that use that (mainly CODAP and Sage) would not work in the activity player anyway without additional work. And we are planning to replace their use of the has_report_url with standard interactiveState handling.

src/utilities/embeddable-utils.ts
```diff
@@ -0,0 +1,89 @@
...
+    } as IExportableInteractiveAnswerMetadata;
+  }
+
+  return exportableAnswer;
+};
+
+/**
+ * this is annoying and possibly a bug? Embeddables are coming through with `refId`'s such
```
scytacki, line 60: There was a discussion about this that resulted in this story: 
https://www.pivotaltracker.com/story/show/174065787
It'd be helpful add a link to that here.

sfentress:
PR for the firestore rules created here: https://github.com/concord-consortium/report-service/pull/30

sfentress:
Hi @scytacki,

The most recent work (from "Remove tool_user_id from metadata" onward)

1. Removes `tool_user_id` from metadata
2. Sets `tool_id` to be the activity-player hostname
3. Changes the embeddable data handling such that an embeddable will only be displayed once we have returned initial data from Firestore (if there is any)

For point 3, some details on this:

a. If we are not planning to load data (we don't have a portal token, or, when run-keys are added, we aren't in preview mode) we won't wait for data
b. Data is requested for the entire activity when a student first loads. Once the first request for data is returned, we mark that we have initial data
c. If we try to load an embeddable while this request is still pending, we will display "loading..." until it has returned.
d. We continue watching the data after the initial grab, so if a student happened to modify the data for embeddables on page 2 while on another tab, and then switches to page 2 on the first tab, they will get the most recent data (barring very tight race conditions)
e. Once an embeddable has mounted, it won't react to new data from Firestore

The main untested chunk of code that codecov is complaining about is the path where we do have real state coming from firestore. This will be much easier to test in the run-key branch with offline firestore.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -61,8 +54,7 @@ export const IframeRuntime: React.FC<IProps> =
       phone.post("initInteractive", {
         mode: report ? "report" : "runtime",
         authoredState,
-        // This is a trick not to depend on interactiveState.
-        interactiveState: interactiveStateRef.current
+        interactiveState: initialInteractiveState
```
scytacki, line 57: @pjanik I just wanted to point this modification out to you.
I guess in some way it is turning the iframe-runtime into an uncontrolled component. 

I'm somewhat paranoid with this component, because it seems like a re-usable piece, so whenever it diverges from other implementations I'd want to understand why.

I think the previous implementation (taken from question-interactives) was more like a controlled component. It did maintain its own reference to the interactiveState, but it was updating this reference on each render.

src/utilities/embeddable-utils.ts
scytacki: @sfentress it looks like you missed this request.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
sfentress: I don't think there's a difference in the result. Previously we were disconnecting the `interactiveState` entirely from the parent component, such that if it changed in the props, we wouldn't reload the iframe. Indeed, if anything this was the "uncontrolled" version.

The change is now we have an `initialInteractiveState`, with a "promise" that we won't change it (hence the "initial" in the name), but if we were to change it we would reload the iframe. When it makes sense to update the initial state (e.g. we want to re-render the component because a dialog has changed) we will reload the component, which I think is what we would want. But currently since we don't change it it behaves as it did before.

pjanik: @scytacki, this is a bit tricky. The previous implementation wasn't a real controlled component. In fact it was uncontrolled one. If you think about the logic of this component, not much changed here. `(init)interactiveState` is sent to the iframe only during `initInteractive` callback in **both** implementations. We never sent it again to iframe when it was updated (that would a real controlled component), as we don't have Interactive API for that. So, in the prev implementation, if the parent component updated the `interactiveState` property, iframe-runtime would ignore that change (unless `authoredState` or `url` have been updated too). This was actually fine in question-interactives repo and could be okay here too (unless we want to support some external updates coming from Firestore).

I think I like this new implementation. It's more explicit and clear. `inititialInteractiveState` is really only the initial state at the moment. There is no way to update `interactiveState` by parent and this implementation makes it clear. It adds some work to the parent component, as it needs to ensure not to update `inititialInteractiveState` all the time, but it makes sense.

Once we add a clean way to update interactive state without reloading the iframe (probably new Interactive API call), we could just add a new property for this component called `interactiveState`. And parent component would have a choice whether it wants to use controller or uncontrolled version. If I remember correctly, that's how native inputs work in React - dev can choose whether to use `value` or `defaultValue` (which was called `initialValue` in some old React versions I think).

src/utilities/embeddable-utils.ts
sfentress: I see the comment `// note we don't current support has_report_url`. Did you want something further?

scytacki: Sorry I missed that before. I think it was just out of range from 3 line context provided in the github conversation view.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
scytacki: Thanks for humoring me.  So then it sounds like we could copy this change back to the question-interactives if we want. No need to do it now, but it is good to know.

PR #40, opened 2020-08-06
173233360 udl account info

eireland:
Moves user name to between the logos and the nav bar.
Made the font larger
Added a semi-transparent background
Added a theme with an image background for demoing.

### DIFF
--- a/src/components/activity-header/profile-nav-header.scss
+++ b/src/components/activity-header/profile-nav-header.scss
@@ -4,17 +4,17 @@
   display: flex;
   justify-content: flex-end;
   width: $content-width;
-  color: $cc-charcoal-hint;
+  align-items: center;
 
   &.full {
     width: 100%;
   }
 
   .profile-name {
-    height: 25px;
-    padding: 0 0 5px 0;
+    padding: 7px;
     font-size: 16px;
-    line-height: 20px;
+    background-color: #8282824e;
+    color: var(--theme-primary-text-color);
 
     .userName {
       cursor: pointer;
--- a/src/components/activity-header/profile-nav-header.scss
+++ b/src/components/activity-header/profile-nav-header.scss
@@ -4,18 +4,16 @@
   display: flex;
   justify-content: flex-end;
   width: $content-width;
-  color: white;
-  margin-top: 5px;
+  color: $cc-charcoal-hint;
 
   &.full {
     width: 100%;
   }
 
   .profile-name {
-    height: 20px;
-    padding: 0 5px 0 5px;
-    background-color: #414546;
-    font-size: 13px;
+    height: 25px;
+    padding: 0 0 5px 0;
+    font-size: 16px;
     line-height: 20px;
 
     .userName {
--- a/src/components/theme-buttons.tsx
+++ b/src/components/theme-buttons.tsx
@@ -3,29 +3,33 @@
 // Add this component to enable buttons to toggle color theme.
 import React from "react";
 import { setThemeColors } from "../utilities/theme-utils";
-import { setBackgroundImage } from "../utilities/activity-utils";
+import { setBackground } from "../utilities/activity-utils";
 
 import "./theme-buttons.scss";
 
 export const ThemeButtons: React.FC = () => {
+  const imageUrl  = `url("https://upload.wikimedia.org/wikipedia/commons/thumb/0/0c/Hubble%27s_Wide_View_of_%27Mystic_Mountain%27_in_Infrared.jpg/678px-Hubble%27s_Wide_View_of_%27Mystic_Mountain%27_in_Infrared.jpg")`;
   const handleTealTheme = () => {
     setThemeColors("#0592af", "#ff8415");
+    setBackground();
   };
   const handleOrangeTheme = () => {
     setThemeColors("#ff8415", "#0592af");
+    setBackground();
   };
   const handleCBioTheme = () => {
     setThemeColors("#008fd7", "#f15a24");
+    setBackground();
   };
   const handleWatersTheme = () => {
     setThemeColors("#007c8B", "#e86e2f");
+    setBackground();
   };
   const handleInteractionsTheme = () => {
     setThemeColors("#414546", "#2f70b0");
+    setBackground(imageUrl);
   };
-  const handleImageBackground = () => {
-    setBackgroundImage("https://upload.wikimedia.org/wikipedia/commons/thumb/0/0c/Hubble%27s_Wide_View_of_%27Mystic_Mountain%27_in_Infrared.jpg/678px-Hubble%27s_Wide_View_of_%27Mystic_Mountain%27_in_Infrared.jpg");
-  };
+
   return (
     <div className="theme-buttons" data-cy="theme-buttons">
       <button className="button" onClick={handleTealTheme} data-cy="theme-button-teal">teal</button>
@@ -33,7 +37,6 @@ export const ThemeButtons: React.FC = () => {
       <button className="button" onClick={handleCBioTheme} data-cy="theme-button-cbio">cbio</button>
       <button className="button" onClick={handleWatersTheme} data-cy="theme-button-waters">waters</button>
       <button className="button" onClick={handleInteractionsTheme} data-cy="theme-button-interactions">interactions</button>
-      <button className="button" onClick={handleImageBackground} data-cy="theme-button-image">image</button>
     </div>
   );
 };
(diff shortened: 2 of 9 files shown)


src/components/activity-header/profile-nav-header.scss
```diff
@@ -4,19 +4,17 @@
...
   .profile-name {
-    height: 20px;
-    padding: 0 5px 0 5px;
-    background-color: #414546;
-    font-size: 13px;
-    line-height: 20px;
+    padding: 7px;
+    font-size: 16px;
+    background-color: #8282824e;
+    color: var(--theme-primary-text-color);
```
mklewandowski, line 17: We'll need to be a little careful about setting this to the primary text color since the text color will sometimes be light and sometimes be dark.  Since the background is static, this could cause readability issues in some cases.

src/components/theme-buttons.tsx
```diff
@@ -22,13 +22,17 @@ export const ThemeButtons: React.FC = () => {
   const handleInteractionsTheme = () => {
     setThemeColors("#414546", "#2f70b0");
   };
+  const handleImageTheme = () => {
+    setBackgroundImage("https://upload.wikimedia.org/wikipedia/commons/thumb/0/0c/Hubble%27s_Wide_View_of_%27Mystic_Mountain%27_in_Infrared.jpg/678px-Hubble%27s_Wide_View_of_%27Mystic_Mountain%27_in_Infrared.jpg");
+  };
```
mklewandowski, line 27: Not a big deal since this is just for testing, but technically the themes just refer to the primary and secondary colors that will be set by authors.  So the image background isn't a theme and this function might be clearer if it's named something like `setBackground` or `handleSetBackground` to be more consistent with the other button handlers.

src/utilities/theme-utils.ts
```diff
@@ -21,6 +21,8 @@ export const colorShade = (color: string, amt: number) => {
 };
 
 export const setThemeColors = (primary: string, secondary: string) => {
+  const el = document.querySelector(".app") as HTMLElement;
+  el && el.style.setProperty("background-image",`linear-gradient(to bottom, #fff, rgba(255,255,255,0), rgba(255,255,255,0))`);
```
mklewandowski, line 25: Isn't this already set in the initial CSS?  It doesn't seem to do anything related to the theme colors so I don't think it belongs here (unless it's supposed to use on of the theme colors?).  Maybe this is supposed to come out?

src/utilities/theme-utils.ts
```diff
@@ -31,6 +33,11 @@ export const setThemeColors = (primary: string, secondary: string) => {
   document.documentElement.style.setProperty("--theme-secondary-text-color", getContrastYIQColor(secondary));
 };
 
+export const setBackgroundImage = (imageUrl:string) => {
+  const el = document.querySelector(".app") as HTMLElement;
+  el && el.style.setProperty("background-image",`url("${imageUrl}")`);
+};
+
```
mklewandowski, line 40: One note here.  This function doesn't really belong in the theme-utils module since it isn't related to themes.  I would put it in activity-utils or, if you think we might need more similar content, make an image-utils module.

src/components/theme-buttons.tsx
eireland: changed the function name to handleImageBackground

src/utilities/theme-utils.ts
eireland: The color theme buttons only changes the colors on the top dom level. The gradient is set at the .app level. When I add the background image at the .app level when you click on the image button, it never changes back to the gradient when you click on the color theme buttons. This takes care of putting the gradient back in.

mklewandowski: Ah, so we lose the gradient if a background image is set and need to restore it.  What if an author wants to set a background image and theme colors?  Does this mean that they cannot do both?  It at least seems like this chunk should be in its own function since at some point we'll want to find a path to have both theme colors and a background color (and, at present, this would overwrite the background image which shouldn't be a side effect of calling a function that sets the theme colors).

src/utilities/theme-utils.ts
eireland: Moved it to activity-utils.

src/utilities/theme-utils.ts
mklewandowski: Yeah, perhaps a setGradient function should be added to the activity utils (or this might be an argument for an image-utils module that simply does:
```
const el = document.querySelector(".app") as HTMLElement;
  el && el.style.setProperty("background-image",`linear-gradient(to bottom, #fff, rgba(255,255,255,0), rgba(255,255,255,0))`);
```

mklewandowski: I think we want to keep it clean so that setting the theme colors only sets theme colors.  This would allow us, later on, to set a background image and a theme color (which presumably we will need to do) without causing conflict.

mklewandowski: Then when we use our test buttons, we would call both `setThemeColors` and `setBackgroundGradient` while still being able to call `setBackgroundImage` and `setThemeColors` in the future without the calls clashing.

src/components/activity-header/profile-nav-header.scss
mklewandowski: I still think there is a visibility issue here.  Look at what we get with the Waters theme colors:
image
I would vote that this font color should be static and the background (background-color in the class profile-name) should be darker so it is always legible.  What do you think?

src/utilities/activity-utils.ts
```diff
@@ -92,3 +92,11 @@ export const enableReportButton = (activity: Activity) => {
   const hasCompletionPage = activity.pages.find((page: any) => page.is_completion);
   return !hasCompletionPage && activity.student_report_enabled;
 };
+
+export const setBackground = (background?:string) => {
+  const gradient = "linear-gradient(to bottom, #fff, rgba(255,255,255,0), rgba(255,255,255,0))";
+  const el = document.querySelector(".app") as HTMLElement;
+  el && ( background? el.style.setProperty("background-image",background) 
+                    : el.style.setProperty("background-image",gradient)
```
mklewandowski, line 100: This function name is a bit vague, but is probably fine (maybe naming it `setAppBackground` tells a little more, but it isn't clear that is used to both set a gradient and a background image - I probably would have made two functions since the functionality here is somewhat different).  Maybe name the argument `backgroundImageURL` to at least help make this clear.

I would clean the spacing up here a tad:
```
  el && ( background
            ? el.style.setProperty("background-image", background) 
            : el.style.setProperty("background-image", gradient)
        );
```

eireland: I changed it to setAppBackgroundImage. Am wondering though if I need error checking on url that is being passed in?

src/components/activity-header/profile-nav-header.scss
mklewandowski: After a little tinkering, I vote that the CSS class `.profile-name` should have the following styles:
`color: $light-font-color;`
`background-color: #414546;`

I don't think we can have the semi-transparent background in this case because there is far too much variability in what appears behind it.  I think the best bet is to make the background and color static and remove the transparency.  What do you think?

src/utilities/activity-utils.ts
mklewandowski: It doesn't seem to break if the URL is bad, we show the color gradient instead.  What happened to the image button in the theme buttons?  We've lost the ability to demo the interactions theme because when we press the Interactions button we now show the star field background.  Seems worth keeping that Interactions button behavior and adding the image button (that was there previously, but now is gone when I pull latest).

src/components/activity-header/profile-nav-header.scss
mklewandowski: How does this look?
colors

src/utilities/activity-utils.ts
mklewandowski: I would put that image button back in theme-button.tsx:
```
  const handleInteractionsTheme = () => {
    setThemeColors("#414546", "#2f70b0");
    setAppBackgroundImage();
  };
  const handleSetImage = () => {
    setAppBackgroundImage(imageUrl);
  };
```
And
```
<button className="button" onClick={handleSetImage} data-cy="image-button">image</button>
```

mklewandowski: And I don't think I was clear on the spacing in that function.  I meant something like:
```
export const setAppBackgroundImage = (backgroundImageUrl?: string) => {
  const gradient = "linear-gradient(to bottom, #fff, rgba(255,255,255,0), rgba(255,255,255,0))";
  const el = document.querySelector(".app") as HTMLElement;
  el && ( backgroundImageUrl
          ? el.style.setProperty("background-image", backgroundImageUrl)
          : el.style.setProperty("background-image", gradient)
        );
};
```

PR #41, opened 2020-08-07
173900278 udl button size

eireland:
Makes nav buttons bigger.
Makes menu bigger
Adds page number to page title
Make bottom buttons always visible but disabled as appropriate
Removed font colors styled in line in footer text


src/components/activity-page/activity-page-content.tsx
```diff
@@ -57,14 +57,15 @@ export class ActivityPageContent extends React.PureComponent <IProps, IState> {
     const secondaryIsOnLeft = page.layout === PageLayouts.Responsive || page.layout === PageLayouts.SixtyForty;
     const collapsible = page.toggle_info_assessment && page.layout !== PageLayouts.FullWidth;
     const renderSecondary = this.renderSecondaryEmbeddables(visibleEmbeddables.infoAssessment, questionsBeforeSecondary, page.layout, secondaryIsOnLeft, collapsible);
+    const pageTitle = page.name || ``;
```
mklewandowski, line 60: Might be clearer to use double quotes for empty string.

src/components/activity-page/bottom-buttons.scss
```diff
@@ -9,4 +9,10 @@
     justify-content: center;
   }
 
+  button:disabled {
+    background-color: $cc-teal-light2;
```
mklewandowski, line 13: We probably want to add `cursor: auto;` here so disabled buttons don't look clickable.

mklewandowski: Also this color needs to be derived from the primary theme colors.  Otherwise, when a theme color is set, the other buttons will change color and these will remain teal.  Add the themeButtons URL parameter and change a color to see what I mean.

src/components/activity-page/activity-page-content.tsx
```diff
@@ -57,14 +57,15 @@ export class ActivityPageContent extends React.PureComponent <IProps, IState> {
...
 
     const [first, second] = primaryFirst
                             ? [renderPrimary, renderSecondary]
                             : [renderSecondary, renderPrimary];
 
     return (
       <div className={`page-content ${page.layout === PageLayouts.Responsive ? "full" : ""}`} data-cy="page-content">
-        <div className="name">{page.name}</div>
+        <div className="name">{`Page ` + pageNumber + `: ` + pageTitle}</div>
```
mklewandowski, line 68: I wonder if the completion page should have a page number?  It doesn't at present.  Not sure if it should or not.

PR #43, opened 2020-08-11
Lightbox functionality

dstrawberrygirl:
This PR depends on https://github.com/concord-consortium/question-interactives/pull/51 to test modal support. The url parameter `?activity=sample-question-interactive-complex` can be used, with page 10 being the most interesting example.

The changes in the managed-interactive file seem extreme because the indentation was incorrect throughout the file. In reality, there are only a few additions: React-modal, a Ref for the interactive's state, three methods (`getModalContainer`, `toggleModal`, and the `closeModal` handler that is not yet used and will likely be removed).

The toggle for modal or not was less elegant - the modal popup does not give the option to show the same component in a popped-out state, since the modal contents are never shown until the modal is shown, so an additional iFrame is shown in non-modal view with the embedded version.


sfentress:
I haven't looked closely at this yet, but you may want to change the `baseUrl` for the image question to something other than `http://localhost:8080/image-question` for easier testing and demoing.

dstrawberrygirl:
> I haven't looked closely at this yet, but you may want to change the `baseUrl` for the image question to something other than `http://localhost:8080/image-question` for easier testing and demoing.

Thanks for the reminder! As soon as the PR in the Question Interactives repo is deployed to a working branch I can change these. Still having a lot of issues getting the Cypress tests to pass currently.

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -16,86 +17,122 @@ interface IProps {
...
+  };
+
+  const setNewHint = useCallback((newHint: string) => {
+    setHint(newHint);
+  }, []);
+
+  const getModalContainer = () => {
+    const modalParent = document.getElementById("app");
+    const container = modalParent ? modalParent : {};
+    return container;
```
kswenson, line 87:
```suggestion
    return document.getElementById("app") || {};
```

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -16,86 +17,122 @@ interface IProps {
...
+  };
+
+  const toggleModal = () => {
+    setShowModal(!showModal);
+  };
+
+  const closeModal = (modalProps?: any) => {
```
kswenson, line 94:
```suggestion
  const handleCloseModalRequest = (modalProps?: any) => {
```
`closeModal` sounds like an imperative, particularly next to the imperative `toggleModal()` above.

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -16,86 +17,122 @@ interface IProps {
...
+  };
+
+  const divTarget = React.useRef(null);
+  const divSize: any = useSize(divTarget);
+  const proposedHeight: number = divSize && divSize.width / aspectRatio;
+  const containerWidth: number = divSize && divSize.width;
+
+  const embeddableAuthoredState = () => {
+    const parsedState = embeddable.authored_state ? JSON.parse(embeddable.authored_state) : {};
+    parsedState.modalSupported = true;
```
kswenson, line 66: Suggest a comment here, e.g. "enable modal support for activity player only".

dstrawberrygirl: Also wondering if I could change from using Authored State to just passing a prop that enabled Activity Player modal support to all managed interactives

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -20,10 +20,11 @@ interface IProps {
   proposedHeight?: number;
   containerWidth?: number;
   setNewHint: (newHint: string) => void;
+  toggleModal?: any;
```
mklewandowski, line 23: We could probably type this with something like `toggleModal?: (modalProps: any) => void;` although then the `toggleModal` call below needs protection.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -58,6 +59,9 @@ export const IframeRuntime: React.FC<IProps> =
       phone.addListener("hint", (newHint: any) => {
         setNewHint(newHint.text || "");
       });
+      phone.addListener("showModal", (modalProps: any) => {
+        toggleModal(modalProps);
```
mklewandowski, line 63: Is the `modelProps` eventually going to be needed by the `toggleModal` function?  I see that the `toggleModal` function passed via props doesn't use it, but I'm assuming it's going to be needed at some point.

src/data/sample-question-interactive-complex.json
```diff
@@ -0,0 +1 @@
...:true,"custom_aspect_ratio_method":null,"inherit_native_width":true,"custom_native_width":576,"inherit_native_height":true,"custom_native_height":435,"inherit_click_to_play":true,"custom_click_to_play":false,"inherit_full_window":true,"custom_full_window":false,"inherit_click_to_play_prompt":true,"custom_click_to_play_prompt":null,"inherit_image_url":true,"custom_image_url":null,"library_interactive":{"hash":"a5d53c7ce6f9ec9a143fd06bff9108b42f3ff7ff","data":{"aspect_ratio_method":"DEFAULT","authoring_guidance":"","base_url":"https://models-resources.concord.org/question-interactives/branch/interactive-lightbox/image-question","click_to_play":false,"click_to_play_prompt":null,"description":"Image Question","enable_learner_state":true,"full_window":false,"has_report_url":false,"image_url":null,"name":"Image Question","native_height":435,"native_width":576,"no_snapshots":false,"show_delete_data_button":true,"thumbnail_url":"","customizable":false,"authorable":true}},"type":"ManagedInteractive","ref_id":"23-ManagedInteractive","linked_interactives":[]},"section":"interactive_box"}]}],"plugins":[],"type":"LightweightActivity","export_site":"Lightweight Activities Runtime and Authoring"}
```
mklewandowski, line 1: I've been using https://jsonformatter.org/ to format the single line JSON that we get when exporting an activity into something more readable.  It's helpful when reading the activity structure from the JSON sample files.

sfentress: Also the Fix JSON extension VS Code that you can instal from Preferences -> Extensions.

Then on any file, Shift-Cmd-P, hit Fix JSON and it will nicely format it.

sfentress:
I'm happy to pair-program if the merge is tricky/annoying.

src/data/sample-question-interactive-complex.json
mklewandowski: Nice extension!  Even easier!

dstrawberrygirl:
Thanks @sfentress, yes the merge is both tricky and annoying - partly self-inflicted because I had attempted to fix the broken indentation in managed-interactive when I last worked on it, so everything is marked as a conflict now. Rather than drag you through this too I'll see if I can resolve it first.

dstrawberrygirl:
It's mostly fixed now, but I have to go back and fix the missing property to tell the managed interactive that it can be shown in a lightbox. Setting this back to Draft status until I can re-implement that key feature.

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -96,19 +131,13 @@ export const ManagedInteractive: React.FC<IProps> = (props) => {
...
+        </div>
+      }
+      {!showModal && interactiveIframe}
+      <Modal isOpen={showModal} appElement={getModalContainer()} onRequestClose={handleCloseModalRequest}>
+        {interactiveIframe}
+      </Modal>
+    </div>
+  );
+};
+ManagedInteractive.displayName = "ManagedInteractive";
```
mklewandowski, line 143: Add newline at end of file.

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -21,64 +22,98 @@ interface IProps {
...
-      setHint(newHint);
-    }, []);
-
-    return (
+  };
+
+  const interactiveIframe =
+    <IframeRuntime
+      url={url}
+      authoredState={embeddable.authored_state}
```
mklewandowski, line 106: I think we need to add back in your `embeddableAuthoredState` function:
```
  const embeddableAuthoredState = () => {
    const parsedState = embeddable.authored_state ? JSON.parse(embeddable.authored_state) : {};
    parsedState.modalSupported = true;
    return JSON.stringify(parsedState);
  };
```

And then in the props, we'll use `authoredState={embeddableAuthoredState()}`

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -81,44 +88,72 @@ export const ManagedInteractive: React.FC<IProps> = (props) => {
...
+
+    const toggleModal = useCallback(() => {
+      setShowModal(!showModal);
+    }, [showModal]);
+
+    const handleCloseModalRequest = (modalProps?: any) => {
+      if (modalProps) {
+        handleNewInteractiveState(modalProps.interactiveState);
+      }
+    };
```
pjanik, line 110: @dstrawberrygirl, do you remember what was the intention here? How would we get modalProps with interactiveState here? I'm going to change this function, as currently, it's impossible to close the modal, but I'm double-checking if I'm not breaking something.

mklewandowski: I agree that this is hairy, but it does appear to work in its current form (modal can be opened/closed, unless I'm missing something).  See this example:
https://activity-player.concord.org/branch/master/?preview&activity=sample-question-interactive-complex&page=1

dstrawberrygirl: I can't actually recall the details here... I suspect I may have added the possibility of modalProps as a placeholder based on a previous implementation. Sorry I don't have anything more concrete to add, you should probably go ahead and make sure to test in both activity player and LARA to check for breakages.

pjanik: > I agree that this is hairy, but it does appear to work in its current form (modal can be opened/closed, unless I'm missing something). See this example:

Actually modal can't be closed when you click outside modal borders. This function needs to do that.
I think I figured it out. `modalProps` argument can be removed, as it will be never passed to this function. It's pretty random here, maybe it got lost during some refactoring. It's a function called automatically when you click modal background. So, it should just close the modal and that's all. I just wasn't sure if I'm not breaking something as `handleNewInteractiveState(modalProps.interactiveState);` sounded serious, but it doesn't seem so. ;)

PR #44, opened 2020-08-11
Save anonymous  data using runKeys

sfentress:
This continues from #39 and only includes the commits from "Allow anonymous saving and loading with runKey" onward.

If we load an activity url

1. without a `token=` parameter from the portal
2. without `mode=teacher-edition`
3. without `preview`

we will create a random `runKey` and use this to save and load data from the portal, unauthenticated, using the new firestore rules introduced in https://github.com/concord-consortium/report-service/pull/30.

This also adds some round-trip saving and loading tests, using local offline firestore.

### DIFF
--- a/src/portal-api.ts
+++ b/src/portal-api.ts
@@ -362,6 +362,11 @@ export const fetchPortalData = async (): Promise<IPortalData> => {
   // query parameter.
   const sourceKey = queryValue("report-source") || parseUrl(offeringData.activityUrl.toLowerCase()).hostname;
 
+  // for the tool id we want to distinguish activity-player branches, incase this is ever helpful for
+  // dealing with mis-matched data when we load data in originally saved on another branch.
+  // This is currently unused for the purpose of saving and loading data
+  const toolId = window.location.hostname + window.location.pathname;
+
   const rawPortalData: IPortalData = {
     type: "authenticated",
     offering: offeringData,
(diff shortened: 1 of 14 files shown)


src/types.ts
```diff
@@ -202,3 +210,5 @@ export type IExportableAnswerMetadata =
   IExportableMultipleChoiceAnswerMetadata;
 
 export interface LTIRuntimeAnswerMetadata extends ILTIPartial, IExportalbleAnswerMetadataBase { }
+
+export interface AnonymousRuntimeAnswerMetadata extends IAnonymousMetadataPartial, IExportalbleAnswerMetadataBase { }
```
kswenson, line 214:
```suggestion
export interface AnonymousRuntimeAnswerMetadata extends IAnonymousMetadataPartial, IExportableAnswerMetadataBase { }
```
although if this is imported from elsewhere it may not be fixable here.

src/portal-api.ts
```diff
@@ -362,3 +380,31 @@ export const fetchPortalData = async (): Promise<IPortalData> => {
   };
   return rawPortalData;
 };
+
+// metadata for saving and loading work for anonymous users, not actually loaded from the portal
+export const anonymousPortalData = () => {
+  const runKey = queryValue("runKey") || uuidv4();
+  setQueryValue("runKey", runKey);
```
kswenson, line 387: Given that `setQueryValue()` can potentially have side effects, I'd be inclined to avoid it when not necessary, e.g. something like:
```
  let runKey = queryValue("runKey");
  if (!runKey) {
    runKey = uuidv4();
    setQueryValue("runKey", runKey);
  }
```

sfentress: Makes sense, you're right. (Although if we don't trust `setQueryValue` to not have problematic have side-effects, we ought to worry about those even when we're adding the runKey.)

src/types.ts
sfentress: :+1:

PR #45, opened 2020-08-11
Add user name or "Anonymous" to welcome message

sfentress:
This adds "Welcome Student Name" to the welcome message.

@scytacki In LARA, I see "Welcome {username}", e.g. "Welcome sstudent". I don't see this name in the classInfo pulled from the portal. Instead I see first, last, and full names, e.g. "Sam Student." I opted to use this, but if we need to be using usernames, and I'm missing how to get them, let me know.


scytacki:
I think using the real student name should be fine. I think we didn't use it in LARA because LARA didn't have direct access to it. This might make it a little harder for testers with lots of users with the same full name. I think they can deal.

I'd also note that the test coverage dropped. I'd guess this is because PortalAPI is not mocked, so it makes it harder to test this. You might be able to pull that from the portal-report repo. I guess its approach is to include this mock data in the actual runtime, and then have runtime switches to load this data instead of loading the data from the portal.

I'd guess in this case it'd be better to use cypress's network mocking to return the mock class data without hitting the actual portal.

sfentress:
@mklewandowski @eireland I wonder if you have any thoughts on this (failing) Cypress test that I've been trying to write.

I'd like to start writing tests that test that the app responds correctly when the app loads data, starting with the simple one that when we show the student name when we get it from the portal.

Rather than trying to stub out the whole backend with `cy.route`, which would require stubbing out a whole bunch of routes that get called when we load data, I wanted to just stub out the function `fetchPortalData` that returns the chunk of data after doing all the requests.

However, with the test I have written, the stub function never gets called, instead the original one does.

`fetchPortalData` is a stand-alone function that gets exported, but we should be able to stub it using `import * as PortalApi from "..."; cy.stub(PortalApi, "fetchPortalData", ...)`.

Indeed, I added a second commit which uses a throw-away function which itself imports `fetchPortalData`, and that one correctly calls the mock, and its test passes. There's something about loading the main app that causes the mock to be lost.

Any thoughts? Adding this pattern would make it easy to add a bunch of tests to areas currently un-tested, but I haven't been able to crack it.

PR #46, opened 2020-08-12
Handle empty library interactive

mklewandowski:
Ensure that the `library_interactive` object on a library interactive embeddable actually has content.  When authoring, a user can add a library interactive, but cancel the selection of a library type from the list. This results in an embeddable section being added to the activity page that is marked as a library interactive, but has no library interactive section (which causes the app to crash).


mklewandowski:
> Looks good. Is there a test for this?

Not at present.  Ideally this won't be an issue in the future as you shouldn't be able to author empty interactives.  I'll add a test.

PR #48, opened 2020-08-13
Dashboard header

eireland:
Changes activity header to be similar to portal dashboard.
- Moves activity nav-menu to the header
- Moves username to the header
- Project logo is defaulted to Concord logo if project logo is not specified
Note that Logo area currently has white background so logo is more visible. May need a redesign or have a better color background for the header.
Also, not sure whether we want some space between header and page nav bar.


src/components/activity-header/account-owner.scss
```diff
@@ -0,0 +1,26 @@
+@import "../vars.scss";
+
+  .accountOwner {
```
mklewandowski, line 3: This camel case naming convention came from the CSS modules used in the portal dashboard, but is not the desired naming convention here.  We want to stick with kebab case in the activity player.  So in this case, we would use `account-owner`.

src/components/activity-header/account-owner.scss
```diff
@@ -0,0 +1,26 @@
...
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    margin: 0 12px 0 12px;
+  }
+  
+  .accountOwnerName {
```
mklewandowski, line 10: `account-owner-name`

src/components/activity-header/account-owner.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import "./account-owner.scss";
+
+interface IProps {
+  userName: string;
+}
+export class AccountOwnerDiv extends React.PureComponent <IProps> {
+  render() {
+    return (
+      <div className={`accountOwner`} data-cy="account-owner">
```
mklewandowski, line 12: We don't need the extra layer of syntax around the className - we can just use quotes directly and write this as:
`<div className="account-owner" data-cy="account-owner"`

src/components/activity-header/account-owner.tsx
```diff
@@ -0,0 +1,18 @@
...
+        <AccountOwnerIcon className={`icon`} />
```
mklewandowski, line 13: `className="icon"`

src/components/activity-header/account-owner.tsx
```diff
@@ -0,0 +1,18 @@
...
+        <div className={`accountOwnerName`}>{this.props.userName}</div>
```
mklewandowski, line 14: `className="account-owner-name"`

src/components/activity-header/activity-nav-header.tsx
```diff
@@ -16,17 +15,9 @@ interface IProps {
 
 export class ActivityNavHeader extends React.PureComponent <IProps> {
   render() {
-    const { activityName, activityPages, currentPage, fullWidth, onPageChange, singlePage} = this.props;
```
mklewandowski, line 19: activityName can be taken out of the component props since it is no longer used.

src/components/activity-header/header.scss
```diff
@@ -1,40 +1,77 @@
 @import "../vars.scss";
 
-.header {
-  height: 86px;
+.activityHeader {
```
mklewandowski, line 3: use kebab case

src/components/activity-header/header.scss
```diff
@@ -1,40 +1,77 @@
...
+    margin: auto;
+    background-color: $activity-header-gray;
+    color: white;
 
     &.full {
       width: 100%;
     }
 
-    .left {
+    .headerLeft {
```
mklewandowski, line 22: use kebab case

src/components/activity-header/header.scss
```diff
@@ -1,40 +1,77 @@
...
+        height: 80%;
+        background-color: white;
+        padding: 6px 10px;
+
+      }
     }
 
-    .logo {
-      height: 35px;
+    .headerCenter {
```
mklewandowski, line 45: use kebab case

src/components/activity-header/header.scss
```diff
@@ -1,40 +1,77 @@
...
+      flex-direction: row;
+      flex: 1;
+      justify-content: center;
+      margin: 15px 0px 15px 0px;
+      align-items: center;
+      width: 400px;
+      height: 100%;
+      font-size: 18px;
+  
+      .activityTitle{
```
mklewandowski, line 56: use kebab case

src/components/activity-header/header.scss
```diff
@@ -1,40 +1,77 @@
...
+        margin-right: 10px;
+      }
     }
-    .project-logo {
-      height: 50px;
+
+    .headerRight{
```
mklewandowski, line 61: use kebab case

src/components/activity-header/header.tsx
```diff
@@ -1,29 +1,39 @@
...
-export class Header extends React.PureComponent <IProps> {
+export class Header extends React.PureComponent<IProps> {
   render() {
-    const { fullWidth, projectId } = this.props;
+    const ccLogoLink = "https://concord.org/";
+    const { fullWidth, projectId, userName, singlePage } = this.props;
     const projectType = ProjectTypes.find(pt => pt.id === projectId);
-    const logo = projectType?.headerLogo;
-    const projectURL = projectType?.url || "";
+    const logo = (projectType?.headerLogo) || ccLogo;
```
mklewandowski, line 23: Parenthesis are superfluous.  I'd remove them.

src/components/activity-header/header.tsx
```diff
@@ -1,29 +1,39 @@
...
+    const projectURL = projectType?.url || ccLogoLink;
     const linkClass = projectURL ? "" : "no-link";
     return (
-      <div className="header" data-cy="header">
+      <div className="activityHeader" data-cy="header">
```
mklewandowski, line 27: use kebab case

src/components/activity-header/header.tsx
```diff
@@ -1,29 +1,39 @@
...
         <div className={`inner ${fullWidth ? "full" : ""}`}>
-          <div className={`left ${linkClass}`} onClick={this.handleProjectLogoClick(projectURL)}>
-            { logo && <img src={logo} className="project-logo" data-cy="project-logo" /> }
+          <div className={`headerLeft ${linkClass}`} onClick={this.handleProjectLogoClick(projectURL)}>
```
mklewandowski, line 29: use kebab case

src/components/activity-header/header.tsx
```diff
@@ -1,29 +1,39 @@
...
+            {logo && <img src={logo} className="project-logo" data-cy="project-logo" />}
           </div>
-          <div className="right" onClick={this.handleConcordLogoClick}>
-            <img src={ccLogo} className="logo" />
+          <div className={`headerCenter`}>
```
mklewandowski, line 32: Use simple quotes: `className="headerCenter"`

src/components/activity-header/header.tsx
```diff
@@ -1,29 +1,39 @@
...
+            {!singlePage && this.renderActivityMenu()}
+          </div>
+          <div className={`headerRight`}>
```
mklewandowski, line 35: Use simple quotes and kebab case: `className="header-right">`

src/components/activity-header/header.tsx
```diff
@@ -36,7 +46,19 @@ export class Header extends React.PureComponent <IProps> {
     }
   }
 
-  private handleConcordLogoClick = () => {
-    window.open("https://concord.org/");
+  private renderActivityMenu = () => {
+    const { activityName } = this.props;
+    return (
+      <React.Fragment>
+        <div className={`activityTitle`} data-cy ="activity-title">Activity:</div>
```
mklewandowski, line 53: use simple quotes and kebab case

src/components/custom-select.scss
```diff
@@ -21,12 +21,25 @@
       background-color: $cc-teal-light4;
     }
 
+    &.topHeader {
```
mklewandowski, line 24: use kebab case

src/components/custom-select.scss
```diff
@@ -21,12 +21,25 @@
...
+      width: 300px;
+      background-color: $activity-header-gray;
+      &:hover {
+        background-color: $cc-charcoal-light1;
+      }
+      &.showList, &:active {
```
mklewandowski, line 30: use kebab case

src/components/custom-select.scss
```diff
@@ -87,6 +100,11 @@
     transition-duration: .25s;
     pointer-events: none;
 
+    &.topHeader {
```
mklewandowski, line 103: kebab case

src/components/custom-select.tsx
```diff
@@ -9,6 +9,7 @@ interface IProps {
   items: string[];
   onSelectItem: (value: string) => void;
   HeaderIcon?: SvgIcon;
+  isHeader?: boolean;
```
mklewandowski, line 12: This came up in the dashboard, and I'm still not a huge fan of this pattern of passing a prop that implies that the component needs to know something about a named parent or container component within the overall layout.  I think it leads to things getting confusing and needing renaming if components are moved around or reused (like we're doing in this PR) and it makes it harder to reuse the prop without ambiguity.  So, if a button is supposed to be the color green when displayed in the header, I think it's better to pass in a `color` prop rather than an `isHeader` prop (then we can reuse the `color` prop anywhere that a color change is needed rather than having to introduce a new prop and we can remove the header component without having to rename the prop in the custom-select).  In this specific case, we only currently have ONE instance of the custom-select, and that is in the header.  I would make the default styling of the custom select match what is needed in the header so a prop isn't needed at all.  So remove this prop, remove the "topHeader" styling, and then change the default color and width of the custom-select to match what is needed in the header.  We can worry about color and width changes later with new props IF we end up needing this custom-select component again.

src/components/custom-select.tsx
```diff
@@ -48,9 +49,10 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
     const { items, HeaderIcon, isDisabled } = this.props;
     const currentItem = items.find(i => i === this.state.current);
     const showListClass = this.state.showList ? "show-list" : "";
+    const useHeader = this.props.isHeader ? "topHeader" : "";
```
mklewandowski, line 52: I would remove and make the default custom-select styling match what is needed in the header.

src/components/custom-select.tsx
```diff
@@ -48,9 +49,10 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
...
     const disabled = isDisabled ? "disabled" : "";
     return (
-      <div className={`header ${showListClass} ${disabled}`} onClick={this.handleHeaderClick}>
+      <div className={`header ${useHeader} ${showListClass} ${disabled}`} onClick={this.handleHeaderClick}>
```
mklewandowski, line 55: Again, I would remove `useHeader` and make the default custom-select styling match what is needed in the header.

src/components/custom-select.tsx
```diff
@@ -60,8 +62,9 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
 
   private renderList = () => {
     const { items } = this.props;
+    const useHeader = this.props.isHeader ? "topHeader" : "";
```
mklewandowski, line 65: Remove `useHeader` and make the default custom-select styling match what is needed in the header.

src/components/custom-select.tsx
```diff
@@ -60,8 +62,9 @@ export class CustomSelect extends React.PureComponent<IProps, IState> {
...
     return (
-      <div className={`list ${(this.state.showList ?"show" : "")}`}>
+      <div className={`list ${useHeader} ${(this.state.showList ?"show" : "")}`}>
```
mklewandowski, line 67: Remove `useHeader` and make the default custom-select styling match what is needed in the header.

src/components/activity-header/account-owner.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import AccountOwnerIcon from "../../assets/svg-icons/account-circle-icon.svg";
+
+import "./account-owner.scss";
+
+interface IProps {
+  userName: string;
+}
+export class AccountOwnerDiv extends React.PureComponent <IProps> {
```
mklewandowski, line 9: Not a big deal, but we tend to not use div and span in our component names (it might make it harder to refactor or imply certain layout patterns that might not hold true).  I would probably just call this AccountOwner or AccountOwnerContainer if that feels too short.

src/components/activity-header/header.tsx
```diff
@@ -36,7 +46,19 @@ export class Header extends React.PureComponent <IProps> {
...
+        <CustomSelect
+          items={[activityName]}
+          onSelectItem={(item: any) => console.log(item)}
+          HeaderIcon={AssignmentIcon}
+          isHeader={true}
+          isDisabled={false}
```
mklewandowski, line 59: Should this be disabled for now since it is only ever a single item?

src/components/activity-header/header.tsx
```diff
@@ -36,7 +46,19 @@ export class Header extends React.PureComponent <IProps> {
...
-  private handleConcordLogoClick = () => {
-    window.open("https://concord.org/");
+  private renderActivityMenu = () => {
+    const { activityName } = this.props;
+    return (
+      <React.Fragment>
+        <div className={`activityTitle`} data-cy ="activity-title">Activity:</div>
+        <CustomSelect
+          items={[activityName]}
+          onSelectItem={(item: any) => console.log(item)}
```
mklewandowski, line 56: I would just pass an empty function or let onSelectItem be undefined.  We probably don't want a console.log here.  I just used that in my example so we could ensure it was working.

src/components/custom-select.scss
```diff
@@ -42,7 +55,7 @@
       height: 20px;
       width: 20px;
       margin: 0 5px 0 5px;
-      fill: $cc-teal;
+      fill: var(--theme-primary-color);
```
mklewandowski, line 58: Be cautious when using the theme colors against static backgrounds.  This won't show up with the Interactions theme.  I'm assuming at some point we will get actual theme color specs so we don't need to stress about this now, but something that we should remember to return to when we work on theme colors again.

src/components/custom-select.scss
```diff
@@ -61,7 +74,7 @@
       height: 24px;
       width: 24px;
       margin: 0 5px 0 5px;
-      fill: $cc-teal;
+      fill: var(--theme-primary-color);
```
mklewandowski, line 77: Same as above.  Be cautious when using the theme colors against static backgrounds.  This won't show up with the Interactions theme.  I'm assuming at some point we will get actual theme color specs so we don't need to stress about this now, but something that we should remember to return to when we work on theme colors again.

src/components/custom-select.scss
```diff
@@ -108,14 +126,14 @@
...
+        background-color: $cc-charcoal-light2;
       }
 
       &:active {
-        background-color: $cc-teal-light4;
+        background-color: $cc-charcoal;
       }
       &:active .check {
-        fill: $cc-teal;
+        fill: var(--theme-primary-color);
```
mklewandowski, line 136: Same as above - this won't show up with the Interactions theme.

src/components/custom-select.scss
```diff
@@ -125,7 +143,7 @@
         fill: white;
 
         &.selected {
-          fill: $cc-teal;
+          fill: var(--theme-primary-color);
```
mklewandowski, line 146: Same as above - this won't show up with the Interactions theme.

src/components/custom-select.scss
eireland: Would it be better if we made them a static color for now then? I debated about it which is why the account owner icon is a static color.

mklewandowski: As long as it works with the default colors, we're probably safe for now.  I'm assuming we can tackle all of the theme coloring once it is added to the authoring.

src/components/activity-header/header.scss
```diff
@@ -1,40 +1,77 @@
...
+      align-items: center;
+      flex: 1;
+      justify-content: flex-end;
+
+      .icon {
+        width: 32px;
+        height: 32px;
+        object-fit: contain;
+        fill: var(--theme-primary-color);
+      }
```
mklewandowski, line 73: Is this `icon` class used?  I think it's obsolete.

PR #51, opened 2020-08-19
173233108 udl tab order

eireland:
Allows tabbing between interactable elements.
Pressing on spacebar or enter key has the same action as clicks
Note that currently the sidebar tab is the first element that is focused when an activity page has a sidebar. May need to change that if that is not the behavior we want.


src/components/activity-header/header.scss
```diff
@@ -26,19 +26,19 @@
...
         height: 80%;
         background-color: white;
+        cursor: pointer;
       }
       .project-logo {
         height: 80%;
         background-color: white;
         padding: 6px 10px;
-
+        cursor: pointer;
```
mklewandowski, line 41: Should all project logos have a pointer?  Precipitating change (use `"project_id": 11,` in an activity JSON to see it in use) doesn't have an explicit link.  However, right now we're linking to the CC site.  Is that correct?  So this would be the current rules set:
- if the project specifies a logo, use the project logo
- if the project doesn't specify a logo, use the CC logo
- if the project specifies a URL, link to the URL from the logo
- if the project doesn't specify a logo, link to the CC site (which means a project might HAVE a logo like a "Global EduCorp Limited" log, but if it DOESN'T have a URL specified, the logo will open the CC site when clicked.  Is that right?).

src/components/activity-header/header.tsx
```diff
@@ -27,7 +28,9 @@ export class Header extends React.PureComponent<IProps> {
       <div className="activity-header" data-cy="activity-header">
         <div className={`inner ${fullWidth ? "full" : ""}`}>
           <div className={`header-left ${linkClass}`} onClick={this.handleProjectLogoClick(projectURL)}>
-            {logo && <img src={logo} className="project-logo" data-cy="project-logo" />}
+            {logo && <img src={logo} className="project-logo" alt="Project website" data-cy="project-logo" 
+                      onClick={this.handleProjectLogoClick(projectURL)} onKeyDown={this.handleProjectLogoClick(projectURL)}
+                      tabIndex={0}/>}
```
mklewandowski, line 33: Might be easier to read these long prop lists that flow onto multiple lines by putting each prop on a line:
```
            {logo && <img 
                        src={logo} 
                        className="project-logo" 
                        alt="Project website" 
                        data-cy="project-logo" 
                        onClick={this.handleProjectLogoClick(projectURL)} 
                        onKeyDown={this.handleProjectLogoClick(projectURL)}
                        tabIndex={0}
                      />}
```

src/components/activity-header/header.tsx
```diff
@@ -41,7 +44,7 @@ export class Header extends React.PureComponent<IProps> {
   }
 
   private handleProjectLogoClick = (url: string) => () => {
-    if (url) {
+    if ((accessibilityClick(event)) && url ) {
```
mklewandowski, line 47: `if (accessibilityClick(event) && url) {`

src/components/activity-introduction/activity-page-links.tsx
```diff
@@ -14,18 +15,20 @@ export class ActivityPageLinks extends React.PureComponent <IProps> {
...
           </div>
           ))
         }
-        <button className="button begin" onClick={this.handleButtonClick(1)}>Begin Activity</button>
+        <button className="button begin" onClick={this.handleButtonClick(1)} onKeyDown={this.handleButtonClick(1)}>Begin Activity</button>
       </div>
     );
   }
 
   private handleButtonClick = (page: number) => () => {
```
mklewandowski, line 29: Consider renaming since this function is used by the button and key events.  Maybe `handlePageChange`

src/components/activity-introduction/activity-page-links.tsx
```diff
@@ -14,18 +15,20 @@ export class ActivityPageLinks extends React.PureComponent <IProps> {
...
-    this.props.onPageChange(page);
+    if ( accessibilityClick(event) ) {
```
mklewandowski, line 30: `if (accessibilityClick(event)) {`

src/components/page-sidebar/sidebar-panel.scss
```diff
@@ -1,16 +1,16 @@
 @import "../vars.scss";
 
-.sidebarPanel {
+.sidebar-panel {
```
mklewandowski, line 3: Thanks for catching these naming convention issues!

src/components/page-sidebar/sidebar.tsx
```diff
@@ -25,7 +25,7 @@ export class Sidebar extends React.PureComponent<IProps, IState> {
   render() {
     const { content, handleShowSidebar, index, show, style, title } = this.props;
     return (
-      <div className={`sidebar-container ${show ? "sidebar-container expanded" : ""}`} style={style} data-cy="sidebar">
+      <div className={`sidebar-container ${show ? "expanded " : ""}`} style={style} data-cy="sidebar">
```
mklewandowski, line 28: `<div className={`sidebar-container ${show ? "expanded" : ""}`} style={style} data-cy="sidebar">`
extra space sneaked in.

src/utilities/accessibility-helper.ts
```diff
@@ -0,0 +1,14 @@
...
+  else if (event.type === "keydown") {
+    const code = event.charCode || event.keyCode;
+    if ((code === 32) || (code === 13)) {
+      return true;
+    }
+  }
+  else {
+    return false;
+  }
+}
```
mklewandowski, line 14: Add newline at end of file and consider naming module `accessibility-utils.ts` to match others.

src/components/activity-header/header.scss
eireland: Good catch, I forgot that there may be a project logo but not a link associated with it. I was thinking it was always going to be a project or CC logo.

eireland: Ended up pulling the logo code out into its own component since there was other logic and css I needed to add for tabbing to it. And then maybe we can also use it for the logos in the footer.

src/components/page-sidebar/sidebar-panel.scss
eireland: There's actually a lot more. I'll keep changing them as I come across them.

src/components/activity-header/header.scss
mklewandowski: Yeah, I wasn't sure what exactly was the  "correct" thing to do here.

eireland:
Tabbing to sidebar is now the right before the bottom nav buttons. But there is a weirdness that it goes to the URL first before the bottom buttons. So it goes header->questions->interactive->sidebar->url bar->bottom nav buttons.

PR #53, opened 2020-08-27
174285013 page nav ellipsis

eireland:


src/components/activity-header/nav-pages.tsx
```diff
@@ -36,16 +52,11 @@ export class NavPages extends React.PureComponent <IProps> {
...
-      <button className={`page-button ${currentClass}`} key={`index ${index}`} onClick={this.handleButtonClick(index + 1)}>
-        <div className={`label ${currentClass}`}>{(index + 1).toString()}</div>
-      </button>
-    );
+  private handlePaginate = (page: any) => {
+    if (page.selected >= 0) {
+      this.props.onPageChange(page.selected + 1);
+    } else {
+      this.props.onPageChange(0);
+    }
```
mklewandowski, line 60: I think you can simplify the contents of this function to:
```
    const { onPageChange } = this.props;
    page.selected !== undefined ? onPageChange(page.selected + 1) : onPageChange(0);
```

src/components/activity-header/nav-pages.tsx
```diff
@@ -13,20 +14,35 @@ interface IProps {
...
+          previousLabel={"<"}
+          nextLabel={">"}
+          breakLabel={"..."}
+          breakLinkClassName={"page-button"}
+          pageCount={this.props.pages.length}
+          onPageChange={this.handlePaginate}
+          marginPagesDisplayed={2}
+          pageRangeDisplayed={5}
+          containerClassName={"paginate-container"}
+          pageClassName={""}
```
mklewandowski, line 29: I'm assuming you don't need to pass in an empty string and can remove this prop from the component.

src/components/activity-header/nav-pages.tsx
kswenson: or even:
```
this.props.onPageChange(pageSelected != null ? page.selected + 1 : 0);
```

kswenson:
>Should the break label be a clickable link?

I would think it should just render as an ellipsis so it doesn't even look like a button.

src/components/activity-header/nav-pages.tsx
mklewandowski: One small tweak on Kirk's version:
`this.props.onPageChange(page.selected != null ? page.selected + 1 : 0);`

kswenson: Do what I mean, not what I say... 😉

eireland: 👍

PR #54, opened 2020-08-31
nav pages header ellipsis removed and 10 pages always showing

eireland:


src/components/activity-header/nav-pages.tsx
```diff
@@ -19,12 +19,12 @@ export class NavPages extends React.PureComponent <IProps> {
         <ReactPaginate
           previousLabel={"<"}
           nextLabel={">"}
-          breakLabel={"..."}
-          breakClassName={"break"}
+          breakLabel={""}
+          // breakClassName={"break"}
```
mklewandowski, line 23: We can probably just remove this comment

PR #55, opened 2020-09-09
Set tab title based on activity page

mklewandowski:
This PR updates the document title displayed in the browser tab based on the activity page.  The following rules are used:
- the home page shows the activity name
- subsequent pages show " Page X " followed by either the page name (if it exists) or the activity name (if the page name is not defined).

### DIFF
--- a/src/utilities/activity-utils.ts
+++ b/src/utilities/activity-utils.ts
@@ -111,6 +111,6 @@ export const setDocumentTitle = (activity: Activity | undefined, pageNumber: num
   if (activity) {
     document.title = pageNumber === 0
       ? activity.name
-      : `Page ${pageNumber} ${activity.pages[pageNumber - 1].name ? activity.pages[pageNumber - 1].name : activity.name}`;
+      : `Page ${pageNumber} ${activity.pages[pageNumber - 1].name || activity.name}`;
   }
 };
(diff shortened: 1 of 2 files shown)


src/utilities/activity-utils.ts
```diff
@@ -102,7 +102,15 @@ export const setAppBackgroundImage = (backgroundImageUrl?: string) => {
...
+                    ? el.style.setProperty("background-image", backgroundImageUrl)
                     : el.style.setProperty("background-image", gradient)
         );
 };
+
+export const setDocumentTitle = (activity: Activity | undefined, pageNumber: number) => {
+  if (activity) {
+    document.title = pageNumber === 0
+      ? activity.name
+      : `Page ${pageNumber} ${activity.pages[pageNumber - 1].name ? activity.pages[pageNumber - 1].name : activity.name}`;
```
kswenson, line 114:
```suggestion
      : `Page ${pageNumber} ${activity.pages[pageNumber - 1].name || activity.name}`;
```
DRY

kswenson (approved):
👍 Looks good -- just one DRY suggestion

src/utilities/activity-utils.ts
mklewandowski: Fixed.

PR #58, opened 2020-09-22
Make text on page selectable.

emcelroy:
Related PT story: https://www.pivotaltracker.com/story/show/174867887

[#174867887]

### DIFF
--- a/src/index.sass
+++ b/src/index.sass
@@ -4,4 +4,3 @@ html, body
 
 body
   font-family: helvetica, 'Lato', sans-serif
-  user-select: none


emcelroy:
@kswenson, @mklewandowski Fixing the text select bug was very easy. I'm wondering if this fix is safe though since it seems like the style rule causing the problem was probably deliberately added. It's not clear to me why though. Maybe it really was just a mistake?

mklewandowski:
> @kswenson, @mklewandowski Fixing the text select bug was very easy. I'm wondering if this fix is safe though since it seems like the style rule causing the problem was probably deliberately added. It's not clear to me why though. Maybe it really was just a mistake?

I think this is just a side-effect of using the Concord starter project as the template for the activity player.  In the starter project, selection is turned off:
https://github.com/concord-consortium/starter-projects/blob/master/src/index.sass

I can't speak to why that was the case, but most likely someone made that decision in the initial stages of putting together the starter project based on the needs of other projects that were being worked on at the time (maybe CLUE?).

mklewandowski (approved):
👍 LGTM!  I think this is safe to merge in.  `user-select: none` was there only as a side effect of using the starter project as the template for the activity player.

emcelroy:
> I think this is just a side-effect of using the Concord starter project as the template for the activity player. In the starter project, selection is turned off:
> https://github.com/concord-consortium/starter-projects/blob/master/src/index.sass

Ah, I see. Thanks for the explanation.

mklewandowski:
> > I think this is just a side-effect of using the Concord starter project as the template for the activity player. In the starter project, selection is turned off:
> > https://github.com/concord-consortium/starter-projects/blob/master/src/index.sass
> 
> Ah, I see. Thanks for the explanation.

Interestingly, this was in the first commit to the starter project almost 2 years ago.  We might want to consider why we even have this in the starter project (is something that projects built upon it actually want in the initial CSS?).

kswenson (approved):
👍

PR #60, opened 2020-09-23
Add API support for plugins to send custom messages to interactives

kswenson:
- add proper typing for `LARA` globals
- ActivityPlayer references `LARA` globals via `React` context rather than `window.LARA`
- add `onInteractiveSupportedFeatures` and `sendCustomMessage` to `IEmbeddableRuntimeContext`
- add `sendCustomMessage` to `IEmbeddableContextOptions`
- update plugin initialization to use proper types and support `sendCustomMessage`
- `setSupportedFeatures` and `setSendCustomMessage` callbacks managed by `<Embeddable>` => `<ManagedInteractive>` => `<IframeRuntime>`

[[#173989045]](https://www.pivotaltracker.com/story/show/173989045)

[LARA #631](https://github.com/concord-consortium/lara/pull/631) contains the corresponding LARA changes.

Note that while there is no plugin code using these features yet I'm opening up the PR for review so as to get feedback on the implementation of the API and whether there's anything I missed.

### DIFF
--- a/src/components/activity-page/embeddable.tsx
+++ b/src/components/activity-page/embeddable.tsx
@@ -1,11 +1,15 @@
-import React, { useRef, useEffect, useState }  from "react";
+import React, { useCallback, useContext, useEffect, useRef, useState }  from "react";
 import { TextBox } from "./text-box/text-box";
+import { LaraGlobalContext } from "../lara-global-context";
 import { ManagedInteractive } from "./managed-interactive/managed-interactive";
 import { ActivityLayouts, PageLayouts } from "../../utilities/activity-utils";
 import { EmbeddablePlugin } from "./plugins/embeddable-plugin";
-import { initializePlugin } from "../../utilities/plugin-utils";
+import { initializePlugin, IPartialEmbeddablePluginContext, validateEmbeddablePluginContextForWrappedEmbeddable
+        } from "../../utilities/plugin-utils";
 import { EmbeddableWrapper, IEmbeddablePlugin, IExportableAnswerMetadata } from "../../types";
 import { localAnswerPath, getCurrentDBValue } from "../../firebase-db";
+import { IInteractiveSupportedFeaturesEvent } from "../../lara-plugin/events";
+import { ICustomMessage, ISupportedFeatures } from "@concord-consortium/lara-interactive-api";
 
 import "./embeddable.scss";
 
@@ -17,6 +21,8 @@ interface IProps {
   questionNumber?: number;
   teacherEditionMode?: boolean;
 }
+
+type ISendCustomMessage = (message: ICustomMessage) => void;
 
 export const Embeddable: React.FC<IProps> = (props) => {
   const { activityLayout, embeddableWrapper, linkedPluginEmbeddable, pageLayout, questionNumber, teacherEditionMode } = props;
@@ -35,12 +41,35 @@ export const Embeddable: React.FC<IProps> = (props) => {
 
   const embeddableWrapperDivTarget = useRef<HTMLInputElement>(null);
   const embeddableDivTarget = useRef<HTMLInputElement>(null);
+  const sendCustomMessage = useRef<ISendCustomMessage | undefined>(undefined);
+  const setSendCustomMessage = (sender: ISendCustomMessage) => {
+    sendCustomMessage.current = sender;
+  };
 
+  const LARA = useContext(LaraGlobalContext);
   useEffect(() => {
-    if (embeddableWrapperDivTarget.current && embeddableDivTarget.current && linkedPluginEmbeddable && teacherEditionMode) {
-      initializePlugin(linkedPluginEmbeddable, embeddable, embeddableWrapperDivTarget.current, embeddableDivTarget.current);
+    const pluginContext: IPartialEmbeddablePluginContext = {
+      LARA,
+      embeddable: linkedPluginEmbeddable,
+      embeddableContainer: embeddableWrapperDivTarget.current || undefined,
+      wrappedEmbeddable: embeddable,
+      wrappedEmbeddableContainer: embeddableDivTarget.current || undefined,
+      sendCustomMessage: sendCustomMessage.current
+    };
+    const validPluginContext = validateEmbeddablePluginContextForWrappedEmbeddable(pluginContext);
+    if (validPluginContext && teacherEditionMode) {
+      initializePlugin(validPluginContext);
     }
-  }, [linkedPluginEmbeddable, embeddable, initialInteractiveState, teacherEditionMode]);
+  // NOTE: initialInteractiveState is not used in the effect, but is required for plugin to work
+  }, [LARA, linkedPluginEmbeddable, embeddable, initialInteractiveState, teacherEditionMode]);
+
+  const handleSetSupportedFeatures = useCallback((container: HTMLElement, features: ISupportedFeatures) => {
+    const event: IInteractiveSupportedFeaturesEvent = {
+      container,
+      supportedFeatures: features
+    };
+    LARA?.Events.emitInteractiveSupportedFeatures(event);
+  }, [LARA?.Events]);
 
   // `initialInteractiveState.loaded` will be set to true the moment our initial request for `answers` data
   // comes through, whether or not there is any particular initial state for this interactive.
@@ -68,7 +97,13 @@ export const Embeddable: React.FC<IProps> = (props) => {
 
   let qComponent;
   if (embeddable.type === "MwInteractive" || (embeddable.type === "ManagedInteractive" && embeddable.library_interactive)) {
-    qComponent = <ManagedInteractive embeddable={embeddable} initialInteractiveState={initialInteractiveState.state} questionNumber={questionNumber} initialAnswerMeta={initialInteractiveState.answerMeta} />;
+    qComponent = <ManagedInteractive
+                    embeddable={embeddable}
+                    initialInteractiveState={initialInteractiveState.state}
+                    questionNumber={questionNumber}
+                    initialAnswerMeta={initialInteractiveState.answerMeta}
+                    setSupportedFeatures={handleSetSupportedFeatures}
+                    setSendCustomMessage={setSendCustomMessage} />;
   } else if (embeddable.type === "Embeddable::EmbeddablePlugin" && embeddable.plugin?.component_label === "windowShade") {
     qComponent = teacherEditionMode ? <EmbeddablePlugin embeddable={embeddable} /> : undefined;
   } else if (embeddable.type === "Embeddable::Xhtml") {
--- a/src/components/activity-page/plugins/embeddable-plugin.tsx
+++ b/src/components/activity-page/plugins/embeddable-plugin.tsx
@@ -1,6 +1,8 @@
-import React, { useRef, useEffect } from "react";
+import React, { useContext, useEffect, useRef } from "react";
 import { IEmbeddablePlugin } from "../../../types";
-import { initializePlugin } from "../../../utilities/plugin-utils";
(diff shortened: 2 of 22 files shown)


src/lara-plugin/plugin-api/types.ts
```diff
@@ -115,6 +116,19 @@ export interface IEmbeddableRuntimeContext {
   onInteractiveAvailable: (handler: IInteractiveAvailableEventHandler) => void;
   /** True if the interactive is immediately available */
   interactiveAvailable: boolean;
+  /****************************************************************************
+   Function that subscribes provided handler to event that gets called when the interactive's supported features are
+   are available.
+
+   @param handler Event handler function.
+   ****************************************************************************/
+  onInteractiveSupportedFeatures: (handler: IInteractiveSupportedFeaturesEventHandler) => void;
```
kswenson, line 125: Note that the PT story says:
>This [EmbeddableRuntimeContext](https://github.com/concord-consortium/lara/blob/master/docs/lara-plugin-api/interfaces/iembeddableruntimecontext.md) Has methods for getting the interactive state. It is to this context that a few new methods likely need to be added.
>1. `getSupportedFeatures()` -- like `getInteractiveState()` this should return a promise or null. Null if the embeddable isn't an interactive. The promise should resolve to the supported features of the interactive once LARA knows what they are.
>2. `sendCustomMessage(type:string, content: object)` -- this should send the message to the wrapped interactive.

As I got into it, however, it seems to me that supported features are more like the interactive available message in that they are set once during interactive initialization and don't change again, so there's no need for a getter that can be called multiple times. Instead it seems more natural to treat it like the interactive available message and have a single callback that is called when appropriate.

src/components/app.tsx
```diff
@@ -79,8 +82,8 @@ export class App extends React.PureComponent<IProps, IState> {
       this.setState(newState as IState);
 
       if (teacherEditionMode) {
-        createPluginNamespace();
-        loadPluginScripts(activity);
+        this.LARA = initializeLara();
+        loadPluginScripts(this.LARA, activity);
```
mklewandowski, line 86: One thing that is dawning on me - do we need to be careful about checking when `loadPluginScripts` completes?  Is there a chance that somewhere we might be trying to access a script that isn't loaded?

src/lara-plugin/events/index.spec.ts
```diff
@@ -27,4 +27,17 @@ describe("Events helper", () => {
...
+      const handler = jest.fn();
+      events.onInteractiveSupportedFeatures(handler);
+      const e: IInteractiveSupportedFeaturesEvent = { container: document.createElement("div"), supportedFeatures: {} };
+      events.emitInteractiveSupportedFeatures(e);
+      expect(handler).toHaveBeenNthCalledWith(1, e);
+      events.offInteractiveSupportedFeatures(handler);
+      events.emitInteractiveSupportedFeatures(e);
+      expect(handler).toHaveBeenNthCalledWith(1, e);
+    });
+  });
```
mklewandowski, line 42: As in the previous PR, does @scytacki still want us to wrap all additions/subtractions in the `lara-plugin` folder with comments or at some point is it just going to make the code hard to read?

src/utilities/plugin-utils.ts
```diff
@@ -51,35 +55,69 @@ export const loadPluginScripts = (activity: Activity) => {
...
+  const validated = validateEmbeddablePluginContextForPlugin(context);
+  return validated && wrappedEmbeddable && wrappedEmbeddableContainer ? validated : undefined;
+};
+
+export const initializePlugin = (context: IEmbeddablePluginContext) => {
+  const { LARA, embeddable, embeddableContainer,
+          wrappedEmbeddable, wrappedEmbeddableContainer, sendCustomMessage } = context;
   // TODO: will need to change search as we implement other plugin types
   const plugin = Plugins.find(p => p.type === "TeacherEdition");
+  if (!plugin) return;
```
mklewandowski, line 88: Ah, this ties to my earlier comment about the plugin scripts not being loaded.  We need to keep an eye on cases where we might come into here before scripts are loaded.

mklewandowski (approved):
👍 There's a lot here to take in so I want to make sure that someone else looks at this, but what I'm seeing looks good.  All of the existing plugin behavior still works (from what I can tell), and the addition of the types should make our lives easier moving forward.  I'm still unsure about how we currently believe we should be commenting/documenting changes to the `lara-plugin` folder.  I'll look to @scytacki for his current thinking.

src/lara-plugin/events/index.spec.ts
kswenson: I just did a round of synchronization so that changes in this PR were added to the corresponding LARA PR and vice-versa to the extent possible. All remaining changes (mostly import statements) are so marked.

src/utilities/plugin-utils.ts
kswenson: I agree that it's out of scope for this PR. The answer, however, comes from the fact that `initializePlugin()` is called from `useEffect()` hooks that have a bunch of dependencies, so how many times it gets called will depend on the number of times that those hooks' dependencies change and how that relates to the time it takes for the plugin to load. So there could be bugs related to when `initializePlugin()` gets called relative to when the script loads and there also could be issues related to calling `initializePlugin()` too often (or perhaps not enough, although that seems less likely).

src/lara-plugin/plugin-api/types.ts
scytacki: Supported features should work that way: called once at initialization.
However it has been abused to dynamically change the supported aspect ratio. So in practice an interactive might send the supported features message multiple times.
Since we are adding this new property of supported features, we can require that the first supportedFeatures message includes the available messages. To be safe I think we should also require that the interactive include the available messages property in each supportedFeatures message it sends. 

I'm not sure if this changes how you are implementing this in the plugin-api.

src/components/app.tsx
scytacki: That is a good point. In LARA the scripts are injected on the page by the server. So they are guaranteed to be loaded  before the code after them is called. 

However I think based on the plugin design it should be safe: 🤞 . The plugin should include a call to register itself which it wouldn't be able to do until its own code has been loaded. And before the plugin is registered there shouldn't be anything that is trying to access it.

src/lara-plugin/events/index.spec.ts
scytacki: Thanks Kirk. If you are able to drop the jquery from lara-plugin, then the code will probably diverge quite a bit. At that point it might sense to drop the commenting approach.  Or perhaps if that code can be sufficiently isolated (both in LARA and here), then we could consider keeping them.

If we do make it over that jquery hurdle and there is still a significant amount of shared code, we might be in a good place to estimate what it would take to turn the shared code into a module in the LARA code base that is imported in the Activity Player.  I only say 'estimate' because I'm not convinced it is worth doing.

src/utilities/plugin-utils.ts
scytacki: Yes this does seem like it will cause problems. 

In LARA authoring this is handled here:
https://github.com/concord-consortium/lara/blob/f7566989290dd7100ee764ab93103e72e503eb36/app/views/plugins/_author.haml#L72-L101
A handler is added to the script.onload event, before adding the script element to the page. And initPlugin is called in the handler.

However that is only how it is done during authoring, which only loads one interactive's plugins at a time and reloads the page afterwards.

At runtime it is handled here: 
https://github.com/concord-consortium/lara/blob/f7566989290dd7100ee764ab93103e72e503eb36/app/views/plugins/_show.haml#L2-L7
followed by this
https://github.com/concord-consortium/lara/blob/f7566989290dd7100ee764ab93103e72e503eb36/app/views/plugins/_show.haml#L148-L178
Due to browsers synchronous loading of javascript, that guarantees the initInteractive won't be called until the plugin is loaded.

I think the AP will need to assign unique plugin labels to the plugins referenced by the activity. And then need to resolve a promise when a plugin with that label is registered. Then the code here can wait for that promise to be resolved before calling initPlugin.

scytacki (commented):
I don't know if this should be part of this PR, but I think I found an issue where the code will not handle multiple wrapping plugins.  If it doesn't make sense to fix that now, feel free to ignore.

src/lara-plugin/plugin-api/types.ts
kswenson: >Supported features should work that way: called once at initialization.
However it has been abused to dynamically change the supported aspect ratio. So in practice an interactive might send the supported features message multiple times.

This has been said before, but clearly we should just add a `setAspectRatio()` and possibly a `useAspectRatio()` hook so we can get out of this situation.

>To be safe I think we should also require that the interactive include the available messages property in each supportedFeatures message it sends.

Since most/many interactives won't support any custom messages, it seems like we can still allow an empty custom message specification. The current implementation just ignores it if it's not present.

src/components/app.tsx
```diff
@@ -79,8 +82,8 @@ export class App extends React.PureComponent<IProps, IState> {
       this.setState(newState as IState);
 
       if (teacherEditionMode) {
-        createPluginNamespace();
-        loadPluginScripts(activity);
+        this.LARA = initializeLara();
```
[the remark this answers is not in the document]
kswenson: I don't particularly see this as a pain point at the moment. It's annoying to have to rely on a `LARA` global variable, but I've set it up so that the AP itself doesn't rely on the global but the plugins still can.

src/components/activity-page/embeddable.tsx
```diff
@@ -35,12 +41,35 @@ export const Embeddable: React.FC<IProps> = (props) => {
...
+      wrappedEmbeddableContainer: embeddableDivTarget.current || undefined,
+      sendCustomMessage: sendCustomMessage.current
+    };
+    const validPluginContext = validateEmbeddablePluginContextForWrappedEmbeddable(pluginContext);
+    if (validPluginContext && teacherEditionMode) {
+      initializePlugin(validPluginContext);
     }
-  }, [linkedPluginEmbeddable, embeddable, initialInteractiveState, teacherEditionMode]);
+  // NOTE: initialInteractiveState is not used in the effect, but is required for plugin to work
+  }, [LARA, linkedPluginEmbeddable, embeddable, initialInteractiveState, teacherEditionMode]);
```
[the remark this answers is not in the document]
kswenson: I'm inclined to leave this the way it is for now and plan to clean it up later.

src/components/activity-page/plugins/embeddable-plugin.tsx
```diff
@@ -11,11 +13,18 @@ interface IProps {
...
-      if (divTarget.current && embeddable) {
-        initializePlugin(embeddable, undefined, divTarget.current, undefined);
+      const pluginContext: IPartialEmbeddablePluginContext = {
+        LARA,
+        embeddable,
+        embeddableContainer: divTarget.current || undefined
+      };
+      const validPluginContext = validateEmbeddablePluginContextForPlugin(pluginContext);
+      if (validPluginContext) {
+        initializePlugin(validPluginContext);
```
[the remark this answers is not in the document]
kswenson: Instantiating something that doesn't have a view associated with it is essentially a side effect, which is what hooks were designed to do, so maybe we're talking about a hook of some kind.

src/utilities/plugin-utils.ts
kswenson: It seems clear there's a larger body of work here, which seems better left to a subsequent PR.

kswenson:
The primary purpose of this PR is to introduce the custom messages API. While several important issues have been raised, none of them seem to relate to the custom messages API, so my inclination would be to accept this PR if there are no further objects to the custom messages API itself and write stories for the additional work related to asynchronous JavaScript loading, support for multiple wrapping plugins, etc.

scytacki (approved):
Dealing with my comments in a new PRs is fine with me.

PR #61, opened 2020-09-24
174286142 sidebar plugin

mklewandowski:
I'm putting this in draft.  Not quite ready to open a full PR, but would like someone else to look at this.  

This PR adds teacher edition sidetips to activities.  Teacher edition sidetips are only displayed in teacher-edition mode.  For testing or running use URL params: `?preview&mode=teacher-edition&activity=sample-activity-plugins` to enter teacher edition mode and view an activity that includes sidetips.

Changes to the PR include;
- updates JSON plugin sample activity to include sidetips
- adds `sidebar.ts` from LARA
- adds `ExpandableContainer` container component which contains sidebars (defined on page) and sidetips (embeddable that is displayed outside main page content when in teacher edition mode) 
- adds `EmbeddablePluginSideTip` which is used for creating a container context for the plugin script

Some notes on this PR:
- I'm not happy with the introduction of jQuery and the implementation of the sidebar container using jQuery.  I would prefer that when `addSidebar` is called from a plugin, that we find a way to route the contents into a React component for easier layout, maintenance, and styling and for better integration with the rest of the app and other React components.
- There are some styling nuances that I punted on as it would have required a deeper dive in the jQuery code.
- There are still some potential issues when displaying the sidetips and page sidebar content (not the teacher edition tips) simultaneously.  A user can potentially click around to get both a sidebar and sidetip open at the same time (as in the current LARA).  I would prefer that we handle things like this ONLY after working out a full React implementation of the sidetips.  The work that we would need to do to fix some of these edge cases doesn't seem worth the resources at this point.  Especially since I'm hoping we can eventually rewrite this so the entire front-end is in React.

### DIFF
--- a/src/utilities/activity-utils.ts
+++ b/src/utilities/activity-utils.ts
@@ -46,16 +46,20 @@ export const isEmbeddableSectionHidden = (page: Page, section: string | null) =>
 export const getVisibleEmbeddablesOnPage = (page: Page) => {
   const headerEmbeddables = isEmbeddableSectionHidden(page, EmbeddableSections.Introduction)
     ? []
-    : page.embeddables.filter((e: any) => e.section === EmbeddableSections.Introduction && !e.embeddable.is_hidden && !e.embeddable.embeddable_ref_id && !isEmbeddableSideTip(e));
+    : page.embeddables.filter((e: any) => e.section === EmbeddableSections.Introduction && isVisibleEmbeddable(e));
   const interactiveEmbeddables = isEmbeddableSectionHidden(page, EmbeddableSections.Interactive)
     ? []
-    : page.embeddables.filter((e: any) => e.section === EmbeddableSections.Interactive && !e.embeddable.is_hidden && !e.embeddable.embeddable_ref_id && !isEmbeddableSideTip(e));
+    : page.embeddables.filter((e: any) => e.section === EmbeddableSections.Interactive && isVisibleEmbeddable(e));
   const infoAssessEmbeddables = isEmbeddableSectionHidden(page, null)
     ? []
-    : page.embeddables.filter((e: any) => (e.section !== EmbeddableSections.Interactive && e.section !== EmbeddableSections.Introduction && !e.embeddable.is_hidden && !e.embeddable.embeddable_ref_id && !isEmbeddableSideTip(e)));
+    : page.embeddables.filter((e: any) => (e.section !== EmbeddableSections.Interactive && e.section !== EmbeddableSections.Introduction && isVisibleEmbeddable(e)));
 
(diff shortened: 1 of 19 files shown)


emcelroy (commented):
Generally this is looking good to me. I checked it out locally and ran the related sample activity. All the elements essentially act just like they do in LARA now.

I think doing away with jQuery would be a good idea and it makes sense to wait till that's done before fixing the quirky UI issues that already exist on LARA.

A little thing that's probably out of scope, but I wanted to bring up is our use of classes like `.h5` and `.h4` in the CSS for activities. That's always bothered me, but I've never addressed it anywhere. Those classes seem so unnecessary. I could be wrong, but I'd think it'd be better to apply styles directly to the h5 and h4 elements themselves instead of using classes with names that match the HTML tag. Just thought I'd mention that in case it was something you think we might work on fixing as part of the Activity Player work?

src/utilities/activity-utils.ts
```diff
@@ -45,17 +46,34 @@ export const isEmbeddableSectionHidden = (page: Page, section: string | null) =>
 export const getVisibleEmbeddablesOnPage = (page: Page) => {
   const headerEmbeddables = isEmbeddableSectionHidden(page, EmbeddableSections.Introduction)
     ? []
-    : page.embeddables.filter((e: any) => e.section === EmbeddableSections.Introduction && !e.embeddable.is_hidden && !e.embeddable.embeddable_ref_id);
+    : page.embeddables.filter((e: any) => e.section === EmbeddableSections.Introduction && !e.embeddable.is_hidden && !e.embeddable.embeddable_ref_id && !isEmbeddableSideTip(e));
```
kswenson, line 49:
```suggestion
    : page.embeddables.filter((e: any) => e.section === EmbeddableSections.Introduction && isVisibleEmbeddable(e));
```
Looks like an opportunity to introduce a utility function, e.g.
```
function isVisibleEmbeddable(e) {
  return !e.embeddable.is_hidden && !e.embeddable.embeddable_ref_id && !isEmbeddableSideTip(e);
}
```

kswenson (approved):
👍 Other than the jQuery dependency which you've already raised as an issue, this looks pretty good. Just one suggestion to DRY up the code a bit.

src/utilities/activity-utils.ts
mklewandowski: Good call.  Fixed.

mklewandowski:
I think we all agree that we would like to rip the jQuery out when time permits.  I'm adding @scytacki as reviewer on this as well (in case he wants to take a look).  Otherwise @kswenson do you have a preference if this PR or your plugin PR goes in first?

PR #62, opened 2020-09-29
Display feedback on authentication failure.

emcelroy:
Related PT story: https://www.pivotaltracker.com/story/show/174344573

[#174344573]

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -138,7 +138,7 @@ export class App extends React.PureComponent<IProps, IState> {
         />
         { authError
           ? <AuthError />
-          : this.renderNavAndContent(activity, currentPage, totalPreviousQuestions, fullWidth)
+          : this.renderActivityContent(activity, currentPage, totalPreviousQuestions, fullWidth)
         }
         { (activity.layout === ActivityLayouts.SinglePage || currentPage === 0) &&
           <Footer
@@ -158,7 +158,7 @@ export class App extends React.PureComponent<IProps, IState> {
     );
   }
 
-  private renderNavAndContent = (activity: Activity, currentPage: number, totalPreviousQuestions: number, fullWidth: boolean) => {
+  private renderActivityContent = (activity: Activity, currentPage: number, totalPreviousQuestions: number, fullWidth: boolean) => {
     return (
       <>
         <ActivityNavHeader
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -136,34 +136,9 @@ export class App extends React.PureComponent<IProps, IState> {
           activityName={activity.name}
           singlePage={activity.layout === ActivityLayouts.SinglePage}
         />
-        { authError === ""
-          ? <ActivityNavHeader
-              activityPages={activity.pages}
-              currentPage={currentPage}
-              fullWidth={fullWidth}
-              onPageChange={this.handleChangePage}
-              singlePage={activity.layout === ActivityLayouts.SinglePage}
-            />
-          : ""
-        }
-        { authError === ""
-          ? activity.layout === ActivityLayouts.SinglePage
-            ? this.renderSinglePageContent(activity)
-            : currentPage === 0
-              ? this.renderIntroductionContent(activity)
-              : activity.pages[currentPage - 1].is_completion
-                ? this.renderCompletionContent(activity)
-                : <ActivityPageContent
-                    enableReportButton={currentPage === activity.pages.length && enableReportButton(activity)}
-                    isFirstActivityPage={currentPage === 1}
-                    isLastActivityPage={currentPage === activity.pages.filter((page) => !page.is_hidden).length}
-                    pageNumber={currentPage}
-                    onPageChange={this.handleChangePage}
-                    page={activity.pages.filter((page) => !page.is_hidden)[currentPage - 1]}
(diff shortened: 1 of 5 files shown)


src/components/app.tsx
```diff
@@ -194,6 +213,12 @@ export class App extends React.PureComponent<IProps, IState> {
     );
   }
 
+  private renderAuthError = () => {
+    return (
+      <AuthError />
+    );
+  }
```
mklewandowski, line 220: Since this function is just returning `<AuthError />` we probably don't need the `renderAuthError` render function.  We can just stick `<AuthError />` in where `renderAuthError` is being called and remove the `renderAuthError` function.

src/components/app.tsx
```diff
@@ -122,29 +136,34 @@ export class App extends React.PureComponent<IProps, IState> {
...
-                />
+        { authError === ""
+          ? <ActivityNavHeader
+              activityPages={activity.pages}
+              currentPage={currentPage}
+              fullWidth={fullWidth}
+              onPageChange={this.handleChangePage}
+              singlePage={activity.layout === ActivityLayouts.SinglePage}
+            />
+          : ""
```
mklewandowski, line 147: You could simplify this a little bit and remove the empty string in the condition with:
```
        { authError === "" &&
          <ActivityNavHeader
            activityPages={activity.pages}
            currentPage={currentPage}
            fullWidth={fullWidth}
            onPageChange={this.handleChangePage}
            singlePage={activity.layout === ActivityLayouts.SinglePage}
          />
        }
```

src/components/app.tsx
```diff
@@ -122,29 +136,34 @@ export class App extends React.PureComponent<IProps, IState> {
...
+                    enableReportButton={currentPage === activity.pages.length && enableReportButton(activity)}
+                    isFirstActivityPage={currentPage === 1}
+                    isLastActivityPage={currentPage === activity.pages.filter((page) => !page.is_hidden).length}
+                    pageNumber={currentPage}
+                    onPageChange={this.handleChangePage}
+                    page={activity.pages.filter((page) => !page.is_hidden)[currentPage - 1]}
+                    totalPreviousQuestions={totalPreviousQuestions}
+                    teacherEditionMode={this.state.teacherEditionMode}
+                  />
+          : this.renderAuthError()
```
mklewandowski, line 166: You could even collapse this and the above into a single check for `authError === ""` where you would conditionally render both the `ActivityNavHeader` and the contents or the `<AuthError />`.  You would need to wrap the `ActivityNavHeader` and the contents  inside a React fragment (which could all live in a new function).  Just a thought to reduce the check for the failed authentication to a single code location.

mklewandowski (approved):
👍 This looks good!  I added a couple of suggestions to streamline the code:
- instead of adding the `renderAuthError` function, we can just insert `<AuthError />` directly in the render function when it is needed.
- we can remove one of the ternary operators and the empty string render by just using a conditional render with `&&` OR we could remove the two checks for `authError === ""` and just have a single check with a ternary that renders 1 of 2 things: the `ActivityNavHeader` and the activity contents OR the `AuthError` component (might need to use a React Fragment and a new render function to make this clear)

src/components/app.tsx
kswenson: Or even:
```
        { authError &&
          ...
```

kswenson (approved):
👍 Looks good -- not much to add beyond @mklewandowski's comments.

emcelroy:
@kswenson @mklewandowski Thanks for the suggestions. Do the latest changes incorporate those alright? I cringe a tiny bit at having a function named `renderNavAndContent`. To me it intuitively seems like the two sections should be handled separately, but then it does make sense to lump them together since they're the two elements that get hidden on auth failure... Maybe there's at least a more appropriate name I'm not thinking of?

mklewandowski (approved):
Refactor looks good to me.  I know what you mean about the name, but it is at least clear what the function renders.

src/components/app.tsx
```diff
@@ -122,6 +136,31 @@ export class App extends React.PureComponent<IProps, IState> {
           activityName={activity.name}
           singlePage={activity.layout === ActivityLayouts.SinglePage}
         />
+        { authError === ""
+          ? this.renderNavAndContent(activity, currentPage, totalPreviousQuestions, fullWidth)
+          : <AuthError />
```
kswenson, line 141: Purely a matter of taste, but I would probably write this as:
```typescript
  { authError
      ? <AuthError />
      : this.renderActivityContent(activity, currentPage, totalPreviousQuestions, fullWidth)
  }
```
Rationale:
1. All else being equal I think conditions read better in the affirmative than the negative, i.e. "If there's an error..." rather than "If there's no error/the error is empty...".
2. I share your hesitation about `renderNavAndContent()` as a name and it doesn't seem completely unreasonable to think of the navigation as part of the `ActivityContent`.

At the end of the day these are minor nits about which different opinions are reasonable, so do what makes sense to you.

kswenson (approved):
👍 Looks good -- just one comment which you can accommodate or ignore as you see fit. 😉

src/components/app.tsx
emcelroy: I see, I think that makes sense about using affirmative over negative and I've made that change. Thanks.

I misread the part about changing the function name though before making the commit. I thin that makes sense too and I'll make another commit for that.

PR #63, opened 2020-10-01
Show hi-res image in modal image interactive

mklewandowski:
This PR adds the ability to pass a flag in the authored state of an image question interactive so that the interactive can know if it is being shown in modal or non-modal form.  The image question interactive uses this to determine if it should show the high resolution version of the image (if available).


kswenson (approved):
👍 LGTM

PR #64, opened 2020-10-02
Implement snapshot API

pjanik:
[#173594575]

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "webpack-dev-server": "^3.11.0"
   },
   "dependencies": {
-    "@concord-consortium/lara-interactive-api": "^0.5.0-pre.4",
+    "@concord-consortium/lara-interactive-api": "^0.6.0-pre.1",
     "@react-hook/resize-observer": "^1.1.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -137,7 +137,7 @@
     "webpack-dev-server": "^3.11.0"
   },
   "dependencies": {
-    "@concord-consortium/lara-interactive-api": "^0.5.0-pre.4",
+    "@concord-consortium/lara-interactive-api": "^0.6.0-pre.1",
```
kswenson, line 140: I think we've discussed removing the `pre.x` label from the lara-interactive-api at this point,  but that can wait until next time you/we publish a version.

kswenson (approved):
👍 Looks good -- at some point we should add some better tests of `IframeRuntime` but that doesn't have to be now.

package.json
pjanik: Right. I just wanted to clear my queue of PRs and don't open them with breaking tests. This version is published with `beta` tag. I guess when LARA PRs are merged, I should release 0.6.0 and update version number in master.

PR #65, opened 2020-10-05
174959548 sequences

mklewandowski:
This PR adds sequence handling to the Activity Player.  Changes includes:
- new `Sequence` type and helper functions for loading sequence data
- `app.tsx` state now includes `sequence` and `showSequence` to hold the current sequence (if we have one) and it we are showing the sequence intro page or an activity
- new `SequencePageContent` component used to display the sequence intro page with activity navigation
- header changes for when a sequence is open
- sequence test data which can be accessed with URL param `?sequence=sample-sequence`

Note: `getSequenceDefinitionFromLara` might need additional work.  I'm assuming that we will be able to use a URL to access sequence JSON in the same form that we get when we export a sequence from the authoring tools.  

Screen Shot 2020-10-05 at 11 51 14 AM


cypress/integration/sequence-page.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    it("should load sequence thumbnails", () => {
+      sequencePage.getThumbnails().should("have.length", 5);
+    });
+    it("should navigate to activity when activity thumbnail is selected", () => {
+      activityPage.getHeader().should("contain", "Sequence");
+      sequencePage.getThumbnails().eq(0).click();
+      activityPage.getHeader().should("contain", "Activity");
+    });
+  });
+});
```
kswenson, line 22:
```suggestion
});

```
EOL

kswenson (approved):
👍 Looks good -- just a couple of comments

emcelroy (approved):
Code looks good to me 👍

This was likely discussed elsewhere and I missed or forgot the conversation (sorry if so), but is the sequence title not supposed to appear above the activity title on activity pages? It does in LARA. 

I see that the sequence title does appear in the navbar link that acts as a simple replacement for LARA's clunky drop-down menu. Maybe since it's there, we didn't think it was needed at the top of the page? I don't think it's bad necessarily. I just want to make sure dropping it from the top of the activity pages was done consciously.

mklewandowski:
> Code looks good to me 👍
> 
> This was likely discussed elsewhere and I missed or forgot the conversation (sorry if so), but is the sequence title not supposed to appear above the activity title on activity pages? It does in LARA.
> 
> I see that the sequence title does appear in the navbar link that acts as a simple replacement for LARA's clunky drop-down menu. Maybe since it's there, we didn't think it was needed at the top of the page? I don't think it's bad necessarily. I just want to make sure dropping it from the top of the activity pages was done consciously.

I was hesitant to make too many changes to the header since there was an earlier request, story, and round of work to unify the activity player header with the soecs that Michael designed for the portal report.  The goal with that work was that the activity player and the portal dash would begin to have a unified look and feel.  But that was before sequences.  I was concerned that shoe-horning more content into the header (which as you saw I already had to do to some degree to get the basic functionality working) would break the very deliberate work that was done earlier to unify the header work with the portal dash.  The biggest problem is that we simply don't have specs on the activity player yet so we're playing a bit of a guessing game.  I'm not really sure how some of these pieces are supposed to fit together especially as we try to restyle or redesign the UI.  In the previous work, we were told to collapse the headers into one so now we don't really have space for both the sequence and activity title without rethinking the layout.  Maybe we can get some more feedback at demo/planning today.

emcelroy:
> I was hesitant to make too many changes to the header since there was an earlier request, story, and round of work to unify the activity player header with the soecs that Michael designed for the portal report...

OK, thanks for the explanation. Makes sense to me.

cypress/integration/sequence-page.test.ts
mklewandowski: Fixed.

src/components/app.tsx
```diff
@@ -99,18 +106,43 @@ export class App extends React.PureComponent<IProps, IState> {
...
+          sequence={sequence}
+          onSelectActivity={this.handleSelectActivity}
+        />
+        <Footer
+          fullWidth={true}
+          projectId={sequence.project_id}
+        />
+      </React.Fragment>
+    );
+  }
```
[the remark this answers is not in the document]
mklewandowski: Agreed.  Made a new component.  Just need to add some tests,

PR #66, opened 2020-10-06
Support hostFeatures in initInteractive message

kswenson:
[[#174886713]](https://www.pivotaltracker.com/story/show/174886713)

I don't know what's up with the Codecov failures given that this PR changes almost nothing from a coverage perspective.

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -3,8 +3,9 @@ import React, { useEffect, useRef, useState } from "react";
 import { IframePhone } from "../../../types";
 import iframePhone from "iframe-phone";
 import {
-  ClientMessage, IGetFirebaseJwtRequest, IGetInteractiveSnapshotRequest, IGetInteractiveSnapshotResponse,
-  ILinkedInteractive, ServerMessage, ICustomMessage, ISupportedFeatures
+  ClientMessage, ICustomMessage, IGetFirebaseJwtRequest, IGetInteractiveSnapshotRequest,
+  IGetInteractiveSnapshotResponse, IInitInteractive, ILinkedInteractive, IReportInitInteractive,
+  ISupportedFeatures, ServerMessage
 } from "@concord-consortium/lara-interactive-api";
 import Shutterbug from "shutterbug";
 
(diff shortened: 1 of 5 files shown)


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -105,12 +106,50 @@ export const IframeRuntime: React.FC<IProps> =
       addListener("showModal", () => {
         toggleModal();
       });
-      post("initInteractive", {
-        mode: report ? "report" : "runtime",
+      // note: many of the values here are placeholders that require further
+      // consideration to determine whether there are more appropriate values.
+      const baseProps: Omit<IReportInitInteractive, "mode"> = {
```
pjanik, line 111: You've used `IreportInitInteractive` as that's the smallest interface, right?

pjanik (approved):
Looks good to me.

> I don't know what's up with the CodeCov failures given that this PR changes almost nothing from a coverage perspective.

I guess it's just +52 lines that are not covered by tests?

But yeah, in practice nothing changes here. I think we could actually add a chore about iframe-runtime testing, as it seems pretty important long-term.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
kswenson: Yes, of the two interfaces of interest here it has all of the shared properties and only one report-specific property.

scytacki (approved):
Seems OK to me. 
Not great that we are punting the work on the interactive state interfaces though. Is there a separate story to work on that, or can it be rolled into some existing story?

kswenson:
>Not great that we are punting the work on the interactive state interfaces though. Is there a separate story to work on that, or can it be rolled into some existing story?

If you are referring to the stubbed out parts of the `initInteractive` message, I think there should be a separate story. The only contribution of this PR is to properly type the `initInteractive` message. Fleshing out its contents seems like a non-trivial amount of work -- I don't know that the ActivityPlayer has ready access to everything it would need -- and seemed beyond the scope of this particular story/PR.

scytacki:
Yes I meant the initInteractive message. 

Can you put in a story about addressing that? 
I can pair with whoever works on it. I think every property the AP doesn't have information on can be made optional, and the rest can be filled out from info the AP already has.

kswenson:
>Can you put in a story about addressing that?

[[#175207294]](https://www.pivotaltracker.com/story/show/175207294)

PR #70, opened 2020-10-09
Fix theme colors broken by recent css-loader update

kswenson:
Fix css-loader configuration to support `:import` from `.scss` files
- required configuration changed with recent update to 4.3.0
- added cypress test to catch any future such configuration issue


mklewandowski (approved):
👍 LGTM!

PR #71, opened 2020-10-09
Refactor / reimplement modal API, add lightbox implementation

pjanik:
This PR:
- Makes modal API closer to the specification. Options are respected, we don't toggle modals, but listen to show and close messages, and so on.
- Adds lightbox implementation. It might not be fully polished, but I wanted to do it to clean up the image interactive.
- Completely removes injection of `isShowingModal` and `modalSupported` props. I've updated question interactives not to depend on them.

[#175181719]
[#175212002]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,9 @@ cypress/videos
 coverage-cypress
 coverage-jest
 .nyc_output
+.yalc
+yalc.lock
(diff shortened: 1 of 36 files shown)


kswenson (approved):
👍 Looks good!

PR #72, opened 2020-10-12
Improvements related to modals

pjanik:
[#175165872]

1. Request interactive state before closing dialog. This prevents state loss when the user clicks dialog backdrop. Note that interactive API client delays sending state by ~2 seconds, so it was very likely to happen before.
2. Host features hash updated.
3. Dialog API: handle .notCloseable property.
4. Add close icon to lightbox.

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -138,9 +138,11 @@ export const IframeRuntime: React.FC<IProps> = forwardRef((props, ref) => {
                   error: "",
                   mode: "runtime",
                   hostFeatures: {
-                    modalDialog: {
+                    modal: {
                       version: "1.0.0",
-                      imageLightbox: false
+                      lightbox: true,
+                      dialog: true,
+                      alert: false
                     }
                   },
                   globalInteractiveState: null,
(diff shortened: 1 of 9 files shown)


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -174,6 +181,10 @@ export const IframeRuntime: React.FC<IProps> =
   }, [url, authoredState, report, initialInteractiveState, setNewHint, getFirebaseJWT, setSupportedFeatures,
       setSendCustomMessage, showModal, closeModal]);
 
+  useImperativeHandle(ref, () => ({
+    requestInteractiveState: () => phoneRef.current?.post("getInteractiveState")
+  }));
```
kswenson, line 186: My reading of the docs suggests that the `forwardRef()`/`useImperativeHandle()` combination is specifically designed to handle the case of refs to DOM elements/React components which can't be passed as regular props due to the special-case handling of `ref` props. At the very least a comment seems warranted explaining how it's being used in this case. I would have thought that you could achieve the same effect by passing in a standard prop like `onSetRequestInteractiveState` but I suspect there's a reason you went with the `forwardRef()`/`useImperativeHandle()` approach.

kswenson (approved):
👍 Looks good -- just one suggestion to add a comment (unless you think the issue warrants further investigation).

src/components/activity-page/managed-interactive/iframe-runtime.tsx
pjanik: My understanding is that `useImperativeHandle` is a way to provide imperative API that could be done before just by defining a new class method. So, API that is provided by a child component and called by the parent. The opposite of typical callbacks provided via props. An example would be a parent calling `input.focus()` vs `onFocus` called by the chilled itself. 

Here the situation is the same. It's a way for the parent to tell the child (IframeRuntime) to request its state using iframe-phone. It happens when ManagedInteractive component is about to close the dialog with IFrameRuntime.

PR #73, opened 2020-10-13
Send Image Question to report service

pjanik:
[#174824371]
[#174069081]

### DIFF
--- a/src/utilities/embeddable-utils.ts
+++ b/src/utilities/embeddable-utils.ts
@@ -1,6 +1,9 @@
 import { v4 as uuidv4 } from "uuid";
 import { IAuthoringMetadata, IRuntimeMetadata } from "@concord-consortium/lara-interactive-api";
-import { IExportableAnswerMetadata, IManagedInteractive, IReportState, IExportableMultipleChoiceAnswerMetadata, IExportableOpenResponseAnswerMetadata, IExportableInteractiveAnswerMetadata } from "../types";
+import {
+  IExportableAnswerMetadata, IManagedInteractive, IReportState, IExportableMultipleChoiceAnswerMetadata,
+  IExportableOpenResponseAnswerMetadata, IExportableInteractiveAnswerMetadata, IExportableImageQuestionAnswerMetadata
+} from "../types";
 
 export const getAnswerWithMetadata = (
     interactiveState: IRuntimeMetadata,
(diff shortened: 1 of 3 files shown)


src/utilities/embeddable-utils.ts
```diff
@@ -46,6 +49,14 @@ export const getAnswerWithMetadata = (
...
       answer: interactiveState.answerText
     } as IExportableOpenResponseAnswerMetadata;
+  } else if (interactiveState.answerType === "image_question_answer") {
+    exportableAnswer = {
+      ...exportableAnswerBase,
+      answer: {
+        text: interactiveState.answerText,
+        image_url: interactiveState.answerImageUrl
+      }
+    } as IExportableImageQuestionAnswerMetadata;
```
kswenson, line 59: Not an issue specific to this PR, but doesn't this whole section violate the prime directive of the ActivityPlayer that it shouldn't need to know anything about the implementation details of the interactives it's managing? It seems like specifying the form of the `answer` should be the interactive's responsibility.

kswenson (approved):
👍 Looks good -- I had a question below but the answer is outside the scope of this PR.

src/utilities/embeddable-utils.ts
pjanik: I agree, but not sure if we have a better alternative.

> it shouldn't need to know anything about the implementation details of the interactives it's managing?

That's more or less true for regular interactives ("iframe_interactive" type and "interactive_state") - AP doesn't make too many assumptions about their internals.

But Report Service handles a few special cases like an open response, multiple-choice, and image question now, so it doesn't have to render iframe for each answer when it's just text or image (plus it makes aggregation views easier). In this case, there are more specific requirements for data that we send there. It's basically based on the old LARA format. 

That's why there are all the metadata interfaces that specify required props for these question types (both for authored state and user/interactive state). At some point, this data needs to be extracted from the interactive state (= user state) and added to the JSON that we send to Report Service. It's this file in AP and LARA has a similar set of methods. 

We don't just copy these props but reformat them slightly to avoid using underscores, and simplify naming and format when possible, as it originates from old LARA methods. Alternatively, we could force interactives to use a format that can be directly copied, but not sure if it'd make things much simpler. These methods here would be still necessary to extract it and send other properties.

Another option would be to do some of this transformation in Report Service itself. It uses some cloud functions to accept this data. But still, 95% of the code here would be necessary. And we'd add another pre-processing of the answer, so I think it isn't worth it.

PR #74, opened 2020-10-14
Restrict iframe reload to URL changes

kswenson:
[[#175254598]](https://www.pivotaltracker.com/story/show/175254598)

Normally, I’m in favor of following the `exhaustive-deps` rule’s recommendations rather than disabling it, but in this case I think it makes sense to go the other way. With this effect in particular we’ve repeatedly gone through the cycle of adding new code to the effect, adding the recommended identifiers to the list of dependencies, and then figuring out what combination of `useCallback`, `useMemo`, and `useRef` will suffice to make sure that none of them trigger a reload of the `<iframe>`. Rather than perpetuating that cycle, with this PR I propose to say explicitly what we mean, namely that only a URL change should trigger an `<iframe>` reload and any reload triggered otherwise is a bug.

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -184,8 +184,9 @@ export const IframeRuntime: React.FC<IProps> = forwardRef((props, ref) => {
         phoneRef.current.disconnect();
       }
     };
-  }, [url, authoredState, report, initialInteractiveState, setNewHint, getFirebaseJWT, setSupportedFeatures,
-      setSendCustomMessage, showModal, closeModal, setNavigation]);
+    // Re-running the effect reloads the iframe.
+    // The _only_ time that's ever appropriate is when the url has changed.
+  }, [url]);  // eslint-disable-line react-hooks/exhaustive-deps
 
   useImperativeHandle(ref, () => ({
     requestInteractiveState: () => phoneRef.current?.post("getInteractiveState")


mklewandowski (approved):
👍 LGTM! I'm not seeing any negative side effects from this change.  Seems like the right thing to do.

pjanik:
I was off today, but I've just noticed it. I think this particular change would work, but e.g. I'm not sure if setters returned from `useState` remain constant all the time or not. If they can be reinitialized randomly, it can introduce some bugs here. I guess theoretically they can, but in practice, they are not. 😉 

Also, when this useEffect is extended in the future, it might lead to similar issues.

I definitely noticed issues with iframe being reloaded before and I thought a bit about it. I think we shouldn't really try to prevent that by using useCallback/useMemo in parent components, as then parent makes assumptions about child implementation, and useCallback can still use dependencies and it can be re-run when we don't really expect it.

I guess the safe solution here would be to wrap all the necessary variables in refs, but do all that in IframeRuntime component, not in parents (and keep ref values updated). It'd ensure that all the callbacks in this problematic useEffect would always use the most recent value. It doesn't look pretty, but it might be the safest option.

kswenson:
@pjanik I didn't want you to think I'd forgotten about this discussion. 😉 The PR got merged and I got busy with other things, but I think it's an important conversation to have.

>I was off today, but I've just noticed it. I think this particular change would work, but e.g. I'm not sure if setters returned from useState remain constant all the time or not. If they can be reinitialized randomly, it can introduce some bugs here. I guess theoretically they can, but in practice, they are not.

The setters returned from `useState()` are guaranteed to be stable.

>I definitely noticed issues with iframe being reloaded before and I thought a bit about it. I think we shouldn't really try to prevent that by using useCallback/useMemo in parent components, as then parent makes assumptions about child implementation, and useCallback can still use dependencies and it can be re-run when we don't really expect it.

Another way to look at it is that parent components should _always_ use `useCallback`/`useMemo` for props passed to child components precisely because they shouldn't make assumptions about how those props will be used in the child implementation. The key difference is that only the creator of the entity has the ability to specify dependencies in `useCallback`/`useMemo` whereas downstream clients can only use it as provided (i.e. re-run the effect whenever the prop changes according to the parent implementation) or wrap the prop with `useRef` (i.e. never re-run the effect because of it at all). The creator of the entity has a scalpel while the recipient only has a cudgel. The child can't know whether a prop change is meaningful unless the parent restricts changes to meaningful ones.

>I guess the safe solution here would be to wrap all the necessary variables in refs, but do all that in IframeRuntime component, not in parents (and keep ref values updated). It'd ensure that all the callbacks in this problematic useEffect would always use the most recent value. It doesn't look pretty, but it might be the safest option.

There are two risks to be considered here with respect to the "safest option":
1. Future changes to the code could introduce new props that _should_ trigger a re-run of the effect but don't (e.g. because the `exhaustive-deps` rule has been disabled).
2. Future changes to the code could introduce new props that _shouldn't_ trigger a re-run of the effect but do.

A defining characteristic of the `IframeRuntime` component is that reloading the `<iframe>` leads to a terrible user experience and should be avoided except when absolutely necessary. If `IframeRuntime` were implemented as a class component, we would put this code in `componentDidMount` and never give it another thought because doing so precisely expresses the desired behavior of initializing the handlers on initial render and never reloading the `<iframe>` again during the lifetime of the component. It is only because functional components/hooks require us to put our one-time-initialization code in the "render" function via `useEffect` that we're having this discussion at all.

As such, my engineering assessment is that the likelihood of occurrence of the first risk above is essentially 0, because not reloading the iframe is one of the defining requirements of the component. Conversely, the likelihood of occurrence of the second risk is ~100%, because it has already occurred multiple times and seems to occur whenever new props/features are added. Disabling the `exhaustive-deps` rule is analogous to putting the code in `componentDidMount` for a class component. It expresses the engineering assessment that the default behavior of the code should be to _not_ re-render as future changes are made. If some future prop is introduced which _should_ trigger a re-render, then it is incumbent upon that engineer to recognize the unusual nature of that prop and to understand that it should be added to the dependencies without being instructed to do so by the `exhaustive-deps` rule. In that context, wrapping everything in refs just seems like unnecessary busy-work.

I've said it before, but what I'd really like is a version of `useEffect()` which takes a second array of symbols to ignore. Then code like this would work, but if the code changed any new symbols would still be flagged if appropriate:
```typescript
const unusedButShouldTrigger = 1;
const usedButShouldntTrigger = 1;
useEffectIgnorableDeps(() => {
  ... code that uses usedButShouldntTrigger but not unusedButShouldTrigger
}, [unusedButShouldTrigger], [unusedButShouldTrigger, usedButShouldntTrigger]);
```
but you'd have to also write a custom version of the `exhaustive-deps` rule to understand the `useEffectIgnorableDeps()` hook and to respect the second array. I actually went so far as to look at the source for the `exhaustive-deps` rule and I don't think it would be too hard to extend it to support the second "ignore" array, but I don't know that it's worth it.

pjanik:
> As such, my engineering assessment is that the likelihood of occurrence of the first risk above is essentially 0, because not reloading the iframe is one of the defining requirements of the component. Conversely, the likelihood of occurrence of the second risk is ~100%, because it has already occurred multiple times and seems to occur whenever new props/features are added.

I agree with this assessment and everything that's below. `useRefs` would provide additional safety against the first risk in case it's > 0% at the cost of more lines and hitting reload issues again in the future. But I agree the first risk is pretty close to 0% in this particular case, as iframe reloading is so specific and so disturbing for users.

> The child can't know whether a prop change is meaningful unless the parent restricts changes to meaningful ones.

`usePrevious` could be used to check this. Still, if prop change has some critical effects like iframe reload, I think it should be handled in the affected component => child. Either by refs, empty dependencies array, or some other way. I agree that useMemo or useCallback won't hurt and we can use it often, but I just wouldn't make it obligatory or a way to fix issue in a child component. Just an optional (performance) improvement. This can be totally enough when child component rerender isn't that visible as in this particular case.

PR #75, opened 2020-10-16
174201695 logging

mklewandowski:
Adds logging to the Activity Player.  Logging features include:
- log messages from interactives and plugins as well as AP UI events (change page, change sequence activity, toggle hint, toggle modal, toggle sidebar, etc.)
- each question interactive log message follows the form seen in LARA with additional AP-specific fields that follow design in CLUE (things such as username, classhash, etc.).  In essence I tried to make additions, not changes (which is why things like the URL or interactive ID are added as properties on the log message and not stashed in `parameters`).
- in other cases, parameters specific to a log message are stored in the `parameters` field, but this is open to discussion (I originally used `event_value` but was advised to use `parameters` instead).

There's probably some massaging needed before brining this PR in.  I'm still uncertain how to verify that the logging is indeed working.  A few other questions:
- as mentioned above (and maybe outside the scope of this PR), it's unclear what best practices are in terms of how to use `event_value` vs `parameters` vs. other fields.
- anyone have thoughts on logging interactions with generic iframes?
- am I logging the remote endpoint correctly?


kswenson (approved):
👍 Looks okay to me, but I don't have a lot of experience with logging code in other apps.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -121,6 +122,9 @@ export const IframeRuntime: React.FC<IProps> = forwardRef((props, ref) => {
       addListener("closeModal", (options: ICloseModal) => {
         closeModal(options);
       });
+      addListener("log", (logData: any) => {
+        Logger.log(logData.action, logData.data, logData.value, id, url);
+      });
```
pjanik, line 127: I think it'd be good if this function was following LARA implementation (provided the same fields or more), so we get the same logs no matter if we run LARA or AP:
https://github.com/concord-consortium/lara/blob/master/app/assets/javascripts/logger.js#L199-L207
but following the calls chain here, it seems it does, right?

src/lib/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    extends: "../../.eslintrc.js",
+    rules: {
+      "no-console": "off",
+    }
+};
```
pjanik, line 6: Maybe it's enough to disable no-console per line?
I kinda like this rule (as warning), as I sometimes use `console.log` to debug manually and then forget to remove it before I commit. ;)

src/components/app.tsx
```diff
@@ -263,21 +279,27 @@ export class App extends React.PureComponent<IProps, IState> {
     } else if (page >= 0 && (activity && page <= activity.pages.length)) {
       this.setState({currentPage: page, incompleteQuestions: []});
       setDocumentTitle(activity, page);
+      Logger.updateActivityPage(page);
+      Logger.log(LogEventName.change_activity_page, { new_page: page });
```
pjanik, line 283: Not sure if it's worth it, but are we following LARA log format? I guess LARA should emit something similar.

pjanik (approved):
I had time for a quick look and it looks good to me. I've left a few questions, but nothing critical.

> with additional AP-specific fields that follow design in CLUE (things such as username, classhash, etc.). 

Shouldn't we try to follow LARA design/format? Since AP is LARA equivalent. But it seems to me that it's true actually (looking at fields that we're sending). Just mentioning that to double-check as I don't have time to dig into details of the messages.

@scytacki might want to double-check messages format too or have some opinion about parameters vs event_value, as he's involved in design of the new logging system.

scytacki:
I'm not too concerned about following exactly the LARA messages. I don't think they were used much, so if there are small changes that improve the usability of the AP messages that should be fine.

I haven't looked at this closely, but I did want to clarify at question that I guess was asked on slack. The `run_remote_endpoint` is very important to so researchers can find the data they need in the log manager. The name of that property is bad. Its basic function is an id for (assignment + student). A researcher gets a list of these run_remote_endpoints from the portal and then those are used to query the log database for all events with matching run_remote_endpoints.

This section of the docs have more information about all of this: https://docs.google.com/document/d/1LwNA2lI578n8-niZzkb9OlmoZZ-8Mhx05yb-U-9Z62E/edit#heading=h.ag50xzvec9db

I don't think it is critical, but the name of `loggingRemoteEndpoint` confuses me. It makes me think this is an endpoint to send the logs to. The name run_remote_endpoint is bad because it has lost its meaning, but for context: it is called that because LARA used to send the students answers to this Portal endpoint. The 'Run' represents the object storing the student work. So this was a remote endpoint for the Run. In the Activity Player this term Run is also used for anonymous users. It isn't easy to change the name of this property because it is scattered across many systems and many versions of different codebases so we are kind of stuck with it.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
mklewandowski: Yes, eventually the final message gets structured as it is structured in LARA (with extra fields added that were part of the requested additions in the story).  However, I added a commit that restructures the `Logger.log` function so the argument structure is the same as in LARA.  Hopefully it makes it easier to read and easier to compare to LARA.

src/components/app.tsx
mklewandowski: I wasn't entirely sure on these new events.  I tried to follow the LARA structure (`event`, `parameters`, etc.) and then added new fields that were requested by the story (many of which are state values independent of the event - things like current activity, current page, etc.).  Hopefully my last commit makes this function call clearer so you can see what is actually going to be in the final log message.

mklewandowski:
> I don't think it is critical, but the name of `loggingRemoteEndpoint` confuses me. It makes me think this is an endpoint to send the logs to. The name run_remote_endpoint is bad because it has lost its meaning, but for context: it is called that because LARA used to send the students answers to this Portal endpoint.

Yeah, I'm not sure what to call that internal variable.  The field in the message, `run_remote_endpoint`, is in sync with the name used in CLUE - so at least two of the apps are unified.  Maybe the internal variable should just be `runRemoteEndpoint` for now?  It will at least be a little easier for someone looking at the code to see that the private class variable is tied to the field in the log message.

scytacki:
Yes the field in the message has to be `run_remote_endpoint`.  I agree that runRemoteEndpoint for the variable name seems less confusing.

PR #76, opened 2020-10-30
Fix interactive sizing in header section

mklewandowski:
This PR fixes a bug found by @emcelroy while working on the carousels.  Full-width embeddables in the header section were not being sized correctly to fill the horizontal space.  This was partly due to incorrect CSS being applied to the header container itself and incorrect CSS being applied to each individual header embeddable (a remnant of early work from before the header could contain any embeddable content - we never went back and fixed some of this CSS when we changed the header capabilities).

image

### DIFF
--- a/src/components/activity-page/activity-page-content.scss
+++ b/src/components/activity-page/activity-page-content.scss
@@ -18,10 +18,6 @@
     color: var(--theme-secondary-color);
   }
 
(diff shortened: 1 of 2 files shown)


kswenson (approved):
👍

emcelroy (commented):
Thanks 👍

PR #77, opened 2020-11-03
172894510 anonymous student report

eireland:
Adds ability to open an anonymous student's report from the Generate Report button at the end of an activity. Accompanying work was done in report-service, and portal-report repo.


pjanik:
Is this PR opened correctly?
There are 102 commits, 114 changed files, and +7,671 −2,540 lines, what sounds like a week of review. ;) After looking at some random changes, it doesn't seem it's related to reports. Maybe some git issues?

PR #78, opened 2020-11-04
Fix header embeddable sizing bugs

mklewandowski:
This PR adds more fixes to the header content sizing.  Changes include:
- the page header/intro max-width is brought back (was removed in previous PR).  LARA does in fact enforce a max-width of content in the header/intro which is needed when displaying the header/intro on a page with responsive layout.  This was erroneously removed in the last PR.
- the embeddable container in the page header/intro now includes the `responsive` class which allows non-full-width embeddables to wrap and display side-by-side in the page header/intro section.
- refinement of the code in the `Embeddable` component that determines if the embeddable is full or half width.  Half-width embeddables in the page header/intro section should include the `reduced-width` CSS class so that they are sized properly.

### DIFF
--- a/src/components/activity-page/activity-page-content.scss
+++ b/src/components/activity-page/activity-page-content.scss
@@ -18,6 +18,10 @@
     color: var(--theme-secondary-color);
   }
 
+  .introduction {
+    max-width: 900px;
+  }
+
   .embeddables {
     display: flex;
     flex-direction: row;
--- a/src/components/activity-page/activity-page-content.tsx
+++ b/src/components/activity-page/activity-page-content.tsx
@@ -135,7 +135,7 @@ export class ActivityPageContent extends React.PureComponent <IProps, IState> {
     let questionNumber = totalPreviousQuestions;
     return (
       <React.Fragment>
-        { embeddables.map((embeddableWrapper, i: number) => {
+        { embeddables.map((embeddableWrapper) => {
             if (isQuestion(embeddableWrapper)) {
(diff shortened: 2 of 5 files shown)


emcelroy (approved):
This looks good to me. 

Was the only problem with removing that max-width limitation that it didn't match how LARA runtime limits the width? Or did removing it actually cause problems with the layout of items? 

You probably already know this, but there's disagreement over whether that max-width limitation should be there. I think most people don't like it and I don't think anybody would miss it if it was removed. I don't think there's ever been an official decision on that though, so it's best to keep as is for now. It's something we should revisit at some point, though.

kswenson (approved):
👍

mklewandowski:
> Was the only problem with removing that max-width limitation that it didn't match how LARA runtime limits the width? Or did removing it actually cause problems with the layout of items?
> 
> You probably already know this, but there's disagreement over whether that max-width limitation should be there. I think most people don't like it and I don't think anybody would miss it if it was removed. I don't think there's ever been an official decision on that though, so it's best to keep as is for now. It's something we should revisit at some point, though.

I didn't know that there was a debate on this issue.  I put it back once I realized that on responsive pages in LARA that max-width limitation is indeed used (it also requires more changes to get the sizing of the embeddables to work in that same situation.  So if we remove it, there is a little more work to do).  I agree that it's best to leave it in for now.  I'd rather error on the side of what LARA currently does rather than trying to make a decision on something without broader discussion outside of this PR.

PR #79, opened 2020-11-05
Update Firestore settings => ignoreUndefinedProperties: true

pjanik:
[#174502804]

Values set to undefined values were breaking interactive state saving in some cases. I don't really see a reason why would we want to have this option disabled.

### DIFF
--- a/src/firebase-db.ts
+++ b/src/firebase-db.ts
@@ -102,6 +102,13 @@ export async function initializeDB(name: FirebaseAppName) {
     await firebase.firestore().disableNetwork();
   }
 
+  // See: https://concord-consortium.slack.com/archives/C0303QUNU/p1604596779407700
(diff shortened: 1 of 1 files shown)


src/firebase-db.ts
```diff
@@ -102,6 +102,13 @@ export async function initializeDB(name: FirebaseAppName) {
     await firebase.firestore().disableNetwork();
   }
 
+  // See: https://concord-consortium.slack.com/archives/C0303QUNU/p1604596779407700
```
kswenson, line 105: Do we have a policy on linking to internal CC slack conversations from public GitHub repos? Seems like it would be better to copy the relevant content from the slack conversation so that outsiders aren't confused when they can't access it.

kswenson (approved):
👍 Code change looks fine. One suggestion about the comment.

pjanik:
@kevinsantos-cc, thanks, good point, I've updated the comment.

PR #80, opened 2020-11-06
Use Firestore offline mode in AP preview mode

pjanik:
This a small PR to make watching interactive states easier.

Now, AP should always use Firestore, even in the preview mode. This lets me use db-related mechanisms for watching answers (this work is still in progress) => observing interactive state. I don't have to rely on some internal states of components in AP tree, I can just rely on Firestore values. In the future, this will let us use the same approach for observing interactive state from a different page (it should just come for free). And generally this change should generally make things make more consistent, as we can simply assume that Firestore is always used, but sometimes not use the network.

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -5,7 +5,7 @@ import iframePhone from "iframe-phone";
 import {
   ClientMessage, ICustomMessage, IGetFirebaseJwtRequest, IGetInteractiveSnapshotRequest,
   IGetInteractiveSnapshotResponse, IInitInteractive, ILinkedInteractive, IReportInitInteractive,
-  ISupportedFeatures, ServerMessage, IShowModal, ICloseModal, INavigationOptions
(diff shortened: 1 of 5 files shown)


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -5,7 +5,7 @@ import iframePhone from "iframe-phone";
 import {
   ClientMessage, ICustomMessage, IGetFirebaseJwtRequest, IGetInteractiveSnapshotRequest,
   IGetInteractiveSnapshotResponse, IInitInteractive, ILinkedInteractive, IReportInitInteractive,
-  ISupportedFeatures, ServerMessage, IShowModal, ICloseModal, INavigationOptions
+  ISupportedFeatures, ServerMessage, IShowModal, ICloseModal, INavigationOptions, IGetInteractiveSnapshotOptions
```
pjanik, line 8: Ops, that's a bit random, I'll clean it up in the upcoming PRs.

kswenson (approved):
👍 LGTM

PR #81, opened 2020-11-13
Fix / refactor Firestore helpers and interactive state handling

pjanik:
The main problem was that interactives were sometimes sending multiple docs to Firestore, as answerMeta was never updated, and they're generating new UUID each time. This PR fixes that by more general refactoring of firestore helpers and interactive state handling.

- Firestore helpers are simplified and reorganized a bit. There are no separate listeners and local DB copy, as it seemed a bit redundant. Firestore already provides API to observe docs. 
- Interactive state handling is removed from `embeddable.ts` and limited to `managed-interactive.ts`. 
- linked_interactives are passed to MwInteractives too (not only ManagedInteractives).

### DIFF
--- a/src/components/activity-page/embeddable.test.tsx
+++ b/src/components/activity-page/embeddable.test.tsx
@@ -1,15 +1,13 @@
 import React from "react";
-import { waitFor } from "@testing-library/dom";
 import iframePhone from "iframe-phone";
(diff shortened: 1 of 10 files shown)


kswenson (approved):
👍 Looks good -- seems like a nice simplification.

PR #83, opened 2020-11-13
172894510 anonymous student report

eireland:
Adds ability to open an anonymous student's report from the Generate Report button at the end of an activity. Accompanying work was done in portal-report repo. Adds a query param "portal-report" so user can specify which version of the portal report to open. Portal report link is currently defaulted to open portal report master branch.

### DIFF
--- a/src/components/activity-page/activity-page-content.tsx
+++ b/src/components/activity-page/activity-page-content.tsx
@@ -10,7 +10,7 @@ import IconChevronLeft from "../../assets/svg-icons/icon-chevron-left.svg";
 import { Page, EmbeddableWrapper } from "../../types";
 import { INavigationOptions } from "@concord-consortium/lara-interactive-api";
 import { Logger, LogEventName } from "../../lib/logger";
-import { queryValue, queryValueBoolean } from "../../utilities/url-query";
+import { queryValue } from "../../utilities/url-query";
 
 import "./activity-page-content.scss";
 
@@ -126,7 +126,12 @@ export class ActivityPageContent extends React.PureComponent <IProps, IState> {
     this.props.onPageChange(this.props.pageNumber + 1);
   }
   private handleReport = () => {
-
+    const reportLink = "https://portal-report.concord.org/version/v3.0.0/index.html";
+    const runKey= queryValue("runKey");
+    const activity = queryValue("activity");
+    const activityUrl = activity? ((activity.split(".json"))[0]).replace("api/v1/","") : "";
+    const answerSource = window.location.hostname;
+    window.open(reportLink + "?runKey=" + runKey + "&activity=" + activityUrl + "&answerSource="+answerSource);
     // TODO: create report when pressed
     Logger.log({
       event: LogEventName.create_report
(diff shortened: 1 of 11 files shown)


src/firebase-db.ts
```diff
@@ -222,7 +222,8 @@ export function createOrUpdateAnswer(answer: IExportableAnswerMetadata) {
       resource_url: portalData.resourceUrl,
       tool_id: portalData.toolId,
       tool_user_id: "anonymous",
-    };
+      platform_user_id: portalData.runKey
+    } as AnonymousRuntimeAnswerMetadata;
```
pjanik, line 226: Why is this casting necessary? Asking, as I think `as <type>` is being overused sometimes in our projects. It's usually better to declare variable as given type and then make an assigment, like:

`const something: AnonymousRuntimeAnswerMetadata = { ... }`

as then we have benefits from type checking.

pjanik (commented):
Looks good. 👍 Just one inline question.

src/firebase-db.ts
scytacki: In this case postAnswer can be one of two different types.  I double checked and you are right that the cast with `as` doesn't do any type checking of the object being casted.
So added to `const` variables for each type so we get some actual type checking of the object literals here.

So we could just cast the first type as well on line 216 to make this more clear. However based on your comment it sounds like casting doesn't provide as much type checking 

We could just remove the `as` here. However because postAnswer can have two different types. It isn't staightforard it seems useful to say that when `portalData.type` is not "authenticated" that the type of the object should be. Because postAnswer can have two types depending on portalData.type. I think this `as AnonymousRuntimeAnswerMetadata`

pjanik: Looks good. I'd expect that:
`answerDocData: LTIRuntimeAnswerMetadata | AnonymousRuntimeAnswerMetadata;`
would let us use this variable in both if / else blocks, without this intermittent variable narrowed just to one type. But nothing bad in the current approach, maybe TS isn't smart enough yet.

pjanik (approved):

src/components/activity-page/activity-page-content.tsx
```diff
@@ -125,7 +127,12 @@ export class ActivityPageContent extends React.PureComponent <IProps, IState> {
     this.props.onPageChange(this.props.pageNumber + 1);
   }
   private handleReport = () => {
-    // TODO: create report when pressed
+    const reportLink = (queryValue("portal-report") as string) || DEFAULT_PORTAL_REPORT_URL;
```
mklewandowski, line 130: Could we add a note in the readme about the URL param?

src/components/app.tsx
```diff
@@ -31,6 +31,8 @@ import "./app.scss";
 
 const kDefaultActivity = "sample-activity-multiple-layout-types";   // may eventually want to get rid of this
 const kDefaultIncompleteMessage = "Please submit an answer first.";
+// TODO: switch default to production report version before production deploy
+export const DEFAULT_PORTAL_REPORT_URL = "https://portal-report.concord.org/branch/master/";
```
mklewandowski, line 35: I would probably add this to a utils module.  Seems a little odd to have it here.

mklewandowski: Also the AP tends to use kConstant over CONSTANT notation.

mklewandowski (approved):
👍 Looks good! A couple quick things:
- Could we add a note in the readme about the URL param?
- You might consider adding DEFAULT_PORTAL_REPORT_URL to a util module or at least changing it to kConstant instead of CONSTANT.
- EDIT: is this different than the buttons that appear on the completion page and the single page activity?  Do those buttons do something else (and hence belong in a different story).

mklewandowski:
I'm also curious if this report is different than the one that appears on the completion page when we click the Show All Answers button (look for the function `handleShowAnswers`) and in the single page activity when we click the Submit button (look for the `onSubmit` prop in the `SubmitButton`).

PR #84, opened 2020-11-14
Text box styling

eireland:
Adds spacing between paragraphs and listed items

### DIFF
--- a/cypress/integration/data-saving-anonymous.ts
+++ b/cypress/integration/data-saving-anonymous.ts
@@ -27,7 +27,7 @@ context("Saving and loading data as an anonymous user", () => {
 
       cy.visit(activityUrl);
       activityPage.getNavPage(2).click();
-      cy.wait(500);
+      cy.wait(3000);
       getIframeBody("body").find("[data-cy=choices-container] input").eq(1).click({force: true});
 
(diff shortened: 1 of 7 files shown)


cypress/integration/data-saving-anonymous.ts
```diff
@@ -27,7 +27,7 @@ context("Saving and loading data as an anonymous user", () => {
 
       cy.visit(activityUrl);
       activityPage.getNavPage(2).click();
-      cy.wait(500);
+      cy.wait(3000);
```
mklewandowski, line 30: Is this still needed?

src/components/activity-page/text-box/text-box.scss
```diff
@@ -2,8 +2,12 @@
 
 .textbox {
   p {
-    margin: 0;
+    margin: 20px 0;
```
mklewandowski, line 5: 20 on both sides seems big.  Won't back-to-back paragraphs get 40 pixels between them?  I would think that `10px 0` would be better (unless this is what the story requests).

mklewandowski: I take that back.  I think this looks fine (and I appear to be wrong about the duplicated margins - it doesn't become 40px):
image

mklewandowski (approved):
👍

PR #88, opened 2020-11-23
Add async load of plugin scripts before loading page content

mklewandowski:
Use async/await to load the plugin scripts before loading the rest of the page content.  Previously, plugins only worked if the user went to the intro page first (where the plugin scripts would be loaded) and then navigated to the activity internal pages (at this point, the plugin scripts had been loaded on the intro page and could be used by the internal page embeddables).  However, if a user navigated directly to a page or to a single page activity, the embeddable page content would be loaded before the plugins, would attempt to access the not-yet-loaded plugin scripts. and the plugin content would fail to be displayed.

I tested some examples of using both the TE and glossary plugins on a page, and I do not see any noticeable slow down caused by waiting for the plugin scripts to load.

For comparison, here is how direct page load with plugins looks on master (note that content fails to load):
https://activity-player.concord.org/branch/master/?mode=teacher-edition&preview&activity=sample-activity-plugins&page=2

And here is how it looks on the PR branch:
https://activity-player.concord.org/branch/plugin-load/?mode=teacher-edition&preview&activity=sample-activity-plugins&page=2

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -91,6 +91,11 @@ export class App extends React.PureComponent<IProps, IState> {
       const newState: Partial<IState> = {activity, currentPage, showThemeButtons, showSequence, sequence, teacherEditionMode};
       setDocumentTitle(activity, currentPage);
 
+      this.LARA = initializeLara();
+      if (teacherEditionMode) {
+        await loadPluginScripts(this.LARA, activity);
+      }
+
       let classHash = "";
       let role = "unknown";
       let runRemoteEndpoint = "";
@@ -129,11 +134,6 @@ export class App extends React.PureComponent<IProps, IState> {
       }
 
       this.setState(newState as IState);
-
-      this.LARA = initializeLara();
-      if (teacherEditionMode) {
-        loadPluginScripts(this.LARA, activity);
-      }
 
       Modal.setAppElement("#app");
 
--- a/src/utilities/plugin-utils.ts
+++ b/src/utilities/plugin-utils.ts
@@ -22,7 +22,7 @@ export const Plugins: PluginInfo[] = [
   },
 ];
 
-export const loadPluginScripts = (LARA: LaraGlobalType, activity: Activity) => {
+export const loadPluginScripts = async (LARA: LaraGlobalType, activity: Activity) => {
   // load any plugin scripts, each should call registerPlugin if correctly loaded
   const usedPlugins: PluginInfo[] = [];
   for (let page = 0; page < activity.pages.length - 1; page++) {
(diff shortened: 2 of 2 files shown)


kswenson (commented):
I have nothing to add beyond sharing Scott's concern that this will delay loading the page.

mklewandowski:
> I have nothing to add beyond sharing Scott's concern that this could delay loading the page.

Right.  This is at least an improvement over the current behavior which is quite simply that the plugins will not work once we add support for student progress (as well as other situations like a single page activity).  The other alternative is to load page content, but selectively hold off on handling plugin content until the plugin scripts are loaded.  I don't see as obvious of a solution if we go that route, and it will surely be a much large piece of work that affects more of the AP code and components.  Not my decision to make, but it will definitely be a larger story if we go that route.

I would also argue that these additional hurdles are another side effect of having so much of the AP content being loaded from external sites.  Loading wait times and content load orders wouldn't be an issue if these pieces were native to the AP.  Too late for that, but it's worth noting.

kswenson (approved):
>> I have nothing to add beyond sharing Scott's concern that this could delay loading the page.

>Right.  This is at least an improvement over the current behavior

Loading the page slowly is certainly better than not loading it at all. 😉 Ultimately, we'll presumably want to start as many of these requests as we can in parallel rather than serializing them but that's a performance optimization rather than basic functionality.

scytacki:
Yeah I think this is fine, but I hope we can figure out a way for this to work in parallel.

I think in LARA the interactive iframes load before the plugins and then the plugins wrap them. That is why the teacher edition stuff takes a little time to show up. I'd guess this might be fixable when the plugin loading code is improved to handle multiple plugins.  It isn't clear to me from this PR why the plugins need to be loaded first.

scytacki (approved):

mklewandowski:
> Yeah I think this is fine, but I hope we can figure out a way for this to work in parallel.
> 
> I think in LARA the interactive iframes load before the plugins and then the plugins wrap them. That is why the teacher edition stuff takes a little time to show up. I'd guess this might be fixable when the plugin loading code is improved to handle multiple plugins. It isn't clear to me from this PR why the plugins need to be loaded first.

Yeah, I'm starting to question this PR and am currently digging in further to understand why the plugins don't load when accessing a page directly.  This might be caused by a different issue altogether.  For now I'm going to hold off on this PR until I can get a deeper understanding of what is happening.

mklewandowski:
> Yeah I think this is fine, but I hope we can figure out a way for this to work in parallel.
> 
> I think in LARA the interactive iframes load before the plugins and then the plugins wrap them. That is why the teacher edition stuff takes a little time to show up. I'd guess this might be fixable when the plugin loading code is improved to handle multiple plugins. It isn't clear to me from this PR why the plugins need to be loaded first.

My handling here of loading the plugins first was because the following could happen:
- activity page content is loaded in `App` component
- plugin script load is started
- `App` component passes page content to child React components in props
- child React components parse activity page content in props and determine that plugin content needs to be loaded.  In doing so `initializePlugin` is called which attempts to `LARA.Plugins.initPlugin`
- however, the plugin scripts haven't finished loading and we're not actually ready to call `LARA.Plugins.initPlugin`

I'm working on a branch now, but, again, I think there is something else preventing plugin content from loading when we use a URL param to go straight to an internal page.

Granted, I don't think that could be what I'm seeing here:
https://activity-player.concord.org/branch/master/?mode=teacher-edition&preview&activity=sample-activity-plugins&page=2
If the above happened, wouldn't it just crash or at least throw an exception?

Regardless, after thinking this through, it seems like a better path is:
- start with flag `pluginsLoaded = false`.
- start loading all of the embeddable page content, but use `pluginsLoaded` flag so we don't try to render plugin content until we know the plugin scripts are done loading.  
- start loading plugin scripts
- when plugin scripts are done loading, set `pluginsLoaded = true`
- pass flag to any components that need it so they can now call `initializePlugin` and try to display plugin content

mklewandowski:
@scytacki @kswenson here is a branch that tries a different approach (the one mentioned above) where we let the embeddables/question interactives and the plugins load without having the plugins block other content loading.  We then only tell the embeddables to init plugins (and display plugin content) once the scripts are loaded.
https://github.com/concord-consortium/activity-player/compare/175844168-plugin-load-v2

I put in an intentional 4 second delay when loading the plugin scripts so that we can see that the embeddable/question interactive content load isn't hung up if the plugin load is taking a while (or fails).  Here's some examples with the artificial 4 second delay:
https://activity-player.concord.org/branch/plugin-load-v2/?mode=teacher-edition&preview&activity=sample-activity-plugins&page=1
https://activity-player.concord.org/branch/plugin-load-v2/?mode=teacher-edition&preview&activity=sample-activity-plugins-single-page

mklewandowski:
I'm closing this PR.  I don't think it's the best approach, and I'd rather open up a draft PR on this branch:
https://github.com/concord-consortium/activity-player/tree/175844168-plugin-load-v2

PR #89, opened 2020-11-24
Fix header content in single page layout

mklewandowski:
This PR fixes the header content in the single page layout activity.  Previously no activity title was shown and an empty nav bar was shown (a big empty rectangle).  I believe we should show the activity title in the header and only show the nav bar if needed (the only case is if we are in a sequence and need to show a link back to the main sequence page).

Previous version:
image

Updated version:
image


emcelroy (approved):
Leaving the navbar out in single page activities (that aren't part of a sequence) makes sense to me. Looks good 👍

kswenson (commented):
👍 Looks good

PR #91, opened 2020-11-24
Update readme

mklewandowski:
This PR makes small readme updates (I noticed that there were still references to Travis and that the sequence URL parameter usage was never added).


emcelroy (approved):
👍

kswenson (commented):
👍

PR #92, opened 2020-11-25
173400275 glossary highlight

mklewandowski:
This PR adds the initial implementation of the glossary plugin which includes the glossary sidebar and glossary popups.  Glossary terms in the text of activity page text boxes, the intro page summary, and the page sidebar are converted to clickable links (which launch the glossary popup).  

Changes include:
- add `jQuery-ui` and `text-decorator` dependencies
- add LARA plugin modules to `src/lara-plugin` folder (mark changes with comments) - this includes the text decorator and popup modules
- add new sample data
- add `GlossaryPlugin` component which inits the glossary plugin
- support multiple plugins by adding a `data-id` attribute with plugin label value to the loaded plugin script.  We can then obtain the plugin label from the `data-id` attribute using `document.currentScript` in `registerPlugin`

Demo here:
https://activity-player.concord.org/branch/glossary-highlight/?preview&activity=sample-activity-glossary-plugin&mode=teacher-edition

glossary


src/components/app.tsx
```diff
@@ -191,15 +192,19 @@ export class App extends React.PureComponent<IProps, IState> {
...
         }
-        { (activity.layout === ActivityLayouts.SinglePage || (currentPage !== 0 && !activity.pages[currentPage - 1].is_completion)) &&
+        { (activity.layout === ActivityLayouts.SinglePage || !isCompletionPage) &&
           <ExpandableContainer
             activity={activity}
             pageNumber={currentPage}
             page={activity.pages.filter((page) => !page.is_hidden)[currentPage - 1]}
             teacherEditionMode={teacherEditionMode}
             pluginsLoaded={pluginsLoaded}
+            glossaryPlugin={glossaryEmbeddable !== undefined}
```
kswenson, line 202: I would use `!= null` or even just `!!glossaryEmbeddable` here on the grounds that `null` or other falsy values aren't valid either.

src/lara-plugin/plugin-api/sidebar.ts
```diff
@@ -1,7 +1,10 @@
+// LARA_CODE: import * as $ from "jquery";
 // ACTIVITY_PLAYER_CODE:
 import $ from "jquery";
-// LARA_CODE: import * as $ from "jquery";
-// LARA_CODE: import "jquery-ui/ui/widgets/button";
+(window as any).jQuery = $;
+(window as any).$ = $;
+
+import "jquery-ui/ui/widgets/button";
```
kswenson, line 7: Given that an initialization sequence like this occurs in multiple places, I wonder if it would make sense to put it all in a single file that can be imported elsewhere, e.g. `import-jquery.ts` or `init-jquery.ts`.

kswenson (approved):
👍 LGTM

src/lara-plugin/plugin-api/sidebar.ts
mklewandowski: I'm going to leave it inline for now - I think it makes it a little easier to diff and compare against the original (which was Scott's intention with this work).  But I'll be open to this if I have to continue with the import pattern.

PR #93, opened 2020-12-01
175606584 page numbers

mklewandowski:
This PR adds the navigation buttons to the bottom of each activity page.  Note that this styling will most likely be updated once Michael takes a UI pass at the AP, but for now this should get the job done.

Typical activity page:
image

Page with report button:
image

Single page activity (report button shows, no navigation is shown):
image

### DIFF
--- a/cypress/integration/workspace.test.ts
+++ b/cypress/integration/workspace.test.ts
@@ -17,7 +17,7 @@ context("Test the overall app", () => {
   describe("ActivityNavHeader",()=>{
     it("verify ActivityNavHeader loads",()=>{
         cy.get("[data-cy=activity-nav-header]").should("be.visible");
-        cy.get(".page-button").should("have.length", 11);
+        cy.get(".page-button").should("have.length", 22); // top and bottom nav headers exist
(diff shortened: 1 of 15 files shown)


emcelroy (approved):
This looks good 👍  This may be nitpicking, especially since this might be updated later, but would it be good to change some of the names? If I understand correctly, it seems like `renderNavHeader` and `ActivityNavHeader` should be changed to something more generic since they're now used to add the nav buttons at the top and bottom. Maybe just `renderNav` and `ActivityNav`?

mklewandowski:
> This looks good 👍 This may be nitpicking, especially since this might be updated later, but would it be good to change some of the names? If I understand correctly, it seems like `renderNavHeader` and `ActivityNavHeader` should be changed to something more generic since they're now used to add the nav buttons at the top and bottom. Maybe just `renderNav` and `ActivityNav`?

Yes, I think this is a good idea.  I added a commit that renames the render function, the component, the files, any CSS classes, and any data-cy attributes to be in line with how the component is now used.

PR #94, opened 2020-12-08
173944477 completion page state

eireland:
Updates completion page with student status and correct text for activity and sequence completion based on student status. May need some adjustment for number of answers or the definition what a question is. Some interactives (tried with CBIO) are counted as questions, but no answer is saved for them so the user is never finished. 
Had to changes some of the interactives (the ones that load Wikipedia site) in the sample doc to not be questions.
Also makes Cypress ignore uncaught exceptions to ignore the ResizeObserver error as a suggested solution. Similar code is found in CLUE.


src/components/activity-completion/completion-page-content.scss
```diff
@@ -10,33 +10,45 @@
   .completion-text {
     color: $cc-teal-dark1;
     font-size: 22px;
+    font-weight: bold;
   }
 
   .progress-container {
     display: flex;
-    padding: 20px;
+    padding: 0.5em;
```
mklewandowski, line 18: Any reason to use em over px?  em is relative and is based on parent.  I use px since Michael's specs tend to be written that way.

emcelroy (approved):
Looks good 👍  One little nitpick is I would add another line break before the "_n_ out of _n_ questions are answered." sentence. Depending on the activity name, the sentence before it can wrap to two lines and you can end up with something like the below.

image

Also, this is probably something more to do with Activity Player in general, but as I was testing this out locally, my work was often not completely saved. It seemed like it was mostly the last question on a page where my response wasn't recorded. It may be because I was breezing through the questions and moving quickly between pages. Maybe the autosave didn't have time to save those last questions on a page? I don't know how likely that is to happen in the real world, but wanted to mention it somewhere.

src/components/activity-completion/completion-page-content.test.tsx
```diff
@@ -1,25 +1,25 @@
 import React from "react";
 import { CompletionPageContent } from "./completion-page-content";
-import IconCheck from "../../assets/svg-icons/icon-check.svg";
 import { shallow } from "enzyme";
+import {  Activity } from "../../types";
```
mklewandowski, line 4: Extra space sneaked in here.

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+  const { activity, activityName, onPageChange, showStudentReport, thumbnailURL,
+    sequence, activityIndex, onActivityChange, onShowSequence } = props;
+  
+  const [answers, setAnswers] = useState<any>();
+  let activityCompletionArray: boolean[]=[];
+
   const handleExit = () => {
-    onPageChange(0);
+    if (sequence) { onShowSequence?.(); }
+    else { onPageChange(0); }
```
mklewandowski, line 35: I would write this with line breaks for consistency:
```
    if (sequence) { 
      onShowSequence?.(); 
    } else { 
      onPageChange(0); 
    }
```

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
-  const progressText = isActivityComplete
-      ? `Congratulations! You have reached the end of ${quotedActivityName} and your work has been saved`
-      : `It looks like you haven't quite finished ${quotedActivityName} yet. The answers you've given have been saved.`;
-  const nextStepText = isActivityComplete
-      ? `You have completed ${quotedActivityName} and you may exit now.`
-      : `You haven't completed ${quotedActivityName} yet. You can go back to complete it, or you can exit.`;
+
+  const sequenceProgress = () => {
+    let activityStatus;
+    sequence?.activities.map((sequenceActivity)=>{
```
mklewandowski, line 47: spacing:
`sequence?.activities.map((sequenceActivity) => {`

src/components/app.tsx
```diff
@@ -303,7 +309,10 @@ export class App extends React.PureComponent<IProps, IState> {
       parameters: { new_activity_index: activityNum + 1, new_activity_name: this.state.sequence?.activities[activityNum].name }
     });
     this.setState((prevState) =>
-      ({ activity: prevState.sequence?.activities[activityNum], showSequence: false })
+      ({ activity: prevState.sequence?.activities[activityNum],
+        showSequence: false,
+        activityIndex: activityNum
```
mklewandowski, line 314: I think these two should be one more space in to line up.

src/components/app.tsx
```diff
@@ -48,6 +48,8 @@ interface IState {
   portalData?: IPortalData;
   sequence?: Sequence;
   showSequence?: boolean;
+  activityIndex?: number;
+  allActivitiesComplete?: boolean;
```
mklewandowski, line 52: `allActivitiesComplete` doesn't appear to be used.  Should it be deleted?

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+      });
+    }
+    return ({ numAnswers: answerNum, numQuestions: questionNum });
+  };
+
+  useEffect(() => {
+    watchAllAnswers(answerMetas => {
+      setAnswers(answerMetas);
+    });
+  },[]);
```
mklewandowski, line 86: spacing:
`}, []);`

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+
+  if(!answers) {
```
mklewandowski, line 88: spacing:
` if (!answers) {`

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+    );
+  }
+
+  const progress = activityProgress(activity);
+  const isActivityComplete = progress.numAnswers === progress.numQuestions;
+  const activityTitle = (activityName !== "") || (activityName == null) ? activityName : "the activity";
+  const completionText = activityName ? `${activityName} activity complete!` : "Activity complete!";
+  const activityNum = activityIndex ? activityIndex : 0;
+  const completedActivityProgressText =
+    `Congratulations! You have reached the end of ${activityTitle} and your work has been saved.`;
```
mklewandowski, line 106: Nitpicky, but shouldn't there be a comma after the activity title?

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+  const incompleteActivityProgressText =
+    `It looks like you haven't quite finished ${activityTitle} yet. The answers you've given have been saved.`;
+  const isLastActivityInSequence = activityIndex ? sequence?.activities.length === activityIndex + 1 : false;
+  const nextActivityTitle = !isLastActivityInSequence && sequence?.activities[activityNum + 1].name;
+  const nextActivityDescription = !isLastActivityInSequence && renderHTML(sequence?.activities[activityNum + 1].description || "");
+  const nextStepMainContentTitle = sequence ? (sequence.display_title !== "" ? sequence.display_title : "the sequence") : activityTitle;
+  const completedMainContentNextStepText = `You have completed ${nextStepMainContentTitle} and you may exit now.`;
+  const incompletedMainContentNextStepText =
```
mklewandowski, line 114: Maybe call this var `incompleteMainContentNextStepText`

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+  let nextStepText = "";
+
+  if (sequence) {
+    const sequenceComplete = sequenceProgress();
+    
+    if (isLastActivityInSequence) {
+      progressText = isActivityComplete
+        ? completedActivityProgressText + `You have completed all your work for this module!` : incompleteActivityProgressText;
+      nextStepText = sequenceComplete ? completedMainContentNextStepText : incompletedMainContentNextStepText;
+    } else { //if !isLastActivity
```
mklewandowski, line 126: Is this comment needed?

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+      progressText = isActivityComplete ? completedActivityProgressText : incompleteActivityProgressText;
+
+      nextStepText = "";
+    }
+  } else { //if (!sequence) Assumes is single activity
+    progressText = isActivityComplete ? completedActivityProgressText : incompleteActivityProgressText;
+    nextStepText = isActivityComplete ? completedMainContentNextStepText : incompletedMainContentNextStepText;
+
```
mklewandowski, line 134: probably don't need this line break.  There are a couple more above that I would remove too.

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+  }
   return (
     <div className="completion-page-content" data-cy="completion-page-content">
-      { isActivityComplete && <div className="completion-text" data-cy="completion-text">{completionText}</div> }
+      { isActivityComplete && <div className="completion-text" data-cy="completion-text">{completionText}</div>}
```
mklewandowski, line 138: I would keep consistent spacing.  If you're going to use one after curly brace, use one before the closing curly brace too.

mklewandowski: The one's below don't have the space so I would remove the one after the first curly brace.

src/components/app.tsx
```diff
@@ -339,5 +348,5 @@ export class App extends React.PureComponent<IProps, IState> {
   private handleLoadPlugins = () => {
     this.setState({ pluginsLoaded: true });
   }
-
+  
 }
```
mklewandowski, line 352: I think this module was saved without trailing spaces removed.  I would remove trailing spacing using the settings in VS Code.
https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
         <div className="next-step" data-cy="next-step">
-          <div>{nextStepText}</div>
+          <div data-cy="next-step-text">{nextStepText}</div>
+          <div className="progress-text">{`${progress.numAnswers} out of ${progress.numQuestions} questions are answered.`}</div>
+          {(sequence && !isLastActivityInSequence) && <div className="next">Next Up ...</div>}
+          {sequence && <div className="completion-text">{nextActivityTitle}</div>}
+          {sequence && <div>{nextActivityDescription}</div>}
+          { (!isLastActivityInSequence && sequence) &&
+            <span>
+              <button className="button" onClick={handleNextActivity}>Start Next Activity</button>
```
mklewandowski, line 159: I just wanted to say that I'm happy to see these as buttons!  Accessibility!

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+  if (sequence) {
+    const sequenceComplete = sequenceProgress();
+    
+    if (isLastActivityInSequence) {
+      progressText = isActivityComplete
+        ? completedActivityProgressText + `You have completed all your work for this module!` : incompleteActivityProgressText;
+      nextStepText = sequenceComplete ? completedMainContentNextStepText : incompletedMainContentNextStepText;
+    } else { //if !isLastActivity
+      progressText = isActivityComplete ? completedActivityProgressText : incompleteActivityProgressText;
+
```
mklewandowski, line 128: line break not needed

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+  const nextStepMainContentTitle = sequence ? (sequence.display_title !== "" ? sequence.display_title : "the sequence") : activityTitle;
+  const completedMainContentNextStepText = `You have completed ${nextStepMainContentTitle} and you may exit now.`;
+  const incompletedMainContentNextStepText =
+    `You haven't completed ${nextStepMainContentTitle} yet. You can go back to complete it, or you can exit.`;
+  let progressText = "";
+  let nextStepText = "";
+
+  if (sequence) {
+    const sequenceComplete = sequenceProgress();
+
```
mklewandowski, line 121: line break not needed

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+      activityCompletionArray = [...activityCompletionArrayCopy];
+    });             
+    if (activityCompletionArray?.length === sequence?.activities.length && !activityCompletionArray?.includes(false)) {
+      return true;
+    } else {
+      return false;
+    }
+  };
+
+  const activityProgress = (kActivity: Activity) => {
```
mklewandowski, line 61: I wouldn't call this `kActivity` since the k tends to be used for a constant.  I would use `currentActivity`.

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+    let answerNum = 0, questionNum = 0, questionId: string;
+    let i = 0;
+    for (i = 0; i < kActivity.pages.length; i++) {
+      kActivity.pages[i].embeddables.map((embeddableWrapper: EmbeddableWrapper) => {
```
mklewandowski, line 65: maps are probably not appropriate here since you're not really mapping the array to something new.  You're really searching the array and iterating something external.

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+            // after response has been saved
+            if (answer.meta.question_id === questionId) { 
+              answerNum++;
+            }
+          });
+        }
+      });
+    }
+    return ({ numAnswers: answerNum, numQuestions: questionNum });
+  };
```
mklewandowski, line 80: This might be a cleaner way to write the activityProgress function:
```
  const activityProgress = (currentActivity: Activity) => {
    let numQuestions = 0;
    let numAnswers = 0;
    currentActivity.pages.forEach((page: Page) => {
      page.embeddables.forEach((embeddableWrapper: EmbeddableWrapper) => {
        if (isQuestion(embeddableWrapper)) {
          numQuestions++;
          const questionId = refIdToAnswersQuestionId(embeddableWrapper.embeddable.ref_id);
          if (answers.find((answer: any) => answer.meta.question_id === questionId)) {
            numAnswers++; // doesn't take into account is user erases response after saving
          }
        }
      });
    });
    return ({ numAnswers, numQuestions });
  };
```

I need to test it a little more, but I think it simplifies things.

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+  onShowSequence?: () => void;
 }
 
 export const CompletionPageContent: React.FC<IProps> = (props) => {
-  const { activityName, isActivityComplete, onPageChange, showStudentReport, thumbnailURL } = props;
+  const { activity, activityName, onPageChange, showStudentReport, thumbnailURL,
+    sequence, activityIndex, onActivityChange, onShowSequence } = props;
+  
+  const [answers, setAnswers] = useState<any>();
+  let activityCompletionArray: boolean[]=[];
```
mklewandowski, line 31: spacing:
`let activityCompletionArray: boolean[] = [];`

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+    return (
+      <div className="completion-page-content" data-cy="completion-page-content">
+        <div className="progress-container" data-cy="progress-container">
+          <div className="progress-text">
+            Fetching your data ...
+          </div>
+        </div>
+      </div>  
+    );
+  }
```
mklewandowski, line 98: It's a little odd to have these two returns at different parts in the function.  I would try to use a single return with a ternary and even consider putting this piece above in its own small functional component to clean things up

src/components/activity-completion/completion-page-content.tsx
mklewandowski: Also I would scope this inside of the function that uses it.  Doesn't seem to be a need to declare it here.

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -1,50 +1,165 @@
...
+      const activityCompletionArrayCopy = [...activityCompletionArray];
+      activityCompletionArrayCopy?.push(sequenceActivityComplete);
+      activityCompletionArray = [...activityCompletionArrayCopy];
+    });             
+    if (activityCompletionArray?.length === sequence?.activities.length && !activityCompletionArray?.includes(false)) {
+      return true;
+    } else {
+      return false;
+    }
+  };
```
mklewandowski, line 59: I need to test a little more, but I think this is a much cleaner implementation of the sequenceProgress function (need to pass in the sequence):
```
  const sequenceProgress = (currentSequence: Sequence) => {
    const activityCompletionArray = currentSequence?.activities.map((sequenceActivity) => {
      const activityStatus = activityProgress(sequenceActivity);
      return activityStatus.numAnswers === activityStatus.numQuestions;
    });
    return !activityCompletionArray.includes(false);
  };
```

mklewandowski (changes requested):
Looking good.  I think there is a little cleanup needed and then we're ready to roll:
- cleanup in the CompletionPageContent component based on comments below
- remove allActivitiesComplete from app state 
- check spacing in a handful of modules
- check trailing whitespace
- check em vs. px (maybe not important since Michael will get to this eventually)

src/components/activity-completion/completion-page-content.scss
eireland: fixed

mklewandowski (approved):
👍 Awesome!  LGTM!

emcelroy:
@eireland Did you see my comment about the line break. Doesn't look like that was fixed. Is there a reason you don't think it should be that way?

PR #95, opened 2020-12-17
175823492 interactive decoration

mklewandowski:
This PR adds the ability to send and receive text decoration messages to and from interactives.  Changes include:
- add sample data to work with example-interactive
- add mobx observerable to keep track of text decoration info received from glossary plugin
- send and receive text decoration messages in iframe runtime component, call glossary plugin functions to handle events from decorated text

NOTE: this PR will need to be updated to point to a new version of the glossary plugin.

This PR is related to the Glossary Plugin, LARA, and QI PRs below:
https://github.com/concord-consortium/glossary-plugin/pull/79
https://github.com/concord-consortium/lara/pull/664
https://github.com/concord-consortium/question-interactives/pull/79


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -139,6 +148,20 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
       addListener("hint", (newHint: any) => {
         setNewHint(newHint.text || "");
       });
+      addListener("decoratedContentEvent", (msg: IDecoratedContentEventMessage) => {
+        if (textDecorationInfo && msg.type === "click") {
+          if ("type" in textDecorationInfo.eventListeners && textDecorationInfo.eventListeners.type === "click") {
```
scytacki, line 153: Any reason not to just pass the `type` through here? So if new types are added this code doesn't need to be updated.

src/lara-plugin/plugin-api/decorate-content.ts
```diff
@@ -1,8 +1,28 @@
...
+  words: [],
+  replace: "",
+  wordClass: "",
+  eventListeners: [],
+});
 
 export interface IEventListener {
   type: string;
-  listener: (evt: Event) => void;
+  listener: (evt: Event | IPluginEvent) => void;
```
scytacki, line 25: I'd recommend making this more explicitly typed. So it would be more like a `IDecoratedContentEventListener` and `IDecoratedContentEvent`.  
I don't think you need to keep the generic `Event` option since this code is living in the Activity Player it doesn't need to support sending these old style events.

The place where the backward compatibility is in the glossary plugin and the LARA plugin api types in the LARA codebase which the glossary plugin is using.

scytacki:
While testing this locally I did find another potential problem.

If the IFrameRuntime is rendered before the plugin loads, then when the plugin makes a call to decorate the content, this doesn't trigger a message being sent to the interactive.

I see this when I load an activity that has a single example-interactive in it. When I have 6 example interactives, sometimes the first one doesn't get the message but the other 5 do.

This issue might just be an artifact of my development environment, but looking at the code it doesn't seem like it is handling this case. I don't see how changes to `textDecorationInfo` would trigger the `phoneRef.current.post("decorateContent", textDecorationMessage);` to be called. The IFrameRuntime component could be marked as an observer to fix this, but that would trigger a full iframe reload whenever the `textDecorationInfo` changed.  

I don't see any place that is observing the `textDecorationInfo`, so as written it could probably just be a plain javascript object.

I'm going to try to fix this by putting in an autorun here: `textDecorationInfo` within the useEffect: https://github.com/concord-consortium/activity-player/blob/b23eefdc393dc7a72dd58bb09be3ac080f1bb062/src/components/activity-page/managed-interactive/iframe-runtime.tsx#L58-L64 

That way just the new message would be sent instead of reloading the whole iframe.

mklewandowski:
@scytacki hmm... ManagedInteractive was intended to be an observer of textDecorationHandlerInfo, but I struggled with the Typescript syntax so perhaps it's set up incorrectly.  Ideally, iframe runtime should be an observer of textDecorationHandlerInfo.  Again, the change to wrap iframe runtime in a forwardRef caused difficulty in wrapping the component in a mobx observer.  I'm not familiar with the recent spate of changes to iframe runtime so I'm not entirely sure what some of the new changes are for.  I wanted to be cautious not to break them.

mklewandowski:
Also the issues that you raise about reloading are concerning.  The interactives are already painfully slow to load so we can't, at this point, do anything that slows them down further or requires additional re-renders.

mklewandowski:
Yup.  Not working.  I'm not understanding how to integrate a mobx observer into a typed function component.

mklewandowski:
I'm lost.  From the documentation here:
https://mobx.js.org/react-integration.html
"For observer to work, it doesn't matter how the observables arrive in the component, only that they are read."

Managed interactive import the observable and then passes it down via props to iframe runtime.

mklewandowski:
@scytacki @kswenson The last commit has changes that Kirk and I paired on that fix the mobx observer.  See if this works for you if you have time.

scytacki:
Nice the autorun looks good. I see you also returned the disposer so the autorun function gets cleaned up, nice. 👍

src/components/activity-page/managed-interactive/iframe-runtime.tsx
mklewandowski: Yes, good point.  As long as the `type` in the message matches a `type` in our stored listener, then we can pass it on.  Fixed.

src/lara-plugin/plugin-api/decorate-content.ts
mklewandowski: Don't these need to sync up with these types?
https://github.com/concord-consortium/glossary-plugin/blob/bd3d9e1d2f2ab22b76c480ed29adaf58add4ea23/src/components/plugin/plugin-app.tsx#L302
And isn't the existing implementation of `IEventListener ` used below in `decorateContent`?

I don't understand what we're trying to accomplish here.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -139,6 +160,21 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
       addListener("hint", (newHint: any) => {
         setNewHint(newHint.text || "");
       });
+      addListener("decoratedContentEvent", (msg: IDecoratedContentEvent) => {
+        const { textDecorationHandlerInfo } = pluginInfo;
+        if (textDecorationHandlerInfo) {
+          if ("type" in textDecorationHandlerInfo.eventListeners && textDecorationHandlerInfo.eventListeners.type === msg.type) {
```
kswenson, line 166: You could combine the array and non-array paths with something like:
```typescript
          const listeners = Array.isArray(textDecorationHandlerInfo.eventListeners)
							  ? textDecorationHandlerInfo.eventListeners
							  : [textDecorationHandlerInfo.eventListeners];
		  listeners.forEach(listener => ...
```
there's a lodash utility called `castArray` that will do the same thing.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -224,6 +260,15 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
     // The _only_ time that's ever appropriate is when the url has changed.
   }, [url]);  // eslint-disable-line react-hooks/exhaustive-deps
 
+  useEffect(() => {
+    return autorun(() => {
+      if (phoneRef.current && iframeRef && pluginInfo.textDecorationHandlerInfo) {
```
kswenson, line 265:
```suggestion
      if (phoneRef.current && pluginInfo.textDecorationHandlerInfo) {
```
By the same argument as above, I don't think the `iframeRef` test is accomplishing anything here and removing it would presumably allow it to be removed from the dependencies below as well.

Given the lack of other dependencies, are we guaranteed that `phoneRef.current` will be set before the `useEffect` runs the first time? Without other dependencies it will never run again.

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -34,7 +36,7 @@ const getModalContainer = (): HTMLElement => {
   return document.getElementById("app") || document.body;
 };
 
-export const ManagedInteractive: React.FC<IProps> = (props) => {
+export const ManagedInteractive: React.FC<IProps> = observer(({ ...props }) => {
```
kswenson, line 39: It's not clear to me what we're observing here. I believe `pluginInfo` is the only MobX observable around, but it doesn't appear to be referenced explicitly below. It is referenced by the `<IframeRuntime>` component, but that wouldn't explain why this component needs to rerender when it changes. If this is necessary for some non-obvious reason a comment would be useful.

kswenson (changes requested):
👍 Looks good -- just a couple of suggested tweaks and a request for clarification.

src/components/activity-page/managed-interactive/managed-interactive.tsx
mklewandowski: I thought that got ripped out when we worked on the observable - must have been overlooked.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -148,7 +148,7 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
       addListener("hint", (newHint: any) => {
         setNewHint(newHint.text || "");
       });
-      addListener("decoratedContentEvent", (msg: IDecoratedContentEventMessage) => {
+      addListener("decoratedContentEvent", (msg: IDecoratedContentEvent) => {
         if (textDecorationInfo && msg.type === "click") {
           if ("type" in textDecorationInfo.eventListeners && textDecorationInfo.eventListeners.type === "click") {
```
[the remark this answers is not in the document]
mklewandowski: I think this was fixed in a previous commit.  But I agree, we should handle more than just the click.

src/components/activity-page/managed-interactive/managed-interactive.tsx
mklewandowski: Fixed.

package.json
```diff
@@ -148,6 +148,8 @@
     "jquery": "^3.5.1",
     "jquery-ui": "^1.12.1",
     "jsonwebtoken": "^8.5.1",
+    "mobx": "^6.0.4",
+    "mobx-react": "^7.0.5",
```
kswenson, line 152: Given the recent removal of the `observer`, do we even need `mobx-react` at this point?

src/components/activity-page/managed-interactive/iframe-runtime.tsx
mklewandowski: Agreed.  This is cleaner.  Fixed.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
mklewandowski: I removed the `iframeRef` dependency, not sure if it needs some other dependency.  My tests cases are all working, but we need to keep an eye on this,

package.json
mklewandowski: Agreed.  Removed.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -148,7 +148,7 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
...
             textDecorationInfo.eventListeners.listener({ type: "click", text: msg.text });
```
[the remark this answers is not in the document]
mklewandowski: Ah, yes.  Fixed.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -224,6 +258,15 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
...
   }, [url]);  // eslint-disable-line react-hooks/exhaustive-deps
 
+  useEffect(() => {
+    return autorun(() => {
+      if (phoneRef.current && pluginInfo.textDecorationHandlerInfo) {
+        const textDecorationInfo: ITextDecorationInfo = createTextDecorationInfo();
+        phoneRef.current.post("decorateContent", textDecorationInfo);
+      }
+    });
+  });
```
kswenson, line 268:
```suggestion
  }, []);
```
I think you still need the empty dependencies array here -- otherwise the effect will run on every render.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -75,6 +92,10 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
         if (iframeRef.current) {
           setSupportedFeatures(iframeRef.current, features);
         }
+        if (phoneRef.current && iframeRef && pluginInfo.textDecorationHandlerInfo) {
```
kswenson, line 95:
```suggestion
        if (phoneRef.current && pluginInfo.textDecorationHandlerInfo) {
```
I thought I commented on this before, but I don't think this `iframeRef` test is accomplishing anything here.

kswenson (approved):
👍 Changes look good -- one additional tweak that I thought I commented on before but now I can't find it so maybe it got lost in the ether.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
mklewandowski: Fixed.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
mklewandowski: Ah, yes.  Fixed.

mklewandowski:
Hoping to merge this PR.  Is there final work needed here?  If not, I will merge tomorrow in anticipation of demo/planning.

kswenson:
>@scytacki @kswenson Hoping to merge this PR. Is there lingering work needed here? If not, I will merge tomorrow in anticipation of demo/planning.

Looks good to me, as evidenced by my prior approval.

mklewandowski:
I'm going to merge.  We can continue to improve in other PRs if needed.

PR #96, opened 2020-12-22
174046971 logged in user report

eireland:
## Testing updates
- Update Cypress so we can use cy.intercept to mock network requests.
- Update anonymous runKey tests so it reloads the browser page when checking for saved state, otherwise it isn't really checking that the data is saved to firestore.
- Get the Cypress anonymous report button test working using cy.intercept
- Add Jest test of the report-utils.
- Split getReportUrl out of showReport so it is easier to test

## Code updates
- Expose the portalData which is downloaded by the app at the beginning so the report button can use this later
- When there is no runKey generate a reportLink using properties from portalData

This depends on this PR in portal-report: https://github.com/concord-consortium/portal-report/pull/280


scytacki:
@eireland I think this is ready now. You can't be added as a reviewer, but it'd be good if you look at what I updated.

mklewandowski (approved):
This looks good to me.  My only question is if there is anything else that can be added to the readme (I couldn't think of anything of the top of my head).  I just want to make sure that we continue to document how to run the AP in different contexts.

scytacki:
The readme is a good point. I think it is worth adding something about how to locally test the report buttons.

scytacki:
@mklewandowski how is that readme change? 
Is it too detailed and should instead be moved to a separate doc?

mklewandowski:
> @mklewandowski how is that readme change?
> Is it too detailed and should instead be moved to a separate doc?

Nice!  I like seeing that kind of detail.  I vote to keep this in the readme.

eireland (commented):
@scytacki This looks good. All that to test that we get the right URL from activity report?

scytacki:
Yeah it is a lot code to test something that seems pretty simple. But I think it is worth it since this the integration point with the portal report, and it is good to test and document these apis.

I was hoping that the cypress test could also be used for the logged in user report button, but that was harder.

scytacki:
Because I got burned the last time I merged a PR without updating it to the latest master. I've just merged master into this PR and will wait for it to run the tests before I merge this PR.

PR #97, opened 2020-12-28
Add page linking using page id [#176197155]

dougmartin:
NOTE: this relies on, but does not require, changes to LARA in this PR: https://github.com/concord-consortium/lara/pull/667 (the PR adds the page id to the page structure in the exported sequence and activity json)


src/utilities/activity-utils.ts
```diff
@@ -144,3 +144,20 @@ export const setDocumentTitle = (activity: Activity | undefined, pageNumber: num
...
+      if (page.id === pageId) {
+        return page.position;
+      }
+    }
+
+    // default to index page when id not found
+    return 0;
+  }
+
+  return parseInt(pageQueryValue, 10) || 0;
```
mklewandowski, line 162: This is probably a good place to check if the page number specified in the URL param is also less than 0 or greater than `activity.pages.length`.  We should probably return 0 in those cases as well.

mklewandowski (approved):
👍 Looks good.  If it's not too much trouble, I would add a line about how to use the page id in the URL param section of the readme and check if the page number is less than 0 or greater than the number of activity pages in `getPagePositionFromQueryValue `.

dougmartin:
@mklewandowski I've added range check on the page parameter: https://github.com/concord-consortium/activity-player/pull/97/commits/4c992a9bd09b48df0e9e76a40b0e7efaa96009bf

dougmartin:
@mklewandowski I also updated the readme and fixed a lint error introduced in my earlier commit.

scytacki (changes requested):
Seems good to me. It is nice to support both a page index and a page id.

Can you document better what `page_[id]` maps to in the activity export file. Is the id in the file `page_123` or is the id in the file `123`?  

Some of these files are manually created instead just being exported from LARA so it is useful to document this.
Since the way the ids in the exported

dougmartin:
@scytacki is this a better description of the page id parameter? https://github.com/concord-consortium/activity-player/pull/97/commits/e7e083dc88e55a44f8628c0ed83b81c414c2663c

dougmartin:
@scytacki I fixed a build error with a new json file that was added - is this good to merge?

scytacki (approved):
This is good. Sorry I didn't see this sooner.

Next time if you 're-request' a review I'll see it sooner. That is the little recycle icon next to the reviewer names.

PR #98, opened 2021-01-05
176275541 navigation style

mklewandowski:
This PR updates the page navigation toolbar that appears at the top and bottom of each activity page to match the UI spec.

image


emcelroy (approved):
Looks good!

PR #99, opened 2021-01-05
176275540 header style

mklewandowski:
This PR updates the page header to match the UI spec.

image


emcelroy (approved):
Looks good 👍 

I had to Google "min-width: 0" to see why you used that for `.header-center` 🙂 Interesting.

mklewandowski:
> Looks good 👍
> 
> I had to Google "min-width: 0" to see why you used that for `.header-center` 🙂 Interesting.

It was to solve this annoyance that comes about from the flex-containers:
https://stackoverflow.com/questions/41674979/flex-child-is-growing-out-of-parent

PR #101, opened 2021-01-06
176371301 accessibility improvements

mklewandowski:
In response to Leslies request during demo/planning, I used the Chrome Lighthouse feature in dev tools to test accessibility in the Activity Player.  This PR fixes a few small issues that I found to get us to 100% accessibility on the intro, internal, and completion pages:
- add alt attributes to image elements
- add title attributes to iframe elements
- add aria-label attributes to button elements that do not have clear text
- adjust colors to meet contrast requirements (these will probably be added to the spec at some point, but this makes a few small fixes now in case we don't get to it in the near future)
- fix the content description in the page meta tag


eireland (approved):
Looks good to me!

PR #102, opened 2021-01-06
176275540 sequence header

mklewandowski:
This PR updates the header portion of the application when displaying sequences.  Changes includes:
- remove sequence logo from header in sequence intro page
- add icon next to sequence name in header in sequence intro page
- when viewing an activity in a sequence, header, show the sequence name with sequence icon next to it (instead of activity name).  Clicking on sequence name returns user to sequence intro page.
- when viewing an activity in a sequence, a section appears below the header that contains a dropdown menu allowing users to navigate to any activity in the sequence.
- loading just an activity should result in no change in behavior

Can be demoed/tested here:
https://activity-player.concord.org/branch/sequence-header/?preview&sequence=sample-sequence

image

### DIFF
--- a/src/components/activity-header/header.scss
+++ b/src/components/activity-header/header.scss
@@ -7,7 +7,7 @@
   font-size: 18px;
   font-weight: bold;
 
-  &.no-margin {
+  &.in-sequence {
     margin-bottom: 0;
   }
 
--- a/src/components/activity-header/header.scss
+++ b/src/components/activity-header/header.scss
@@ -7,6 +7,10 @@
   font-size: 18px;
   font-weight: bold;
 
+  &.no-margin {
+    margin-bottom: 0;
+  }
+
   .inner {
     display: flex;
     flex-direction: row;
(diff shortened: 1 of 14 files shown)


src/components/activity-header/header.scss
```diff
@@ -7,6 +7,10 @@
   font-size: 18px;
   font-weight: bold;
 
+  &.no-margin {
```
emcelroy, line 10: It seems like `no-margin` is only used for the activity header when a sequence name is shown. If that's true, would it be slightly better to use a class name that's more indicative of that? Something like `show-sequence` or `in-sequence`? Or do you foresee this class possibly being used for other purposes?

emcelroy (approved):
This looks good to me! I left one picky inline comment about a class name. I don't think it's very important though.

src/components/activity-header/header.scss
mklewandowski: Good call.  More semantic this way. Fixed.

PR #103, opened 2021-01-11
Request interactive states when user tries to change page

pjanik:
[#176178847]

The `App` component uses refs and imperative APIs to request all the interactive states just before the current page number is updated. The final call is executed by iframe-runtime. The page number is updated only after all the interactives respond or after 2 seconds when a promise gets rejected. The user will see an error message in this case. Usually, this shouldn't happen, and hopefully will be caught by authors. The easiest way to trigger it is to check "save state" on the interactive that actually doesn't save state. Otherwise, Interactive API Client handles that automatically for interactives, so all the new ones should be safe. Generally, this feature should work the same as LARA + timeout support + error message.

### DIFF
--- a/src/components/activity-header/nav-pages.tsx
+++ b/src/components/activity-header/nav-pages.tsx
@@ -15,7 +15,25 @@ interface IProps {
   lockForwardNav?: boolean;
 }
 
-export class NavPages extends React.PureComponent <IProps> {
+interface IState {
+  pageChangeInProgress: boolean;
+}
+
+export class NavPages extends React.Component <IProps, IState> {
+  constructor(props: IProps) {
+    super(props);
+    this.state = {
+      pageChangeInProgress: false
+    };
+  }
+
+  componentDidUpdate(prevProps: IProps) {
+    if (prevProps.currentPage !== this.props.currentPage) {
+      // Unlock buttons after page has been changed.
+      this.setState({ pageChangeInProgress: false });
+    }
+  }
+
   render() {
     return (
       <div className="nav-pages" data-cy="nav-pages">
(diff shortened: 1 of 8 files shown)


src/components/activity-header/nav-pages.tsx
```diff
@@ -102,6 +124,9 @@ export class NavPages extends React.PureComponent <IProps> {
   }
 
   private handleChangePage = (page: number) => () => {
-    this.props.onPageChange(page);
+    if (!this.state.pageChangeInProgress) {
+      this.props.onPageChange(page);
+      this.setState({ pageChangeInProgress: true });
```
dougmartin, line 129: Do you see any issue with setting the state after you call the callback versus calling it in the `setState` callback like:

`this.setState({ pageChangeInProgress: true }, () => this.props.onPageChange(page))`;

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -68,6 +70,12 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
   const setInteractiveStateRef = useRef<((state: any) => void)>(setInteractiveState);
   setInteractiveStateRef.current = setInteractiveState;
   const linkedInteractivesRef = useRef(linkedInteractives?.length ? { linkedInteractives } : { linkedInteractives: [] });
+  // const interactiveStateRequestPromise = useRef<Promise<void>>();
```
dougmartin, line 73: Did you mean to leave these two commented out `const`s in the code?

dougmartin (approved):
Looks good :+1: Just a couple of questions inline.

btw, this is the first time I've seen forward refs used.  Works well.

src/components/activity-header/nav-pages.tsx
pjanik: To be super safe that callback can't be executed twice? 
I think it won't hurt. Some micro delay maybe, but I don't think it'd be noticeable by humans. ;) 
I made this change in the last comment, thanks.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
pjanik: No, fixed, thanks.

dougmartin:
Updated commit looks good :+1:

kswenson (approved):
👍 Looks good!

PR #104, opened 2021-01-13
176472992 sequence intro

mklewandowski:
This PR updates the sequence intro page and activity headers when an activity header is a part of a sequence.  Changes include:
- update intro text and text style (font size, weight, color) on sequence intro page to match spec
- update thumbnails on sequence intro to match spec
- update padding and margins on sequence intro page to match spec
- update text, back button icon and button states of the sequence title button in the activity header to match spec
- update activity selection dropdown menu to match spec

image

image

### DIFF
--- a/src/assets/svg-icons/arrow-back-icon.svg
+++ b/src/assets/svg-icons/arrow-back-icon.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="#0592AF" >
+    <path d="M15.67 3.87L13.9 2.1 4 12 13.9 21.9 15.67 20.13 7.54 12z" transform="translate(-340 -173) translate(0 50) translate(40 103) translate(290 10) translate(10 10)"/>
+</svg>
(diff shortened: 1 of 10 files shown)


emcelroy (approved):
This looks good to me 👍

This is most likely out of scope of this work, but was there any discussion/thought about whether the activity images on the sequence intro page should appear at a uniform height and width even if the images themselves don't have uniform aspect ratios? Seeing everything else lined up better makes me want those images to line up with each other better. 

I think it's often the case that authors don't realize using images with varying aspect ratios will result in that lack of consistency. So I think it'd be good if there could be a little magic done to ensure that the images are displayed at the same size and aspect ratio, even if it means not showing the entire image in some cases.

mklewandowski:
> This looks good to me 👍
> 
> This is most likely out of scope of this work, but was there any discussion/thought about whether the activity images on the sequence intro page should appear at a uniform height and width even if the images themselves don't have uniform aspect ratios? Seeing everything else lined up better makes me want those images to line up with each other better.
> 
> I think it's often the case that authors don't realize using images with varying aspect ratios will result in that lack of consistency. So I think it'd be good if there could be a little magic done to ensure that the images are displayed at the same size and aspect ratio, even if it means not showing the entire image in some cases.

I agree.  I think that more work should be done to get the image sizing to look nicer.  However, I think this is a bigger piece of work that merits discussion before diving into it.  Do we want to try to zoom images up if they are small?  Do we want things to line up vertically?  At present, we try to set the same width and crop vertically.  Do some folks like that?  I think there are a variety of ways we can handle this.  For now, I'm going to merge this PR.  I'm going to write a new story about the images.

PR #107, opened 2021-01-13
remove orange bar

mklewandowski:
Remove orange warning bar from the activity player.

### DIFF
--- a/src/components/app.scss
+++ b/src/components/app.scss
@@ -18,7 +18,7 @@
   align-items: center;
   width: 100%;
   height: fit-content;
-  padding: 0 20px 20px 20px;
+  padding: 40px;
   line-height: 1.35;
   color: $cc-charcoal;
   font-size: 18px;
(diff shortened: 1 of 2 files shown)


scytacki:
I don't think we should merge this until we have the report service updated so we can actually save production data. 

Right now someone can use the activity-player in production, but their data will be saved to the `report-service-dev` firestore database. That'd be bad is someone actually used this with students and their data was in the dev database.

mklewandowski:
> I don't think we should merge this until we have the report service updated so we can actually save production data.
> 
> Right now someone can use the activity-player in production, but their data will be saved to the `report-service-dev` firestore database. That'd be bad is someone actually used this with students and their data was in the dev database.

I'm also wondering about the plugin links (less impactful).  Glossary is still pointing at master and I was going to move TE to do the same (to make dev easier), but if we're moving to production, both plugins should point to stable, release versions.

scytacki:
Yeah good point about the plugin links, we have a story in Orange to deal with the plugin links. It is currently scheduled before the production release.

mklewandowski:
> Yeah good point about the plugin links, we have a story in Orange to deal with the plugin links. It is currently scheduled before the production release.

Good to know.  I won't touch them in the meantime.

kswenson (approved):
👍 LGTM

eireland (approved):
Looks good. It was nice to see it without the warning banner!

PR #109, opened 2021-01-14
176503551 fix page links

mklewandowski:
This PR updates the pages names in the page links in the activity home page to show the actual page name (if it exists) instead of showing "Page N" (and includes handling of long page names).  

This also includes a small tweak to remove the ":" in the internal page when there is no page name so we show, for example, "Page 3" instead of "Page 3:".

image


emcelroy (approved):
Looks good!

PR #110, opened 2021-01-14
172877789 te mc wrapper

mklewandowski:
This PR includes minor updates to the custom message callback which is needed to send messages from the TE plugin to the MC Library Interactive.  Also includes updates to the sample data to make testing a wrapped MC question easier.

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -246,7 +246,8 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
       // Re-init interactive, this time using a new mode (report or runtime).
       phoneRef.current = new iframePhone.ParentEndpoint(iframeRef.current, initInteractive);
       setSendCustomMessage((message: ICustomMessage) => {
-        phoneRef.current?.post(message.type, message.content);
+        // TODO: structure message type and content correctly
+        phoneRef.current?.post("customMessage", message);
       });
     }
     // Cleanup.
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -246,7 +246,6 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
       // Re-init interactive, this time using a new mode (report or runtime).
       phoneRef.current = new iframePhone.ParentEndpoint(iframeRef.current, initInteractive);
       setSendCustomMessage((message: ICustomMessage) => {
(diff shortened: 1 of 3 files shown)


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -246,7 +246,8 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
       // Re-init interactive, this time using a new mode (report or runtime).
       phoneRef.current = new iframePhone.ParentEndpoint(iframeRef.current, initInteractive);
       setSendCustomMessage((message: ICustomMessage) => {
-        phoneRef.current?.post(message.type, message.content);
+        // TODO: structure message type and content correctly
+        phoneRef.current?.post("customMessage", message);
```
mklewandowski, line 250: My understanding was that we post "customMessage" and the message itself contains a `type` and optional `content`.  In the TE case that we've been working on, the `type` on the `message` would specify showing/hiding overlays and the `content` could be empty.  Do I have this correct or am I not understanding the message structure?

kswenson (approved):
👍 LGTM

src/components/activity-page/managed-interactive/iframe-runtime.tsx
scytacki: Yes that is how I thought it should work.

scytacki (changes requested):
This looks good to me.
I'd remove the `TODO` since I don't think there is anything needed there.

My only concern is that it seems there is no test that is actually triggering the `post("customMessage", message)`

It looks like that iframe-runtime.tsx file is covered some by cypress and some by jest.
I don't know the activity player and its test well enough to know if this is something easy to cover or if it would be hard.

mklewandowski:
> This looks good to me.
> I'd remove the `TODO` since I don't think there is anything needed there.
> 
> My only concern is that it seems there is no test that is actually triggering the `post("customMessage", message)`
> 
> It looks like that iframe-runtime.tsx file is covered some by cypress and some by jest.
> I don't know the activity player and its test well enough to know if this is something easy to cover or if it would be hard.

I removed the TODO comment - I put that in only to make sure we looked at it before merging.  Not sure on the test.  I would have assumed any needed tests were in place in past PRs.  Not sure exactly how we would test that.  Will require more thought.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
mklewandowski: Thanks.  I removed the now obsolete comment as well.  I would vote that we merge this so I can move on to other work.

scytacki (approved):
Sounds good.

PR #112, opened 2021-01-15
Removed hardcoded plugin urls [#176481121]

dougmartin:
The plugin urls are now found in the LARA export data.

### DIFF
--- a/src/components/activity-page/embeddable.tsx
+++ b/src/components/activity-page/embeddable.tsx
@@ -54,7 +54,7 @@ export const Embeddable: React.ForwardRefExoticComponent<IProps> = forwardRef((p
       wrappedEmbeddable: embeddable,
       wrappedEmbeddableContainer: embeddableDivTarget.current || undefined,
       sendCustomMessage,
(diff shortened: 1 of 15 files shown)


scytacki (approved):
This looks good. 

I think there is more work to do in the future to remove the sidetip, and glossary-plugin components which seem specific to the teacher edition and glossary plugins.
And the `approvedScriptLabel` properties which are specific to those plugins.

I made a story about that work: https://www.pivotaltracker.com/story/show/176579437

scytacki:
I guess we need to be careful about merging this. 

Since most people are testing with activity player master. And this change requires lara to be updated. So I guess we need to wait until lara 1.59.0 is on production before merging this.

scytacki:
I've update authoring.concord.org to use the production branch of activity-player. 
authoring.staging.concord.org is using is using activity-player master and has 1.59.0 which includes the plugin exporting code. So it is **now safe to merge this**.

However there is a now a conflict with one of the demo activities.

dougmartin:
@scytacki I've rebased and fixed the conflicting json file.

PR #113, opened 2021-01-19
Add custom tinymce classes defined in Lara [#176540074]

dougmartin:
Adds tinymce-img-float-left and tinymce-img-float-right classes.  These are the only custom tinymce classes defined in Lara.


mklewandowski (approved):
Looks good!

PR #114, opened 2021-01-19
Remove page number

pjanik:
See: https://www.pivotaltracker.com/story/show/176540257


mklewandowski (approved):
👍 Looks good!  Curious where this request came from.  I'm assuming we put these in intentionally in the original work.  Are they no longer wanted?

pjanik:
Not sure, but @lbondaryk was the author of the PT story and seemed confident about it during the demo today.

mklewandowski:
> Not sure, but @lbondaryk was the author of the PT story and seemed confident about it during the demo today.

Got it!  Thanks!

PR #115, opened 2021-01-20
Fix lightbox sizing

pjanik:
[#176540138]

The previous logic was slightly broken.

Example case that was failing before:
- image 2000px x 2000px
- window 1000px x 500px
- allowUpscale = false (so, "original dimension" option in the Image Interactive)

First `if` was setting height to 500px and width to 500px.
That would be fine, but there was another `if` that was checking width independently.
So, it'd set the final width to 1000px and height to 1000px following image aspect ratio.
So, the image would be taller than the window and the close "X" icon won't be visible.

New logic just calculates aspect ratios first and uses pretty clear logic to check for constraining dimension.


kswenson (approved):
👍 Looks good -- much clearer as well.

PR #116, opened 2021-01-20
Fixed type error found when lara-interactive-api was updated to add built types

dougmartin:
Also removes lint error about unused variable introduced in a recent commit


mklewandowski (approved):
👍 Looks good!

PR #117, opened 2021-01-20
Save answers even when there's no authored state

pjanik:
[#176576965]

Some interactives don't define any authored state.

Note that I changed how the "version" property is set in the report state. It's no longer taken from the authored state. I think it's a separate version / JSON format. Currently, it's a JSON with mode, authoredState, and interactiveState properties. It doesn't seem related to authored state format version that is interactive specific.

### DIFF
--- a/src/components/activity-page/managed-interactive/managed-interactive.tsx
+++ b/src/components/activity-page/managed-interactive/managed-interactive.tsx
@@ -48,6 +48,8 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
   const shouldWatchAnswer = isQuestion(props.embeddable);
   const [loading, setLoading] = useState(shouldWatchAnswer);
 
+  console.log(props.embeddable);
+
   const embeddableRefId = props.embeddable.ref_id;
   useEffect(() => {
     if (shouldWatchAnswer) {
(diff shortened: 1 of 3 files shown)


pjanik:
I've just pushed a second, unrelated small change - `url_fragment` is now appended to iframe URLs. This property can be set by authors. Some interactives rely on query parameters for authoring and they didn't work in Activity Player.
LARA was appending url_fragment, but AP didn't.

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -48,6 +48,8 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
   const shouldWatchAnswer = isQuestion(props.embeddable);
   const [loading, setLoading] = useState(shouldWatchAnswer);
 
+  console.log(props.embeddable);
```
kswenson, line 51: Remove?

kswenson (approved):
👍 Looks good

kswenson (approved):
👍 Looks good

src/components/activity-page/managed-interactive/managed-interactive.tsx
pjanik: Thanks, removed.

PR #118, opened 2021-01-21
Improve error messages, track network issues (based on Firestore saving)

pjanik:
Requirments are described here: 
https://www.pivotaltracker.com/story/show/174052604

RequestTracker tracks Firestore update requests (promises) and calls two kinds of handlers:
- timeout - it happens when a request took too long, most likely network issues
- success after timeout - it happens when all the delayed requests finally resolve. The network is back.

Not sure if this naming is clear, so if you have some suggestions, I'm happy to update it. :)


src/firebase-db.ts
```diff
@@ -61,6 +62,19 @@ const configurations: IConfigs = {
   }
 };
 
+const MAX_FIRESTORE_SAVE_TIME = 1000;
+const requestTracker = new RequestTracker(MAX_FIRESTORE_SAVE_TIME);
+// The provided handler will be invoked when some answer doesn't get saved in Firestore within MAX_FIRESTORE_SAVE_TIME.
+// This lets the app to notify users that there are some network issues.
```
kswenson, line 68:
```suggestion
// This lets the app notify users that there are some network issues.
```

src/firebase-db.ts
```diff
@@ -61,6 +62,19 @@ const configurations: IConfigs = {
...
+export const onFirestoreSaveTimeout = (handler: () => void) => {
+  requestTracker.timeoutHandler = handler;
+};
+// The provided handler will be invoked when all the requests that took longer than MAX_FIRESTORE_SAVE_TIME
+// finally succeed. This lets the app to notify users that network issues have been resolved.
```
kswenson, line 73:
```suggestion
// finally succeed. This lets the app notify users that network issues have been resolved.
```

src/firebase-db.ts
```diff
@@ -61,6 +62,19 @@ const configurations: IConfigs = {
   }
 };
 
+const MAX_FIRESTORE_SAVE_TIME = 1000;
```
kswenson: One second seems short to me, but I don't have much confidence in my intuition here. I guess we can see how it behaves in the field. We don't want to spam users with flaky networks with repeated alerts.

kswenson (approved):
👍 Very nice! I'll probably be reusing some of this in CLUE at some point.

src/components/error/error.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const errorMsg: Record<ErrorType, string> = {
+  auth: "We can't establish a connection to your account.",
+  network: "Your network connection has been lost or interrupted.",
+  timeout: "Your session has expired."
+};
+
+export const Error: React.FC<IProps> = ({ type }) => {
+  return (
+    <div className="error" data-cy="error">
+      <h1>Hmm... we&apos;re having trouble connecting.</h1>
```
mklewandowski, line 18: I would probably just say "We're having trouble connecting." but don't feel strongly about it.

mklewandowski (approved):
👍 LGTM!

src/firebase-db.ts
pjanik: Right, it was there just to make some manual tests easier. :)

src/components/error/error.tsx
pjanik: I have no opinion here, but this exact wording comes from Kiley who designed this error message:
https://www.pivotaltracker.com/story/show/174052604/comments/218220039
and it's also used by Micheal in UI mocks. So, I think I'll leave it for now. Easy to change later too.

PR #119, opened 2021-01-21
Add plugin data saving [#174286329]

dougmartin:


scytacki (changes requested):
The code coverage of these changes is a little low. 
Details of what isn't covered can be seen here:
https://codecov.io/gh/concord-consortium/activity-player/pull/119/diff

It looks like many of the lines are error checking. 
It looks like the anonymous path in: `setLearnerPluginState` is untested.
And the new `getFirebaseJwtFromPortal` function is untested.

Note in this repo we don't have coverage tags/labels setup, so it isn't easy to see whether it is cypress or jest that is covering certain lines.

dougmartin:
@scytacki I improved the code coverage of the anonymous and no portal data paths in the firebase-db.ts file in this follow on commit:  https://github.com/concord-consortium/activity-player/pull/119/commits/05f72a94078243ad0eb47c33ab8de8f0da54b059

In the original PR commits I commented out the old getFirebaseJWT tests (as they used fetch which are easily mockable) whereas the new functions call into the existing portal-api code which uses superagent which is very difficult to mock.

Since we use both fetch and superagent in the code I started to go down the path of replacing superagent with fetch but that felt like it was outside scope of this story.  I think we should add a story to switch over to using fetch everywhere in the codebase.

scytacki (approved):
Seems like the best we can do in this PR.

Regarding removing superagent, the other thing we should do is add nock to the project. I've seen examples of using that to test superagent, so we could potentially avoid removing superagent if we just test with nock instead of the jest network mocking that the repo is currently using.  

Removing superagent is also a good thing to do, and seems like it might not be that hard. Kirk has been suggesting for a while that we make this portal communication into a library so it can be shared by CLUE, AP, Portal-Report and any other SPA that is talking to the portal. Right now I think we are just copying code from the most recent SPA when starting a new one.  Anyhow that is all just a tangent. This PR is good to go.

PR #121, opened 2021-01-22
repurpose the orange bar to warn when not using production database

scytacki:


scytacki:
@mklewandowski I added you as a reviewer since you've worked on this orange bar before.

mklewandowski (approved):
👍 LGTM!

PR #122, opened 2021-01-25
Enforce session timeout

pjanik:
[#173359395]

I've added a new utility called IdleDetector. It's based on the existing library, although it was a bit dead (updated 2-3 years ago) and written in JS. I quickly rewrote it to TS as I was hoping I could support iframes. Finally, it turned out I can't really do it. I've left this copy here, as it's in TS now and I've added some tests that the original version didn't have.

If the user interacts only with iframes, we might trigger the warning about being idle. I can't really see a solution for that. But we don't log out user automatically. He'll still have 5 minutes to get back to the activity. Given that the initial timeout is 20 minutes, I think it's very unlikely to happen. And even if it does, there's still an easy way to continue work.

Screenshot 2021-01-25 at 13 41 10

@scytacki, we discussed that a bit, you might be interested in the final error UI.

If you have some suggestions re wording, I'm happy to update it. It hasn't been designed by Kiley or Michael, I just based it on Michael's generic error UI.

### DIFF
--- a/src/components/error/idle-warning.tsx
+++ b/src/components/error/idle-warning.tsx
@@ -3,18 +3,23 @@ import "./error.scss";
 
 interface IProps {
   username: string;
+  anonymous: boolean;
+  timeout: number;
   onTimeout: () => void;
   onContinue: () => void;
   onExit: () => void;
-  timeout: number;
 }
 
 const kMaxTimeInterval = 500; // ms
 
-export const IdleWarning: React.FC<IProps> = ({ username, timeout, onTimeout, onContinue, onExit }) => {
+export const IdleWarning: React.FC<IProps> = ({ username, anonymous, timeout, onTimeout, onContinue, onExit }) => {
   const [ time, setTime ] = useState(0);
 
   useEffect(() => {
+    if (anonymous) {
+      // Don't start the counter in anonymous mode.
+      return;
+    }
     const startTime = Date.now();
     const intervalId = setInterval(() => {
       const currentTime = Date.now();
@@ -25,7 +30,7 @@ export const IdleWarning: React.FC<IProps> = ({ username, timeout, onTimeout, on
       }
     }, Math.min(timeout / 2, kMaxTimeInterval));
     return () => window.clearInterval(intervalId);
-  }, [timeout]);
+  }, [timeout, anonymous]);
 
   useEffect(() => {
     if (time >= timeout) {
@@ -39,13 +44,24 @@ export const IdleWarning: React.FC<IProps> = ({ username, timeout, onTimeout, on
 
   return (
     <div className="error" data-cy="idle-warning">
-      <h1>{username }, your session is about to expire.</h1>
-      <p>You will be logged out automatically in 
-        <b>{min > 1 && ` ${min} minutes and`}{min === 1 && ` ${min} minute and`}{` ${sec} seconds` }</b>.
-      </p>
-      <p>Please choose to continue your session or to log in as a different user.</p>
-      <button className="button" onClick={onExit} data-cy="exit">Log in as a different user</button> 
-      <button className="button" onClick={onContinue} data-cy="continue"> Continue your session </button> 
+      { 
+        anonymous ? 
+        <div>
+          <h1>You&apos;ve been idle for too long.</h1>
+          <p>Click continue below to keep working.</p>
+          <button className="button" onClick={onContinue} data-cy="continue"> Continue your session </button> 
+        </div>
+        :
+        <div>
+          <h1>{username }, your session is about to expire.</h1>
+          <p>You will be logged out automatically in 
+            <b>{min > 1 && ` ${min} minutes and`}{min === 1 && ` ${min} minute and`}{` ${sec} seconds` }</b>.
+          </p>
+          <p>Please choose to continue your session or to log in as a different user.</p>
+          <button className="button" onClick={onExit} data-cy="exit">Log in as a different user</button> 
+          <button className="button" onClick={onContinue} data-cy="continue"> Continue your session </button> 
+        </div>
+      }
     </div>
   );
 };
--- a/src/components/error/idle-warning.tsx
+++ b/src/components/error/idle-warning.tsx
@@ -9,6 +9,8 @@ interface IProps {
   timeout: number;
 }
 
+const kMaxTimeInterval = 500; // ms
+
 export const IdleWarning: React.FC<IProps> = ({ username, timeout, onTimeout, onContinue, onExit }) => {
   const [ time, setTime ] = useState(0);
 
@@ -21,7 +23,7 @@ export const IdleWarning: React.FC<IProps> = ({ username, timeout, onTimeout, on
       if (timeDiff >= timeout) {
         window.clearInterval(intervalId);
       }
-    }, Math.min(timeout / 2, 500));
+    }, Math.min(timeout / 2, kMaxTimeInterval));
     return () => window.clearInterval(intervalId);
   }, [timeout]);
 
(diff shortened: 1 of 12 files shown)


src/components/error/idle-warning.tsx
```diff
@@ -0,0 +1,50 @@
...
+  useEffect(() => {
+    const startTime = Date.now();
+    const intervalId = setInterval(() => {
+      const currentTime = Date.now();
+      const timeDiff = currentTime - startTime;
+      setTime(timeDiff);
+      if (timeDiff >= timeout) {
+        window.clearInterval(intervalId);
+      }
+    }, Math.min(timeout / 2, 500));
```
mklewandowski, line 24: Should this 500 be stored in a constant?

src/components/error/idle-warning.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }, [onTimeout, time, timeout]);
+  
+  const timeLeft = timeout - time;
+  const sec = Math.round((timeLeft / 1000) % 60);
+  const min = Math.floor((timeLeft / 1000) / 60);
+
+  return (
+    <div className="error" data-cy="idle-warning">
+      <h1>{username }, your session is about to expire.</h1>
+      <p>Session has been idle over its time limit.</p>
```
mklewandowski, line 41: Is this the wording from LARA? There's probably a more user-friendly way to say this instead of "Session has been idle over its time limit." (especially since our audience is kids).

mklewandowski (approved):
👍 This looks good!  I think some of the wording could be a bit clearer for our users, but I'm guessing much of this came from LARA so we want to leave as is.

kswenson (approved):
👍 LGTM -- we may want to wrap this into an npm library under `@concord-consortium` at some point. I can see this being reused in other projects (like CLUE).

src/utilities/idle-detector.ts
```diff
@@ -0,0 +1,138 @@
...
+  events.forEach(function (event) {
+    object.removeEventListener(event, callback, true);
+  });
+};
+
+const visibilityEvents = ["visibilitychange", "webkitvisibilitychange", "mozvisibilitychange", "msvisibilitychange"];
+
+const defaultOptions: IOptions = {
+  idle: 10000, // idle time in ms
+  events: ["mousemove", "keydown", "mousedown", "touchstart", "scroll"], // events that will trigger the idle resetter
```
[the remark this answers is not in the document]
pjanik: This comes from the original lib. Hm, I guess `touchstart` will be fired at the beginning of the `touchmove`?

src/components/error/idle-warning.tsx
pjanik: Thanks, fixed.

pjanik:
> 👍 This looks good! I think some of the wording could be a bit clearer for our users, but I'm guessing much of this came from LARA so we want to leave as is.

Actually, this one doesn't come from LARA. I think I just googled for similar dialogs. If you have some suggestions (as native speakers), I'm happy to edit. :)

src/components/error/idle-warning.tsx
scytacki: @pjanik I'd suggest just removing "Session has been idle over its time limit.".  That seems to be just restating the heading text with different words. 

I'd guess a more concise message means students are more likely to read all of the message. :)

scytacki (changes requested):
@pjanik it would be good if this event of "continuing the session" is logged. I didn't check your changes carefully, but I didn't see it being logged to both GA and our log system. I think the AP has a way to log events to both systems with one function, but I might be wrong.

I'd also note that the code coverage is a little lower than the target. You can see the details here: https://codecov.io/gh/concord-consortium/activity-player/pull/122/diff
Perhaps it would be possible to use cypress's clock function to make a cypress test which brings up the dialog and exercises it: https://docs.cypress.io/api/commands/clock.html#Yields

Another question is how this handles anonymous users. It seems that case should be handled differently. For example they often will not be launched from the portal, so telling the user to go back to the portal would be wrong. In the anonymous case the user can just reload the page and the run_key in the URL will continue to work and they could see their data. 
I'd guess the easiest approach is to disable the idle timer when anonymous. 
Or we could show a different dialog that basically says, "You've been idle for too long. Click continue below to keep working".  So that way there'd be kind of a similar experience between the auth'd and anonymous and we can know in the log messages whether the student is actually paying attention. 
I don't think this case is a priority, so doing whichever is easier seems best. Perhaps the anonymous dialog will be harder, but make it easier to get more complete code coverage. In that case it might be worth it.

src/components/error/idle-warning.tsx
pjanik: Thanks, updated.

pjanik:
> @pjanik it would be good if this event of "continuing the session" is logged. I didn't check your changes carefully, but I didn't see it being logged to both GA and our log system. I think the AP has a way to log events to both systems with one function, but I might be wrong.

Done.

> I'd also note that the code coverage is a little lower than the target. You can see the details here: https://codecov.io/gh/concord-consortium/activity-player/pull/122/diff
> Perhaps it would be possible to use cypress's clock function to make a cypress test which brings up the dialog and exercises it: https://docs.cypress.io/api/commands/clock.html#Yields
> Another question is how this handles anonymous users. It seems that case should be handled differently. For example they often will not be launched from the portal, so telling the user to go back to the portal would be wrong. In the anonymous case the user can just reload the page and the run_key in the URL will continue to work and they could see their data.
> I'd guess the easiest approach is to disable the idle timer when anonymous.
> Or we could show a different dialog that basically says, "You've been idle for too long. Click continue below to keep working". So that way there'd be kind of a similar experience between the auth'd and anonymous and we can know in the log messages whether the student is actually paying attention.
> I don't think this case is a priority, so doing whichever is easier seems best. Perhaps the anonymous dialog will be harder, but make it easier to get more complete code coverage. In that case it might be worth it.

I think mostly app.tsx changes are untested, new components and utils should be. Generally, I'd say there's some growing tech debt in AP as its core parts, which require some more effort and planning in terms of testing, are not really covered yet - app.tsx logic, managed interactive, iframe runtime. I guess it'd be nice to work on that separately from feature stories at this point. I'm mostly talking about unit testing. Cypress will cover some of that, although it doesn't seem we have a pattern to test the launching from (fake) Portal. So, it might be difficult to test some features when we have separate code paths for logged in and anonymous student.

Re anonymous case and different dialog - I thought it'd be still useful to have the same error with the large username ("Anonymous") and the button to log in as a different user. If many students share the same machine, it might help to avoid completing the activity as anonymous by accident. Although, when two or more students share one machine, I guess they will be all either logged in or anonymous. Note that error messages (network, auth) designed by Kiley and Michael have "Click here" link that also redirects to Portal and defaults to learn.concord.org for anonymous users (where portal info it's not available). If we're changing it here, should we change these generic errors too?
This new idle warning shouldn't have any counter, just a button to continue?

pjanik:
OK, to move forward:

1. I've added anonymous variant of the idle warning following @scytacki suggestions:
Screenshot 2021-01-27 at 17 57 37

2. I've added Cypress tests. Finally, I've managed to use `cy.clock` and `cy.tick`. The key was the correct order of Cypress commands. When AP downloads sample activity, it actually uses a timer too (to fake network requests). The code execution was getting stuck there because of the cy.clock. So, that's why I had to call cy.tick(1000) between cy.visit and cy.get(activity-summary).

3. To test logged in user variant, I've added `__cypressLoggedIn` URL param. Not pretty, but I think works fine for this use case. When we setup Cypress to handle URLs with some fake `token=`, this param should be removed.

scytacki:
@pjanik looks good, but there is a failing Jest test.

scytacki (approved):
Looks good.

PR #124, opened 2021-02-03
Prevent long activity names from overlapping username in header.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176668780

[#176668780]

This change ensures that content within the `.header-center` container doesn't overflow and overlap the `.header-right` container.


mklewandowski (approved):
👍 LGTM!

PR #125, opened 2021-02-04
Added start of service worker for offline use [#176557502]

dougmartin:


scytacki (approved):
Looks good to me. 
I made a separate story about adding more tests: https://www.pivotaltracker.com/story/show/176851489

PR #128, opened 2021-02-17
Added manifest.json for PWA installation [#176575777]

dougmartin:


dstrawberrygirl (approved):
I think this is fine, I found conflicting information on manifest.json formats, but seems that's related more to crafting browser extensions, not PWAs. So, assuming that the v2 or v3 confusion is irrelevant in our situation we're probably good with this as a starter!

PR #129, opened 2021-02-17
Fix check for plugin content, ensure we check last page

mklewandowski:
Fix bug that caused us to skip the last activity page when checking for plugin content.  If the plugin content was on the last page ONLY, the plugin would fail to be loaded.

### DIFF
--- a/src/utilities/plugin-utils.ts
+++ b/src/utilities/plugin-utils.ts
@@ -26,7 +26,7 @@ const addUsedPlugin = (plugin: Plugin) => {
 
 export const findUsedPlugins = (activity: Activity, teacherEditionMode: boolean) => {
   // search each page for teacher edition plugin use
-  for (let page = 0; page <= activity.pages.length - 1; page++) {
+  for (let page = 0; page < activity.pages.length; page++) {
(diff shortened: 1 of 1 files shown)


src/utilities/plugin-utils.ts
```diff
@@ -26,7 +26,7 @@ const addUsedPlugin = (plugin: Plugin) => {
 
 export const findUsedPlugins = (activity: Activity, teacherEditionMode: boolean) => {
   // search each page for teacher edition plugin use
-  for (let page = 0; page < activity.pages.length - 1; page++) {
+  for (let page = 0; page <= activity.pages.length - 1; page++) {
```
kswenson, line 29:
```suggestion
  for (let page = 0; page < activity.pages.length; page++) {
```
Seems like the fix could be (slightly) simpler.

kswenson (approved):
👍 Fix is fine. I made a suggestion to simplify the code slightly, but the effect is the same.

src/utilities/plugin-utils.ts
mklewandowski: Fixed

PR #130, opened 2021-02-18
Various fixes for offline authoring and usage

dougmartin:
- Change cache loader to use no-cors mode since we don't need the actual response
- Enable offline mode when in authoring so the service worker starts
- Change to only load launch list saved in localstorage when in offline mode
- Skip the current offline mode rendering when in authoring mode


knowuh (approved):
Looks good.

PR #131, opened 2021-02-18
Begin work on saving / loading student data to indexedDB

knowuh:
@dougmartin — when you have a second could you take a look.

### DIFF
--- a/src/dexie-storage.ts
+++ b/src/dexie-storage.ts
@@ -1,0 +1,19 @@
+import Dexie from "dexie";
+import { IExportableAnswerMetadata } from "./types";
+
+// Copy and pasted from the example: https://dexie.org/docs/Typescript
+export class DexieStorage extends Dexie {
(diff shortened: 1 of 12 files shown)


src/dexie-storage.ts
```diff
@@ -0,0 +1,19 @@
...
+export class DexieStorage extends Dexie {
+  // Declare implicit table properties.
+  // (just to inform Typescript. Instantiated by Dexie in stores() method)
+  answers: Dexie.Table<IExportableAnswerMetadata, string>; // number = type of the primkey
+  //...other tables goes here...
+
+  constructor () {
+      super("ActivityPlayer");
+      this.version(2).stores({
+          answers: "id, question_id, question_type",
```
dougmartin, line 14: I think we will need to add some reference later to the activity id so we can query answer status per activity.

dougmartin (approved):
Looks good :+1:

PR #132, opened 2021-02-22
Improved code coverage of plugin-utils

dougmartin:
NOTE: this is based on the `master` branch and not the `offline-mode` branch as we haven't touched this file in `offline-mode`.


dougmartin:
@knowuh / @dstrawberrygirl - could you take a look at this older PR?  I'd like to merge it to master and and merge master back into the offline-mode branch in a follow on PR today so we keep them in sync.

src/utilities/plugin-utils.test.ts
```diff
@@ -0,0 +1,200 @@
...
+      expect(getUsedPlugins().length).toEqual(1);
+      addUsedPlugin(plugin);
+      expect(getUsedPlugins().length).toEqual(1);
+    });
+  });
+
+  describe("#findUsedPlugins", () => {
+    it("works in teacher edition mode", () => {
+      const usedPlugins = findUsedPlugins(activity, true);
+      expect(usedPlugins.length).toBe(3);
```
scytacki, line 59: Since the usedPlugins length is the same with and without teacher edition mode, does this mean that the test activity doesn't have any teacher edition plugins?

Could you update this to use an activity with teacher edition plugins so it would exercise more of the code?

src/utilities/plugin-utils.test.ts
```diff
@@ -0,0 +1,200 @@
...
+      const result = validateEmbeddablePluginContextForWrappedEmbeddable(partialContext);
+      expect(result?.LARA).toEqual(partialContext.LARA);
+      expect(result?.embeddable).toEqual(partialContext.embeddable);
+      expect(result?.embeddableContainer).toEqual(partialContext.embeddableContainer);
+    });
+  });
+
+  describe("#loadLearnerPluginState", () => {
+    it("returns the state for all plugins in teacher mode", async () => {
+      const pluginState = await loadLearnerPluginState(activity, true);
```
scytacki, line 123: How does this work since there is no firebase connection? It seems like this should make a firebase connection to look for the state, is that being mocked somehow?

scytacki (changes requested):
This looks good. My only request is to use an activity with teacher edition plugins in it. At least it seems like the current one used, doesn't have them since the counts of plugins doesn't change in the tests with and without teacher edition. But maybe I'm missunderstanding.

src/utilities/plugin-utils.test.ts
dougmartin: This has been fixed in this commit (https://github.com/concord-consortium/activity-player/pull/132/commits/4b9eaad5d731de51119f9ce09c85894f2d9bc565).  I needed to clear the used plugins cache between tests so that accurate counts could be tested.  I also changed from just counting the plugins to comparing the plugin script labels.

src/utilities/plugin-utils.test.ts
dougmartin: This has been fixed in this commit (https://github.com/concord-consortium/activity-player/pull/132/commits/353170d182a5f9e1a4424f704d414f1910b02cb2).  There were multiple issues:

1. The lack of firebase connection caused the method to exit before setting the cache.
2. The lack of a return on the await in `loadLearnerPluginState` caused `null` to be returned.

dougmartin:
@scytacki the tests have been updated to address your comments.

dougmartin:
@scytacki I forgot to re-request a review on this last week when I added the commit.

scytacki (approved):
Looks good.

PR #134, opened 2021-02-23
Added offline activity list [#176963850]

dougmartin:
- Startup code now saves launch list activities to a Dexie table
- In offline mode when no activity is selected a list of activities is shown
- In offline mode when an activity is selected a header is added to return to the list of activities


knowuh (approved):
LGTM

PR #136, opened 2021-02-23
Offline manifest for Precipitating Change

dstrawberrygirl:
Sample of the manifest for the Precipitating Change activity setup on Staging


dougmartin (approved):
Looks good :+1:

PR #137, opened 2021-02-23
Read/Write IndexedDB Answers

dstrawberrygirl:
- Added the current ActivityID / URL to the Answers table in IndexedDB to link answers to the current activity
- Moved some utility functions out of `firebase-db.ts` as common functions to both the online and offline storage handlers
- Added an export utility to the `storage-facade.ts` to gather all answers and make a "Save My Work" button on the completion page that the student could click to download their work so far to a local json file.

### DIFF
--- a/src/components/activity-completion/completion-export-answers.tsx
+++ b/src/components/activity-completion/completion-export-answers.tsx
@@ -1,0 +1,35 @@
+import React from "react";
+import { Storage } from "../../storage-facade";
+
+import "./completion-export-answers.scss";
+
+interface IProps { }
+interface IState {
+  activityJSON: string
+}
+
+export class CompletionExportAnswers extends React.PureComponent<IProps, IState> {
+
+  public constructor(props: IProps) {
+    super(props);
+    this.state = {activityJSON: ""};
+  }
+  async componentDidMount() {
+    const activityJSON = await Storage.exportActivityToJSON();
+    this.setState({ activityJSON });
+  }
+
+  render() {
+    const { activityJSON } = this.state;
+    const bb = new Blob([activityJSON], { type: "text/plain" });
+    const activityLink = window.URL.createObjectURL(bb);
+
+    return (
+      <div className={"activity-export-download"}>
+        <a href={activityLink} download={"allmyanswers.json"}>
+          <div className={"button"}>Save My Work</div>
+        </a>
+      </div>
+    );
+  }
+}
--- a/src/components/activity-completion/completion-export-answers.tsx
+++ b/src/components/activity-completion/completion-export-answers.tsx
@@ -5,28 +5,30 @@ import "./completion-export-answers.scss";
 
 interface IProps { }
 interface IState {
-  activityJSON: string
+  activityJSON: string,
+  filename: string
 }
 
 export class CompletionExportAnswers extends React.PureComponent<IProps, IState> {
 
   public constructor(props: IProps) {
     super(props);
-    this.state = {activityJSON: ""};
+    this.state = {activityJSON: "", filename: "activity"};
   }
   async componentDidMount() {
-    const activityJSON = await Storage.exportActivityToJSON();
-    this.setState({ activityJSON });
+    const activityJSONComplete = await Storage.exportActivityToJSON();
+    const filename = activityJSONComplete.filename;
+    this.setState({ activityJSON: JSON.stringify(activityJSONComplete), filename });
   }
 
   render() {
-    const { activityJSON } = this.state;
+    const { activityJSON, filename } = this.state;
     const bb = new Blob([activityJSON], { type: "text/plain" });
     const activityLink = window.URL.createObjectURL(bb);
 
     return (
       <div className={"activity-export-download"}>
-        <a href={activityLink} download={"allmyanswers.json"}>
+        <a href={activityLink} download={`${filename}.json`}>
           <div className={"button"}>Save My Work</div>
         </a>
       </div>
--- a/src/dexie-storage.ts
+++ b/src/dexie-storage.ts
@@ -1,18 +1,19 @@
 import Dexie from "dexie";
-import { OfflineActivity, IExportableAnswerMetadata } from "./types";
+import { OfflineActivity } from "./types";
+import { IIndexedDBAnswer } from "./storage-facade";
 
 // Copy and pasted from the example: https://dexie.org/docs/Typescript
 export class DexieStorage extends Dexie {
   // Declare implicit table properties.
   // (just to inform Typescript. Instantiated by Dexie in stores() method)
-  answers: Dexie.Table<IExportableAnswerMetadata, string>; // number = type of the primkey
   offlineActivities: Dexie.Table<OfflineActivity, string>;
+  answers: Dexie.Table<IIndexedDBAnswer, string>; // number = type of the primkey
   //...other tables goes here...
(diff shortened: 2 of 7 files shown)


src/components/activity-completion/completion-export-answers.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Storage } from "../../storage-facade";
+
+import "./completion-export-answers.scss";
+
+interface IProps { }
+interface IState {
+  activityJSON: string
+}
+
+export class CompletionExportAnswers extends React.PureComponent<IProps, IState> {
```
dougmartin, line 11: `React.PureComponent` should work here since the state is "shallow" (only has top level keys).  But it should probably be changed to `React.Component` in case the state is changed to have "deep" keys as it will then subtly break as noted in the React docs:

> React.PureComponent’s shouldComponentUpdate() only shallowly compares the objects. If these contain complex data structures, it may produce false-negatives for deeper differences. Only extend PureComponent when you expect to have simple props and state, or use forceUpdate() when you know deep data structures have changed. Or, consider using immutable objects to facilitate fast comparisons of nested data.

> Furthermore, React.PureComponent’s shouldComponentUpdate() skips prop updates for the whole component subtree. Make sure all the children components are also “pure”.

src/components/activity-completion/completion-export-answers.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  render() {
+    const { activityJSON } = this.state;
+    const bb = new Blob([activityJSON], { type: "text/plain" });
+    const activityLink = window.URL.createObjectURL(bb);
+
+    return (
+      <div className={"activity-export-download"}>
+        <a href={activityLink} download={"allmyanswers.json"}>
```
dougmartin, line 29: What are you thoughts of adding a date to the `allmyanswers.json` file like `allmyanswers-MM-DD-YYYY.json`?

src/components/activity-completion/completion-export-answers.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export class CompletionExportAnswers extends React.PureComponent<IProps, IState> {
+
+  public constructor(props: IProps) {
+    super(props);
+    this.state = {activityJSON: ""};
+  }
+  async componentDidMount() {
+    const activityJSON = await Storage.exportActivityToJSON();
```
dougmartin, line 18: I'm not sure it matters other than improving rendering speed but instead of doing this at mount time you could also do it when the link is clicked like it is done to download the authoring json here:

https://github.com/concord-consortium/activity-player/blob/offline-mode/src/components/offline-manifest-authoring-nav.tsx#L19-L26

with the link markup looking like:

```
<a href="" download={`${offlineManifestAuthoringId}.json`} onClick={handleDownloadClicked}>Download JSON</a>
```

src/dexie-storage.ts
```diff
@@ -1,18 +1,19 @@
...
-  answers: Dexie.Table<IExportableAnswerMetadata, string>; // number = type of the primkey
   offlineActivities: Dexie.Table<OfflineActivity, string>;
+  answers: Dexie.Table<IIndexedDBAnswer, string>; // number = type of the primkey
   //...other tables goes here...
 
   constructor () {
       super("ActivityPlayer");
-      this.version(2).stores({
-          answers: "id, question_id, question_type",
+      this.version(4).stores({
```
dougmartin, line 15: Seems weird that version 4 comes before version 3.  My reading of the docs was that you added versions and left old versions alone.  New versions can add columns to existing tables as shown in the docs (https://dexie.org/docs/Tutorial/Design#database-versioning):

```
In Dexie >= 3.0:

Just increase the version number and add shoeSize along the indexes:
db.version(2).stores({friends: "++id,name,shoeSize"});
```

src/storage-facade.ts
```diff
@@ -3,10 +3,37 @@ import * as FirebaseImp from "./firebase-db";
...
+import { refIdToAnswersQuestionId } from "./utilities/embeddable-utils";
 
 export interface IStorageInitializer { name: FirebaseImp.FirebaseAppName, preview: boolean }
 
+export interface IWrappedDBAnswer {
+  meta: IExportableAnswerMetadata;
+  interactiveState: any;
+}
+
+export type IIndexedDBAnswer = IExportableAnswerMetadata & { activity: string };
```
dougmartin, line 15: Nice.  One of these days I'm going to sit down and figure out the difference between extending an `interface` and "intersection types" like this one (https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#intersection-types).

dougmartin (approved):
Looks good.  Just a couple of suggestions in the review.

src/components/activity-completion/completion-export-answers.tsx
dstrawberrygirl: Good point, thinking something like activity-name-MM-DD-YYYY.json may be even better.

src/components/activity-completion/completion-export-answers.tsx
dstrawberrygirl: For consistency we should probably standardize on one approach or the other, and since you wrote yours first, I'll update this one to match :)

src/dexie-storage.ts
dstrawberrygirl: Version 3 was local, it just wasn't committed before being surpassed by 4

src/components/activity-completion/completion-export-answers.tsx
dstrawberrygirl: if we change the link to build the json on click, similar to the offline manifest builder, then I don't think this will be an issue? I'm not sure this button will do much else

src/storage-facade.ts
```diff
@@ -45,25 +75,71 @@ const FireStoreStorageProvider: StorageInterface = {
...
+    getAnswerFromIndexDB(questionId).then( (answer: IIndexedDBAnswer|null) => {
+      if (answer) {
+        callback(docToWrappedAnswer(answer));
+      } else {
+        callback(null);
+      }
+    });
+  },
+
+  exportActivityToJSON: (activityId?: string) => {
```
dougmartin, line 124: @knowuh / @dstrawberrygirl  - one thing I just thought of - we should add a version string at the top level of this json so that we can import old data if the structure changes in the future.

src/components/activity-completion/completion-export-answers.tsx
dstrawberrygirl: Update: having discussed this one we've decided to leave the link as-is because of the async call to IndexedDB

src/storage-facade.ts
knowuh: Yup, that is a great idea.

src/storage-facade.ts
```diff
@@ -58,7 +79,7 @@ export interface StorageInterface {
   checkIfOnline: () => Promise<boolean>,
 
   // for saving a whole activity to JSON
-  exportActivityToJSON: (activityId?: string) => Promise<string> | string
+  exportActivityToJSON: (activityId?: string) => Promise<any> | any
```
knowuh, line 82: I think we can tighten up this `any` here.

knowuh (commented):
Looks good to me,  we should probably just address the one `any` typing.

PR #139, opened 2021-02-25
Add new import component

dstrawberrygirl:
Building on the option to export data as a JSON file, add UI element to import previously-exported answers from a JSON file back into the Activity Player and update the IndexedDB stored answers to the newly-imported answers.


knowuh (approved):
Looks good!

PR #140, opened 2021-02-25
Remove border around image and video interactives

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176855442

[#176855442]

With these changes, only interactives that save state (and therefore get a question number) will have a border. So in addition to images and videos that don't save state, Model My Watershed models and Lab interactives will also be borderless.

### DIFF
--- a/src/components/activity-page/managed-interactive/managed-interactive.tsx
+++ b/src/components/activity-page/managed-interactive/managed-interactive.tsx
@@ -192,6 +192,7 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
   // with this fragment if necessary. ActivityPlayer doesn't have knowledge about URL format and provided url_fragment 
   // to perform this merge automatically.
   const iframeUrl = activeDialog?.url || (embeddable.url_fragment ? url + embeddable.url_fragment : url);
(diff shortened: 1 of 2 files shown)


src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -192,6 +192,7 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
   // with this fragment if necessary. ActivityPlayer doesn't have knowledge about URL format and provided url_fragment 
   // to perform this merge automatically.
   const iframeUrl = activeDialog?.url || (embeddable.url_fragment ? url + embeddable.url_fragment : url);
+  const miContainerClass = questionNumber ? "managed-interactive has-question-number" : "managed-interactive";
```
mklewandowski, line 195: I'm assuming questionNumber is never 0, right?

mklewandowski (approved):
👍 LGTM!

src/components/activity-page/managed-interactive/managed-interactive.tsx
emcelroy: Yeah, it's either `undefined` or some number greater than zero.

PR #141, opened 2021-02-25
Links in text boxes will not open in new tabs.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177104163

[#177104163]

This change will prevent the `target` attribute from being removed from HTML. This is needed so that links in activities can open in new windows/tabs.

### DIFF
--- a/src/utilities/render-html.ts
+++ b/src/utilities/render-html.ts
@@ -15,5 +15,5 @@ interface DomElement {
 export type ParseHTMLReplacer = (domNode: DomElement) => JSX.Element | void | undefined | null | false;
 
 export function renderHTML(html: string, replace?: ParseHTMLReplacer) {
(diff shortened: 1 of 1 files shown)


emcelroy:
@scytacki Kirk came up with this fix, which nicely fixes the problem described in the [PT story](https://www.pivotaltracker.com/story/show/177104163). He wondered a little about any security issues with allowing the use of `target`, but it seems pretty safe considering `rel="noopener"` is automatically added to any links. Are you aware of any potential issues that aren't solved by `rel="noopener`?

kswenson (approved):
👍 LGTM -- Given that I don't think LARA is doing much sanitization at all, I suspect allowing the `target` attribute in the ActivityPlayer won't introduce any new issues but I'm curious whether @scytacki can think of any.

PR #142, opened 2021-02-26
Pass the sourceKey and firebase-app to the portal-report from show work button

scytacki:
This was done for logged in users, but it wasn't being done for anonymous users.
This caused a problem when working on production because the firebase-app defaults to the dev database.
The sourcekey passing is helpful just be more explicit so the portal-report doesn't have to guess this value.


scytacki:
@eireland can you review this?

eireland (approved):
Looks good

PR #143, opened 2021-03-02
Updated smoke-test manifest

dougmartin:
- Assets updated
- Changed activity to point to copy of main smoke test activity so we don't have to update again


knowuh (approved):
LGTM

PR #144, opened 2021-03-02
Fixed flash of "no offline activities found" at startup when loading a manifest

dougmartin:
The flag that denotes the manifest is loading has been hoisted into the constructor from the componentDidMount method so that the initial render does not flash an error message about no offline activites found before the manifest starts loading.


knowuh (approved):
LGTM

PR #145, opened 2021-03-03
177118245 update answer fields

scytacki:
This fixes the handling of generic interactives that don't have an authored state and or don't specify their type of answer or question. Interactives like that make up the majority of our interactives.

It also handles basic string, null, boolean, and numeric interactiveState

It adds the remote_endpoint to the answer documents. LARA runtime adds this, but the AP was not. This isn't needed to restore data in the AP, or to show it in the portal-report. But it will be needed for efficient searching in the new Athena Reports. 

It improves the tests so we can see the shape of the answer documents better and be notified when new properties are added or properties are removed.

It fixes the resource_url (called resourceUrl in the code) saved in the answer documents. Previously it was a special AP url form which wouldn't match the url in the resource (aka report-structure) stored in the report-service. So if a researcher wanted to search for all answers from a particular activity across offerings they'd have problems with this mis-match AP url form.

It also adds a lot of comments around the conversion of the interactiveState to the answer document.

TODO:
- [x] add test for numeric state
- [x] test coverage
- [x] typing in embeddable-utils.ts

### DIFF
--- a/src/components/activity-page/managed-interactive/managed-interactive.tsx
+++ b/src/components/activity-page/managed-interactive/managed-interactive.tsx
@@ -59,7 +59,7 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
     }
   }, [embeddableRefId, shouldWatchAnswer]);
 
-    const handleNewInteractiveState = (state: IRuntimeMetadata) => {
+    const handleNewInteractiveState = (state: any) => {
       // Keep interactive state in sync if iFrame is opened in modal popup
       interactiveState.current = state;
 
@@ -187,9 +187,9 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
   }));
 
   // embeddable.url_fragment is an optional string (path, query params, hash) that can be defined by author.
-  // Some interactives are authored that way. Note that url_fragment is not merged with the dialog URL. 
-  // Each interactive will be first loaded inline with the url_fragment appended. So it can merge its custom dialog URL 
-  // with this fragment if necessary. ActivityPlayer doesn't have knowledge about URL format and provided url_fragment 
+  // Some interactives are authored that way. Note that url_fragment is not merged with the dialog URL.
+  // Each interactive will be first loaded inline with the url_fragment appended. So it can merge its custom dialog URL
+  // with this fragment if necessary. ActivityPlayer doesn't have knowledge about URL format and provided url_fragment
   // to perform this merge automatically.
   const iframeUrl = activeDialog?.url || (embeddable.url_fragment ? url + embeddable.url_fragment : url);
   const miContainerClass = questionNumber ? "managed-interactive has-question-number" : "managed-interactive";
--- a/src/components/activity-page/managed-interactive/managed-interactive.tsx
+++ b/src/components/activity-page/managed-interactive/managed-interactive.tsx
@@ -4,7 +4,7 @@ import { IframeRuntime, IframeRuntimeImperativeAPI } from "./iframe-runtime";
 import useResizeObserver from "@react-hook/resize-observer";
 import {
   ICloseModal, INavigationOptions,
-  ICustomMessage, IRuntimeMetadata, IShowDialog, IShowLightbox, IShowModal, ISupportedFeatures
+  ICustomMessage, IShowDialog, IShowLightbox, IShowModal, ISupportedFeatures
 } from "@concord-consortium/lara-interactive-api";
 import { PortalDataContext } from "../../portal-data-context";
 import { IManagedInteractive, IMwInteractive, LibraryInteractiveData, IExportableAnswerMetadata } from "../../../types";
--- a/src/firebase-db.test.ts
+++ b/src/firebase-db.test.ts
@@ -88,7 +88,7 @@ describe("Firestore", () => {
       resourceUrl: "http://example/resource",
       toolId: "activity-player.concord.org",
       userType: "learner",
-      runRemoteEndpoint: ""
+      runRemoteEndpoint: "https://portal.com/learner/1234"
     });
 
     const embeddable = {
@@ -108,6 +108,7 @@ describe("Firestore", () => {
 
     expect(appMock.firestore().doc).toHaveBeenCalledWith(`sources/localhost/answers/${exportableAnswer.id}`);
     expect(appMock.firestore().doc().set).toHaveBeenCalledWith({
+      version:1,
       answer: "test",
       answer_text: "test",
       context_id: "context-id",
@@ -116,7 +117,7 @@ describe("Firestore", () => {
       platform_user_id: "1",
       question_id: "managed_interactive_123",
       question_type: "open_response",
-      remote_endpoint: "",
+      remote_endpoint: "https://portal.com/learner/1234",
       report_state: "{\"mode\":\"report\",\"authoredState\":\"{\\\"version\\\":1,\\\"questionType\\\":\\\"open_response\\\",\\\"prompt\\\":\\\"<p>Write something:</p>\\\"}\",\"interactiveState\":\"{\\\"answerType\\\":\\\"open_response_answer\\\",\\\"answerText\\\":\\\"test\\\"}\",\"version\":1}",
       resource_link_id: "2",
       resource_url: "http://example/resource",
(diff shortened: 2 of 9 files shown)


scytacki:
@dougmartin, @knowuh, and @dstrawberrygirl you might want to watch this PR. I suspect it will conflict with some of the offline work. I've got a few more things to wrap up on it, but I think it is pretty much done.

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -59,7 +59,7 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
     }
   }, [embeddableRefId, shouldWatchAnswer]);
 
-    const handleNewInteractiveState = (state: IRuntimeMetadata) => {
+    const handleNewInteractiveState = (state: any) => {
```
dougmartin, line 62: I don't think we should be dropping down to `any` here is we can help it.

src/firebase-db.test.ts
```diff
@@ -88,7 +88,7 @@ describe("Firestore", () => {
       resourceUrl: "http://example/resource",
       toolId: "activity-player.concord.org",
       userType: "learner",
-      runRemoteEndpoint: ""
+      runRemoteEndpoint: "https://portal.com/learner/1234"
```
dougmartin, line 91: Should we use `example.com` here in case the tests ever accidentally use this endpoint?

src/firebase-db.ts
```diff
@@ -228,32 +228,38 @@ export function createOrUpdateAnswer(answer: IExportableAnswerMetadata) {
 
   let answerDocData: LTIRuntimeAnswerMetadata | AnonymousRuntimeAnswerMetadata;
 
+  const commonFields = {
```
dougmartin, line 231: I think you should type this to get some extra compile type protection.

dougmartin:
@scytacki I looked at the existing code and added a couple of review comments.

src/components/activity-page/managed-interactive/managed-interactive.tsx
scytacki: This any is one of the key parts of this PR. Interactives can send us pretty much anything. objects, strings, numbers, booleans.

dougmartin: I've handled this in other repos by aliasing the `any` type (eg `type IRuntimeState = any`).  That allows you to trace a typed value through the code whereas directly using `any` doesn't.

scytacki: I wasn't going to try to add some type checking here:
https://github.com/concord-consortium/activity-player/pull/145/files#diff-51cdda74ca9043dfc36956d3c97861042a6434794e12e2b8f80f6d760a3d9e16R92-R99

But I wasn't sure yet how to do it. At least that way someone editing the function can't try to reference a property that isn't defined by the interactive api.

scytacki: I just checked and aliases don't seem to provide any benefit more than documentation. 
If I have a function with an argument of InteractiveState and InteractiveState is aliased to any, then I can pass any type to this function. In other words it won't force me to cast to InteractiveState. That makes some sense when I think about it.

So adding this type doesn't seem worth it since it doesn't guarantee the documentation will be up-to-date. Maybe for an exposed API of a library it is useful, but in internal code I don't see the benefit.

dougmartin: Right - `any` matches anything for the compiler but when you are in an IDE you can right click on the type and find all references to it - that's what I meant by "trace a typed value though the code".  Sorry I wasn't clear.

src/utilities/embeddable-utils.test.ts
```diff
@@ -36,23 +36,26 @@ describe("Embeddable utility functions", () => {
...
-    expect(exportableAnswer.answer_text).toBe("test");
-    expect(exportableAnswer.answer).toBe("test");
-    expect(exportableAnswer.submitted).toBeNull();
-    const expectedReport = JSON.stringify({
-      mode: "report",
-      authoredState: `{"version":1,"questionType":"open_response","prompt":"<p>Write something:</p>"}`,
-      interactiveState: `{"answerType":"open_response_answer","answerText":"test"}`,
-      version: 1
+    // overwrite this id so we can do a full equals test below without a random value
+    exportableAnswer.id = "overwritten-answer-id";
```
kswenson, line 41: FYI, in CLUE I recently used an `omitDeep()` function to solve this problem:
```typescript
  expect(omitDeep(someObject, ["id"])).toEqual({ ... object without ids ... })
```
where `omitDeep()` looks like this
```typescript
// cf. https://github.com/lodash/lodash/issues/723#issuecomment-677383457
const omitDeep = (input: Record<string, unknown>, excludes: string[]): Record<string, unknown> => {
  return Object.entries(input).reduce((nextInput, [key, value]) => {
    const shouldExclude = excludes.includes(key);
    if (shouldExclude) return nextInput;

    if (Array.isArray(value)) {
      const arrValue = value;
      const nextValue = arrValue.map((arrItem) => {
        if (arrItem && typeof arrItem === "object") {
          return omitDeep(arrItem, excludes);
        }
        return arrItem;
      });
      nextInput[key] = nextValue;
      return nextInput;
    } else if (value && typeof value === "object") {
      nextInput[key] = omitDeep(value as Record<string, unknown>, excludes);
      return nextInput;
    }

    nextInput[key] = value;

    return nextInput;
  }, {} as Record<string, unknown>);
};

export default omitDeep;
```

I thought briefly about implementing it as a jest matcher so that one could use it like so:
```
  expect(someObject).toEqualOmitting(["id"], { ... obj without ids ...});
```
but decided it wasn't worth the extra effort at this point, but if it were being used across projects it might well be worth it.

src/utilities/embeddable-utils.ts
```diff
@@ -1,38 +1,100 @@
...
+// Instead of using this map, we look directly at the authoredState this ought to give us more
+// flexibility to support more special types in the report without needing to update
+// the activity player
+export const questionType = (rawAuthoredState: string | null | undefined): string => {
+  if (!rawAuthoredState) {
+    return "iframe_interactive";
+  }
+
+  // There is a IAuthoringMetadata type, but because we don't know if this has
+  // valid authored state or not, that type isn't not used here.
```
kswenson, line 23:
```suggestion
  // valid authored state or not, that type isn't used here.
```

src/utilities/embeddable-utils.ts
```diff
@@ -1,38 +1,100 @@
...
+
+export const remoteEndpoint = (): string => {
+  const portalData = getPortalData();
+  if (!portalData) {
+    return "";
+  }
+
+  // This is a way to verify the portalData is of type IPortalData
+  if (portalData.type === "authenticated") {
+    return portalData.runRemoteEndpoint;
```
kswenson, line 46: [Type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) can be useful for doing run-time type checks and then returning that information into the type system. Could potentially be useful here and/or with the untyped authored state above.

src/components/activity-page/managed-interactive/managed-interactive.tsx
kswenson: >I was going to try to add some type checking ... But I wasn't sure yet how to do it

I mentioned this in another comment elsewhere, but [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) can be useful for doing run-time type-checking in a way that feeds into the type system.

src/firebase-db.ts
scytacki: I just reverted this change so we can get type checking. I didn't try it but I figured any typing of this would add more lines of code than just duplicating the 3 fields twice.

src/firebase-db.test.ts
scytacki: done

src/components/activity-page/managed-interactive/managed-interactive.tsx
scytacki: I need to get this out the door, so I didn't do anything more with this `any` above.
If I had more time I'd at least switch it to `unknown` since that is more accurate than `any`.

I don't see the benefit of the alias since it doesn't guarantee searches for that type will find all of the instances since a developer could decide to use any (or unknown) or its alias and typescript wouldn't complain.

scytacki: I did add more type checking to `getAnswerWithMetadata` (referenced in the code URL above)

dougmartin (approved):
Looks good :+1:

PR #147, opened 2021-03-04
switch portal-report url based on the activity-player url

scytacki:
More details here:
https://www.pivotaltracker.com/story/show/176611702

### DIFF
--- a/src/utilities/report-utils.ts
+++ b/src/utilities/report-utils.ts
@@ -2,8 +2,33 @@ import { queryValue } from "../utilities/url-query";
 import { getPortalData } from "../firebase-db";
 import { IPortalData, firebaseAppName } from "../portal-api";
 
-// TODO: switch default to production report version before production deploy
-export const DEFAULT_PORTAL_REPORT_URL = "https://portal-report.concord.org/branch/master/index.html";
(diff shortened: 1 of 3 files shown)


src/utilities/report-utils.ts
```diff
@@ -2,8 +2,33 @@ import { queryValue } from "../utilities/url-query";
...
+  const portalReportUrlParam = queryValue("portalReport");
+  if (portalReportUrlParam) {
+    return portalReportUrlParam;
+  }
+
+  const { origin, pathname } = window.location;
+  // According to the spec an empty path like https://activity-player.concord.org
+  // will still have a pathname of "/", but just to be safe this checks for the
+  // falsey pathname
+  if(origin === "https://activity-player.concord.org" &&
```
dougmartin, line 24: Maybe move the url here to the top of the file as a constant?

dougmartin (approved):
Looks good :+1:

src/utilities/report-utils.ts
scytacki: Done. 

I was tempted to go further with this but stopped myself.
portal-api.ts also uses this approach of looking at the URL, so they could share the constant. But even better there could be some shared functions for determining if this is a root or version url. But I didn't want to change all of that code while trying to get this released today. :)

PR #151, opened 2021-03-08
Temporarily remove breaking Cypress tests to enable builds

dougmartin:
We need the offline-mode branch merges to build and they are not with these tests breaking.  A follow-on PR will fix the tests once we have a deployed offline-mode branch.


knowuh (approved):
Thanks Doug, as discussed this will help us setup tests on staging.

PR #155, opened 2021-03-10
Added offline logging storage [#176575357]

dougmartin:


src/components/app.tsx
```diff
@@ -645,12 +651,19 @@ export class App extends React.PureComponent<IProps, IState> {
...
       this.addActivityToOfflineManifest(this.state.offlineManifestAuthoringId, selectedActivity, url);
     }
   }
 
   private handleShowOfflineActivities = () => this.setState({ activity: undefined });
 
+  private trackOfflineActivityId(apiUrl: string) {
+    const resourceUrl = getResourceUrl(apiUrl);
+    // TODO: use resourceUrl instead of apiUrl in storage facade
+    TrackOfflineActivityId(apiUrl);
```
knowuh, line 665: I will change the storage interface to match this signature: `setActivity` 👍

src/lib/logger.ts
```diff
@@ -1,6 +1,8 @@
 import { v4 as uuid } from "uuid";
 import { DEBUG_LOGGER } from "../lib/debug";
 import { LaraGlobalType } from "../lara-plugin/index";
+import { dexieStorage } from "../dexie-storage";
+import { LogMessage } from "../types";
```
knowuh, line 5: We could probably have a religious war about this, but I like keeping the interfaces near their implementations. It prevents a `types` god-file which would be prone to code-churn and merge conflicts for one thing, and means you can typically import both the interfaces you need and the functions from a single import statement.  This also helps me out in autocomplete situations.

src/storage-facade.ts
```diff
@@ -115,18 +112,16 @@ const FireStoreStorageProvider: StorageInterface = {
   importStudentAnswersFromJSONFile: (studentAnswers: string, filename: string) => true
 };
 
-const indexDBConnection = new DexieStorage();
```
knowuh, line 118: 👍  Thank you, good catch.

knowuh (approved):
Looks really good, thanks.

src/lib/logger.ts
kswenson: A cross-cutting concern here is [Cumulative Cyclic Dependency](https://structure101.com/2014/03/10/why-cyclic-dependencies-explode-complexity/#:~:text=Cumulative%20Component%20Dependency%20(CCD)) -- if there are clients that only need the types/interfaces (which there often are) but have to import entire heavyweight modules just to get those definitions then the CCD goes up dramatically. When working on Fathom this was one of the metrics that we tracked regularly.

dougmartin: Related to what @kswenson said I had to move this to the common types file otherwise there was a module resolution order error because of interdependant requires (module A requiring module B and module B requiring module A).

PR #156, opened 2021-03-11
Changed offline mode from query parameter to hostname value [#177163781]

dougmartin:


dougmartin:
@scytacki / @knowuh I setup https://activity-player-offline.concord.org/ in as an alternate CNAME in the existing activity player CloudFront distribution.  This PR switches to checking the host vs the `offline` query parameter and only enables the other offline query parameters (like the manifests) when running from an offline host.

I also added a helper util to get the canonical hostname.  Right now it does nothing but we may want to rewrite `activity-player-offline.concord.org` to `activity-player.concord.org` in that function.

dougmartin:
I've also disabled the offline-manifest test for now.  This PR only enables offline parameters when running from the offline host so we'll need to figure out how to run this test on a separate port in GitHub actions.

scytacki:
@dougmartin I took the host-utils part of this PR and made a new PR on master with it:
https://github.com/concord-consortium/activity-player/pull/158

Here are some points of this other PR:
- the tool id should be the actual host as it will help us differentiate answers if we need to
- the determination of the production report-service and portal-report urls should be made based on both origins.
- when there is no activityUrl we should use the canonical host for the answerSourceKey and sourceKey, you were doing that here, but I just changed the conical host to convert activity-player-offline to activity-player

I didn't get to bringing over the tests. Or writing new tests. 

I thought it'd be better to put this part of it in master to reduce merge conflicts later, and it gave me a reason to clean up the report-service and portal-report code each of which was referring to the "https://activity-player.concord.org" string.

scytacki:
The PR based on master now has tests, and seems to be ready to go.

scytacki (approved):
Looks good to me.

scytacki (approved):
looks good.

PR #157, opened 2021-03-11
Added monitoring of network connection

dougmartin:


dougmartin:
@knowuh there wasn't a story for this but we will need to monitor the network connection so I wrote this which keeps a top level state variable for the network connection status.

knowuh (approved):
Looks great. Thanks

PR #158, opened 2021-03-11
Add support for an special offline domain

scytacki:
This also extracts the isProductionOrigin logic so the portal-api.ts and report-utils.ts aren't each implementing it

### DIFF
--- a/src/utilities/host-utils.ts
+++ b/src/utilities/host-utils.ts
@@ -5,7 +5,7 @@ export const getCanonicalHostname = () => {
     return "activity-player.concord.org";
   }
   return window.location.hostname;
-}
+};
 
 export const isProductionOrigin = (origin: string) => {
   return origin === "https://activity-player.concord.org" || origin === "https://activity-player-offline.concord.org";
--- a/src/utilities/host-utils.ts
+++ b/src/utilities/host-utils.ts
@@ -1,0 +1,12 @@
+export const isOfflineHost = (host: string) => (host === "localhost:11002") || (host === "activity-player-offline.concord.org");
+
+export const getCanonicalHostname = () => {
+  if(window.location.hostname === "activity-player-offline.concord.org") {
+    return "activity-player.concord.org";
+  }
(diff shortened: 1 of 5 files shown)


scytacki:
This is pretty much ready to go, but I'm stuck on some markdown escaping issues.

src/utilities/host-utils.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const getCanonicalHostname = () => {
+  if(window.location.hostname === "activity-player-offline.concord.org") {
+    return "activity-player.concord.org";
+  }
+  return window.location.hostname;
+};
+
+export const isProductionOrigin = (origin: string) => {
+  return origin === "https://activity-player.concord.org" || origin === "https://activity-player-offline.concord.org";
```
kswenson, line 11:
```suggestion
  return getCanonicalHostname(origin) === "https://activity-player.concord.org";
```
DRY

src/utilities/host-utils.ts
```diff
@@ -0,0 +1,12 @@
+export const isOfflineHost = (host: string) => (host === "localhost:11002") || (host === "activity-player-offline.concord.org");
```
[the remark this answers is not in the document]
scytacki: Does anyone use `127.0.0.1` when doing development?  
The localhost check is just for development.
Because this will trigger a service worker setup on each hostname, it won't work to switch between localhost and `127.0.0.1`, all of the resources would be cached on each domain individually.

I'd vote to just stick with localhost.
But I'll leave it up to @dougmartin

src/utilities/host-utils.ts
scytacki: This won't work because in one case it is hostname in the other case it the origin (which includes the protocol).
We could strip the protocol and then use getCanonicalHostname, but I'm reluctant to add any string based url manipulation. So this would mean making a URL object and getting the hostname. At which point just repeating our selves seems just as good.

PR #159, opened 2021-03-11
177304245 save student plugin state

knowuh:
You can checkout this branch and test it using an activity with a glossary, and append `&offline=true`. Glossary definitions should save to indexedDB.

Here is an example URL to try:
http://localhost:11000/index.html?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F20926.json&offline=true&runKey=3fa6900c-f040-4eb2-b1a0-9050c016c954

TODO: 
Replay plugin state saving data back to FireStore.

Note that this PR is based on `177147970-upload-when-launch-from-portal` -- If that branch is changed we will have to update this one too.

[#177304245] Plugin student state in APO


dougmartin (approved):
Looks good :+1:

dougmartin:
@scytacki and I decided to merge this without the replay to Firebase as it is demoable in the current form and the replay can be done in a future PR.  I'll update the PR story to reflect the merge.

PR #160, opened 2021-03-12
Moved network monitor to component mounting phase

dougmartin:
The `setState` call was being called in the callback which can't be done until the component is mounted.


scytacki (approved):
seems fine to me.

PR #162, opened 2021-03-12
Moved activity.json files into repo and added version numbers to filenames

dougmartin:
This will let us version both the manifest and activity files and reduce the chance of manifest breaking if a plugin or library interactive is updated


scytacki (approved):
Looks good to me.

PR #163, opened 2021-03-13
update several parts of offline-mode 

scytacki:
- split the url of the offlineActivity table into a resourceUrl and contentUrl: this makes it possible to load the activity json from a frozen place, but still use the resourceUrl as a id to find the report-structure in the report-service. The resourceUrl also makes it possible to to load new versions of the content without losing all of the answers stored in indexedDb.
- change report-source url parameter to sourceKey (not required for this work, but I've wanted to do it)
- change default sourceKey to use the canonical hostname. This also replaced the unnecessary getting of the hostname from the offering url instead of just getting it from window.location. The two should be the same. And it added the ability to override the sourceKey when running anonymously which wasn't supported well before.
- rename TrackOfflineActivityId to TrackOfflineResourceUrl since that is more accurate now
- add contentUrl query parameter to support the splitting of resourceUrl from contentUrl. In the url parameters the activity parameter cooresponds to the resourceUrl. This parameter wasn't renamed because it is assumed to be there by the portal-report codebase. Also this `activity` parameter is processed into the resourceUrl by stripping out the `api/v1` and `.json`. But if the activity parameter doesn't have those strings it also works fine. So if you are using the contentUrl then you can just set the activity parameter to the resourceUrl.
- change the answers table to store the resource_url field and use that to look up answers.
- change the answers table to look for `resource_url+question_id` since there this no guarruntee that a question_id will be unique outside of the a particular resource_url. For example an author could hand craft a resource instead of using LARA. 
- unify the Dexie schema into a single version. The use of multiple versions did not seem to be implemented right.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Inside of your `package.json` file:
 * firebaseApp={id}:  override default firebase app. https://activity-player.concord.org/ and https://activity-player-offline.concord.org/ without a path, defaults to `report-service-pro` every other url defaults to `report-service-dev`. For example https://activity-player.concord.org/branch/foo will use `report-service-dev` by default.
 * token={n}:         set by the portal when launching external activity, to authenticate with portal API
 * domain={n}:        set by the portal when launching external activity
-* report-source={id}: which source collection to save data to in firestore (defaults to own hostname)
+* sourceKey={id}:    which source collection to save data to in firestore (defaults to canonical hostname)
 * runkey={uuid}:     set by the app if we are running in anonymous datasaving mode
 * preview:           prevent running in anonymous datasaving mode
 * enableFirestorePersistence: uses local offline firestore cache only
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ Inside of your `package.json` file:
 
 * activity={id|url}:    load sample-activity {id} or load json from specified url
 * sequence={id|url}:    load sample-sequence {id} or load json from specified url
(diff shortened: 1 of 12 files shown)


scytacki:
I've been able to do a round trip with all of this using this approach:
1. install a single PC activity:
https://activity-player-offline.concord.org/branch/split-resourceUrl-and-contentUrl/?confirmOfflineManifestInstall=true&offlineManifest=precipitatingchange-test-v1
2. turn off computer network
3. open https://activity-player-offline.concord.org/branch/split-resourceUrl-and-contentUrl/
4. open PC activity
5. put in some answers
6. Assign "Scott TEST APO branch with public url" to a class
7. run this assignment as a student
8. verify your offline answers show up
9. on the last page click "Report my work" button
10. as a teacher click the "AP Report" button next to the assignment in the portal

The is some flakiness with these steps:
1. the install doesn't seem to work all of the time. I have had to do it multiple times before the files are actually cached and available when offline. The info in indexedDB is all right after the first install, but the application cache is not filled up.
2. multiple copies of an answer are are stored in indexedDB and then uploaded to firestore, usually just 2 copies per answer, but it can be more.  Because of this the report might not show the latest answer to the question.

dougmartin:
I've looked over the changes but I think @knowuh is the best person to do an actual review.

scytacki:
Add this to the readme: https://activity-player-offline.concord.org/branch/split-resourceUrl-and-contentUrl/?activity=https://authoring.staging.concord.org/activities/20926&contentUrl=offline-activities%2Fprecipitatingchange-test-v1.json

knowuh (approved):

PR #164, opened 2021-03-15
Add multi activity manifest

scytacki:
I put this together so we'd have a more complelling demo, and also I think we are going to find several issues when working with multiple activities. 

For example the ordering of the activities in the list doesn't match the manifest.  
Also the saving of the rawPortalData into the top level of localStorage means it would be applied to any of the activities when in reality the information is specific to a single assignment.


dougmartin (approved):
Looks good :+1: 

Re: ordering of the activities in the homepage.  An idea would be to add the manifest name to the activity name stored in indexdb along with an index from the manifest and then sort based on name + index.

PR #165, opened 2021-03-15
hacky untyped fix for the duplicate answers problem

scytacki:
This has no tests and isn't typed, but it might be worth merging since it seems to fix a critical bug when demoing this stuff. 
I'll leave it up to you to decide.

I think I problem is described pretty well in the story:
https://www.pivotaltracker.com/story/show/177333639

The technical issue is that the host code using the the storage system is assuming that the watchAnswer callback will be called after the createOrUpdateAnswer is complete.


dougmartin:
I think @knowuh should do the review on this.

knowuh (approved):
Thanks @scytacki this looks good.

PR #166, opened 2021-03-15
Type out query parameter keys, add `force_offline_data` key.

knowuh:
* Add types for query parameter keys for documentation purposes
* Add new query parameter key `force_offline_data` to force saving student data to indexedDB for testing.

[#177348535] -- Add optional`force_offline_data=true` query parameter

https://www.pivotaltracker.com/story/show/177348535

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,26 +107,36 @@ Inside of your `package.json` file:
 3. `cypress run --spec 'cypress/integration/examples/smoke-test.js'` will point to a smoke-test file rather than running all of the test files for a project.
 
 ## Url Parameters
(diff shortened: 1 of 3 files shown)


scytacki (changes requested):
Looks good. 
In addition to these changes, you should update the readme, it also lists all of the available url parameters.

scytacki (changes requested):
Now that I see them all together. I'm remember our decision to use camel case for parameters. So I think `force_ofline_data` should be changed to `__forceOflineData` for consistency. 

Also I think the other `__*` parameters should be moved into the Testing... section.

scytacki (approved):
looks good to me.

PR #168, opened 2021-03-15
Added wrapper methods for all console methods and fixed misc lint errors

dougmartin:


knowuh (approved):
Looks good to me.

scytacki (approved):
This seems fine for now.

When we have time, it seems like we should just globally disable the check on console.* by the linter for this project or at least for the offline-mode branch. If all I have to do is write consoleLog to get around the linter then there isn't much point in having the linter check for this.

kswenson:
What we've done in other projects is to configure lint to allow some methods (e.g. 'warn', 'error', etc.) that were intended to be left in place but to warn for others that were assumed to be temporary for debugging only (e.g. `log`). If the goal is to leave `console.logs` in place permanently, then I'd agree that the lint rule isn't serving much purpose.

dougmartin:
I think this is still a good way to go.  The linter will still find console.logs added by devs as they are testing and will leave in those meant to run in production.

dougmartin:
I'm going to close this PR and just turn off the warnings for console.logs in the lint config and put in another PR.

PR #169, opened 2021-03-15
Fix merge regression _currentOfflineActivityId → _currentOfflineResourceURL

knowuh:
Not sure how this happened, but we had lingering instances of `_currentOfflineActivityId` after all the recent merging was done.


scytacki (approved):
looks good to me

the answerWatchers stuff was added in parallel to the switch to currentOfflineResourceUrl change so it makes sense that it would have be a problem.

PR #170, opened 2021-03-15
Save plugin state when exporting student data to JSON

knowuh:
This will make saving plugin state easier to demo tomorrow.

It just adds the plugin states to the JSON export of student data.


scytacki (approved):
Looks good to me.

PR #171, opened 2021-03-15
Add 0 as valid cacheable response

dougmartin:
The 0 status code is returned for opaque responses.  Since our fetch based cache filler use the no-cors option we get opaque responses for all external non-cors enabled servers.


scytacki (approved):
seems fine to me

PR #172, opened 2021-03-16
Disable warnings for console logs in production build

dougmartin:


knowuh (approved):
This looks fine to me. It will reduce noise and help us focus on more important issues.

scytacki:
Should clean up the lint exceptions for the existing console logs.

kswenson:
Is this considered the long-term solution or just a temporary one? If the latter, it could be changed to
```javascript
      "no-console": ["off", { allow: ["warn", "error"] }],
```
with a comment indicating the conditions under which it should be reenabled.

PR #174, opened 2021-03-16
Added save of manifest name and activity order [#177347793]

dougmartin:
Also added missing tests for offline-manifest-api (minus dexie tests)


knowuh (approved):
LGTM

PR #176, opened 2021-03-16
Summary Table on Completion Page

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177287957

[#177287957]


src/components/activity-completion/summary-table.tsx
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <table className="summary-table" data-cy="summary-table">
+      <thead>
+        <tr>
+          <th>Question</th>
+          <th>Complete</th>
+        </tr>
+      </thead>
+      <tbody>
+      {questionsStatus.map((question: {number: number, page: number, prompt: string, answered: boolean}, index) => {
```
knowuh, line 23: You could extract the interface for a single `interface IQuestionStatus {number: number, page: number, prompt: string, answered: boolean}` and then make `interface IProps { questionsStatus: Array< IQuestionStatus> }` — You might then be able to reuse `IQuestionStatus` in `completion-page-content.tsx`.  Just a thought.

knowuh (approved):
This looks good to me.  Thanks.

I had one suggested change if you have time, otherwise we can just merge.

PR #180, opened 2021-03-18
This is a WIP of handling the service worker states

scytacki:
The app.tsx is getting big so it'd be great to separate this out.
Right now there is a mix of workbox and native serviceworker apis, hopefully that can be improved.

It also doesn't handle showing the activated state.
We might want to call claimClients once we are activated so we don't need to reload.


scytacki:
This is replaced by this PR: https://github.com/concord-consortium/activity-player/pull/186

PR #181, opened 2021-03-18
`IPluginSyncRequest` and handler types. Sync Tracker

knowuh:
DataSyncTracker tracks the status of various plugins and promises while syncing data.
Tests include some fake plugins responding to events.

Part of
[#177304920] — Plugin API changes in Activity Player

@dougmartin this is all I could get done Today.

I still need to add a DataSyncTracker to the StorageFacade sync method.  But that should be easy.


dougmartin (approved):
Looks good :+1:

PR #182, opened 2021-03-19
Open teacher edition specific page

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177207076

[#177207076]

Add console logging to aid troubleshooting.

### DIFF
--- a/src/components/expandable-content/expandable-container.tsx
+++ b/src/components/expandable-content/expandable-container.tsx
@@ -27,10 +27,12 @@ export const ExpandableContainer: React.FC<IProps> = (props) => {
   return (
     <div className="expandable-container" id="expandable-container" key={pageNumber} data-cy="expandable-container">
       { teacherEditionMode && sideTips.map((sideTip: any) =>
+          // Technically, the EmbeddablePluginSideTip won't be rendered if the plugins aren't loaded yet, 
+          // but passing pluginLoaded into it makes it more consistent with the other plugins components
           <EmbeddablePluginSideTip
             key={sideTip.embeddable.ref_id}
             embeddable={sideTip.embeddable}
-            {...(pluginsLoaded && {pluginsLoaded: pluginsLoaded})}
+            pluginsLoaded={pluginsLoaded}
           />)
       }
       { sidebars && sidebars.length > 0 &&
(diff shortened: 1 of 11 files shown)


src/components/expandable-content/expandable-container.tsx
```diff
@@ -30,6 +30,7 @@ export const ExpandableContainer: React.FC<IProps> = (props) => {
           <EmbeddablePluginSideTip
             key={sideTip.embeddable.ref_id}
             embeddable={sideTip.embeddable}
+            {...(pluginsLoaded && {pluginsLoaded: pluginsLoaded})}
```
scytacki, line 33: perhaps you tried it, but I think you do 
`pluginsLoaded={pluginsLoaded}` here.
or perhaps even
`{..., pluginsLoaded}`

scytacki: I see that in theory you don't need to add the pluginsLoaded property, because on line 24, the sideTips array will be empty until pluginsLoaded is true. But It seem good for consistency to add this reduancy.
I'd just suggest adding a comment something like:
"technically the EmbeddablePluginSideTip won't be rendered if the plugins are not loaded yet, but passing pluginLoaded into it makes it more consistent with the rest of the plugins components"

src/lara-plugin/plugins/plugins.ts
```diff
@@ -45,6 +45,13 @@ export const initPlugin = (label: string, options: IPluginContextOptions) => {
 };
 
 const initRuntimePlugin = (label: string, options: IPluginRuntimeContextOptions) => {
+  // tslint:disable-next-line:no-console
+  console.info("initRuntimePlugin called with label: ", label);
+  if (!pluginClasses[label]) {
+    // tslint:disable-next-line:no-console
+    console.error("No pluginClasses label.");
```
scytacki, line 52: it is probably good to remove the rest of this console.log statements. But i'd keep this check and this console.error. That way an issue in the plugin system won't break the whole app, and we should still see the error in the console. 
I'd change the error message to include the label that couldn't be found. so maybe something like:
``console.error(`initRuntimePlugin cannot find plugin class with label: ${label}, plugin instance is not loaded`);``

src/components/expandable-content/expandable-container.tsx
emcelroy: I did try `pluginsLoaded={pluginsLoaded}` first, but it caused an error. I've tried it again though and now it works. I'm not sure what changed, but it seems to be fine.

scytacki (approved):
looks good to me

PR #184, opened 2021-03-19
 Add plugin API support for offline [#177304920]

dougmartin:
NOTE: much of this change is drilling the `offlineMode` top level state variable down the props to all the components (and updating their tests).  This could have also been done with a context but there were no contexts in use already so I went with prop drilling instead.

### DIFF
--- a/src/components/activity-page/activity-page-content.test.tsx
+++ b/src/components/activity-page/activity-page-content.test.tsx
@@ -17,6 +17,7 @@ describe("Activity Page Content component", () => {
       totalPreviousQuestions={5}
       setNavigation={stubFunction}
       pluginsLoaded={true}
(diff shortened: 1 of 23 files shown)


scytacki (approved):
Seems fine to me.

kswenson (commented):
It may be water under the bridge at this point, but would it make sense to consider online/offline mode the kind of global that could be put in React context so you don't need to drill it down through all of the component props?

dougmartin:
@kswenson see the note in the description for this PR:

> NOTE: much of this change is drilling the offlineMode top level state variable down the props to all the components (and updating their tests). This could have also been done with a context but there were no contexts in use already so I went with prop drilling instead.

PR #186, opened 2021-03-22
177370190 workbox updates

scytacki:
- import workbox-window/index so we can see the debugging log messages of workbox-window
- use messageSkipWaiting instead of sending our own messages
- put up page loading message so the offline manifest isn't cached before the service worker is controlling the page

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,11 @@
       "<rootDir>src/setupTests.js"
     ],
     "transform": {
-      "^.+\\.tsx?$": "ts-jest"
+      "^.+\\.[jt]sx?$": "ts-jest"
     },
+    "transformIgnorePatterns": [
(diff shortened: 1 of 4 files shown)


scytacki:
This is begging for tests, but I'm not sure how to add them. Most of the logic is mixed into the `app.tsx`. Probably the best is to figure out how to extact this serviceworker stuff from app.tsx. Then perhaps we can mock a few things in order to test the logic without using a real serviceworker.

But in order to get this out as a demo for I don't think we should wait for tests to merge it.

dougmartin (approved):
Looks good :+1:   With the lack of tests this will need to be carefully QAed to check all the state changes.

scytacki:
I'm going to try to document the various cases we should be handling and how to manually test them.

PR #187, opened 2021-03-22
WIP: Add glossary test activity [#177304920]

dougmartin:
This is a WIP branch as we will need to remove the temp code to trigger the plugin syncs


knowuh:
We don't need these changes anymore. Closing.

PR #188, opened 2021-03-22
Notify offline plugins to sync & update Report button status.

knowuh:
This commit sends notifications to plugins about requests to synchronize data with FireStore. 
It also adds some UI hints that an update is in progress, and when an update fails.

@emcelroy  -- The UI for progress feedback could probably use your help eventually.

[#177144932]

### DIFF
--- a/src/storage/storage-facade.ts
+++ b/src/storage/storage-facade.ts
@@ -4,6 +4,7 @@ import { fetchPortalData, IAnonymousPortalData, IPortalData } from "../portal-ap
 import { IExportableAnswerMetadata } from "../types";
 import { dexieStorage, IDexiePluginRecord, kOfflineAnswerSchemaVersion } from "./dexie-storage";
 import { refIdToAnswersQuestionId } from "../utilities/embeddable-utils";
+import { DataSyncTracker } from "./data-sync-tracker";
 
 export interface IInitStorageParams {
   name: FirebaseImp.FirebaseAppName,
@@ -107,7 +108,7 @@ export interface IStorageInterface {
   exportActivityToJSON: (activityId?: string) => Promise<ExportableActivity>,
   importStudentAnswersFromJSONFile: (studentAnswers: string, filename: string) => boolean,
   canSyncData(): boolean,
-  syncData(): void
+  syncData(): Promise<boolean>
 }
 
 class FireStoreStorageProvider implements IStorageInterface {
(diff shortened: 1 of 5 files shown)


src/storage/storage-facade.ts
```diff
@@ -400,48 +401,85 @@ class DexieStorageProvider implements IStorageInterface {
...
-      }
+      // DataSyncTracker will emit an event that tells plugins
+      // that we are online, and its time to save data:
+      const mySyncTracker = new DataSyncTracker(60 * 30, 5);
+
+      // Save student answers ...
+      const answerSavingPromise = this.ensureFirebaseConnection()
+        .then((fsProvider)  => {
+          dexieStorage.answers
+            .where({resource_url: _currentOfflineResourceUrl})
```
scytacki, line 440: The refactor lost my FIXME comment:
> // FIXME: this should be checking the resourceUrl of the rawPoralData
> // we don't want to send answers for the wrong resource url to an assignment

The currentOfflineResourceUrl might not match the assignment this being uploaded to. That is based on the info in rawPortalData.

scytacki (changes requested):
Looks good to me. I just found one comment that got lost in the refactor.

src/storage/storage-facade.ts
```diff
@@ -400,48 +401,85 @@ class DexieStorageProvider implements IStorageInterface {
...
-      }
+      // DataSyncTracker will emit an event that tells plugins
+      // that we are online, and its time to save data:
+      const mySyncTracker = new DataSyncTracker(60 * 30, 5);
+      const portalResourceUrl = portalData.resourceUrl;
+      // Save student answers ...
+      const answerSavingPromise = this.ensureFirebaseConnection()
+        .then((fsProvider)  => {
+          dexieStorage.answers
+            .where({resource_url: portalResourceUrl})
```
scytacki, line 440: This might still result in some weird behavior where I'm looking at an activity in the APO that is different than the portalData.resourceUrl and then doing a sync will cause it to upload data for a different activity.
It might be safer to raise an error if currentOfflineResourceUrl doesn't match the portalResourceUrl.

scytacki (changes requested):
The change from my previous comment might result in un-expected user behavior. When we discussed this before it seemed like the best option was to not enable 'report my work' button on activities that don't match the portalData. 
So if that is our plan, it might be better for this backend code to just raise an exception instead of uploading a different set of answers.

src/storage/storage-facade.ts
knowuh: We can disable the button in that case where the current_activity doesn't match the portals resource URL too.

scytacki: The new change looks good.

scytacki (approved):

PR #189, opened 2021-03-23
Added automatic caching of glossary assets [#177398187]

dougmartin:


src/utilities/activity-utils.test.ts
```diff
@@ -118,4 +124,65 @@ describe("Activity utility functions", () => {
...
+    // then test an activity with a glossary
+    urls = await getAllUrlsInActivity(glossaryActivity);
+    expect(urls).toEqual([
+      "https://models-resources.concord.org/question-interactives/branch/master/open-response/",
+      "https://teacher-edition-tips-plugin.concord.org/version/v3.5.6/plugin.js",
+      "https://example.com/manifest.json",
+      "https://glossary-plugin.concord.org/plugin.js",
+      "https://glossary-plugin.concord.org/manifest.json",
+      "https://token-service-files.s3.amazonaws.com/glossary-plugin/ZWjh4IciSnBkWfu4gODs/5cacbe00-1c44-11ea-90e3-39c0ba8d079c-sticky note.svg",
+      "https://token-service-files.s3.amazonaws.com/glossary-plugin/ZWjh4IciSnBkWfu4gODs/5d298f20-1c44-11ea-90e3-39c0ba8d079c-IMG_8603.jpeg"
```
scytacki, line 185: Shouldn't this also include the glossary json itself?
`https://token-service-files.s3.amazonaws.com/glossary-plugin/ISnn8j8r2veEFjPCx3XH/glossary.json`

scytacki (changes requested):
It seems like the glossary handling isn't adding the glossary json itself to the cache.

src/utilities/activity-utils.test.ts
dougmartin: Good catch.  I've added it in this follow on commit:  https://github.com/concord-consortium/activity-player/pull/189/commits/17c79fddb8c62bd7b3a7e71723691487941b74f8

I also updated the mocked response to use the same id so it won't potentially confuse folks in the future.

scytacki (approved):
Looks good.

PR #192, opened 2021-03-23
switch to cors requests when loading offline manifest

scytacki:
doing this exposed an issue with the existing code. There are two css files referenced
by index.html that are referenced before we start installing the offline manifest.
These references were breaking the cache.
My solution is to add them to the pre-cache so they are available before the index.html
request even happens (when the page is controlled).

This also adds the range-request plugin which is needed to support videos.
And it removes a broken url from production pc manifest.

I think part of this is that by using cors we are see any download errors. With no-cors
those download errors are cached as if they succeeded so we never knew they failed

Note: I ran `npm install`, but no changes showed up in package-lock.json

### DIFF
--- a/src/service-worker.ts
+++ b/src/service-worker.ts
@@ -4,6 +4,7 @@ import { precacheAndRoute } from "workbox-precaching";
 import { registerRoute } from "workbox-routing";
 import { StaleWhileRevalidate } from "workbox-strategies";
 import { CacheableResponsePlugin } from "workbox-cacheable-response";
+import { RangeRequestsPlugin } from "workbox-range-requests";
 
 const ignoredGets: RegExp[] = [
   /\/sockjs-node\/info/,             // webpack-dev-server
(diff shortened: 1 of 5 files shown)


src/service-worker.ts
```diff
@@ -12,9 +13,21 @@ const ignoredGets: RegExp[] = [
...
-  ignoreURLParametersMatching: [/.*/]
+const precacheEntries = [
+    // Add URLs that are referenced directly in index.html
+    {url: "https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css", revision: null},
+    {url: "https://fonts.googleapis.com/css2?family=Lato:wght@400;700;900&display=swap", revision: null},
+  ].concat((self as any).__WB_MANIFEST);
+
+precacheAndRoute(precacheEntries, {
+  // Ignore most URL parameters. We want to ignore the runKey on index.html
+  // as well as other paramters on index.html
```
kswenson, line 24:
```suggestion
  // as well as other parameters on index.html
```

src/service-worker.ts
```diff
@@ -12,9 +13,21 @@ const ignoredGets: RegExp[] = [
...
+  // The exception are the params needed by the lato request above
+  // This approach of not ignoring family and dispaly is hacking.
```
kswenson, line 26:
```suggestion
  // This approach of not ignoring family and display is hacking.
```

src/service-worker.ts
```diff
@@ -12,9 +13,21 @@ const ignoredGets: RegExp[] = [
...
+  // It'd be better if we just ignored all params on index.html and let other
+  // params go through. We are about to change this precaching code, so it
+  // it doesn't seem worth the effort to find a better way
```
kswenson, line 29:
```suggestion
  // doesn't seem worth the effort to find a better way
```

dougmartin (approved):
Looks good :+1:

PR #193, opened 2021-03-24
Search in author_data when looking for external urls to cache [#177449302]

dougmartin:


src/utilities/activity-utils.ts
```diff
@@ -232,7 +239,6 @@ export const getAllUrlsInActivity = async (activity: Activity, urls: string[] =
       if (authorData?.s3Url) {
         const response = await fetch(authorData.s3Url);
         const glossaryJson = await response.json();
-        urls.push(authorData.s3Url);
```
dougmartin, line 235: This was removed as it is now being picked up with the other urls in the author_data

scytacki (approved):
Looks good to me.

src/utilities/activity-utils.ts
```diff
@@ -214,10 +214,17 @@ export const removeDuplicateUrls = (urls: string[]) => urls.filter((url, index)
...
+    walkObject(object, (s, key) => {
       if (isExternalOrModelsResourcesUrl(s)) {
         urls.push(s);
       }
+      // add all external urls in the stringified author data json
+      if (key === "author_data") {
+        try {
+          const authorData = JSON.parse(s);
+          addExternalUrls(authorData);
+        } catch (e) {} // eslint-disable-line no-empty
```
kswenson, line 226: You can also just put a comment like `// ignore errors` inside the braces.

PR #195, opened 2021-03-24
Remove red bar when running in offline-mode branch [#177458105]

dougmartin:
NOTE: this adds the offline-mode branch as a production url

### DIFF
--- a/src/portal-api.ts
+++ b/src/portal-api.ts
@@ -233,7 +233,7 @@ export const firebaseAppName = ():FirebaseAppName => {
     return _firebaseAppName;
   }
 
-  if (isProduction(window.location)) {
+  if (isProduction(window.location, {allowVersions: false})) {
     _firebaseAppName = "report-service-pro";
   } else {
     _firebaseAppName = "report-service-dev";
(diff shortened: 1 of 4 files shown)


scytacki (changes requested):
There was a nuance here that was dropped. 
The firebaseAppName was not report-service-pro even with version urls.
This is because we don't actually use version'd urls in production. We might use them for testing a version before updating the production branch. And I thought we wouldn't want that testing data making its way into report-service-pro. Also it would prevent testers from sending out versioned urls to their friends and their friends thinking their data was safe. We might change the schema and the version'd url wouldn't get updated.

The other place this url check was made was to determine which portal-report version to use. In that case it seemed we'd want versioned urls to use the production report, since that is what would happen when the version was released to production, and it wouldn't have any side effects.

dougmartin:
@scytacki I've added a follow on PR here:  https://github.com/concord-consortium/activity-player/pull/195/commits/22e93ee62e6c1cd19e9a69581f93d26b719108fc

I should have called out in the PR description that I was "regularizing" the check as I saw that we checked version in one and not in the other.  I thought that was a bug and not a feature :)

scytacki (approved):
looks good

PR #196, opened 2021-03-24
Don't add runKey query param in offline mode [#177374559]

dougmartin:


scytacki (changes requested):
I think the effect of this will be to store answers in indexedDB with different runKeys for each time a user opens an activity.  I think those will get removed when uploading the answers to firestore, but if we can do a little bit more on this so the indexedDB was consistent that would be better. 

Perhaps we just define a fixed dummy runkey for offline like what is done in `preview`?

dougmartin:
@scytacki Good catch. I've set a constant `runKey` for ofline mode in this follow on commit: https://github.com/concord-consortium/activity-player/pull/196/commits/50b4d501411ffc7107cb7da882c1df16961af09b

src/portal-api.ts
```diff
@@ -465,8 +465,13 @@ export const anonymousPortalData = (preview: boolean) => {
   } else {
     runKey = queryValue("runKey");
     if (!runKey) {
-      runKey = uuidv4();
-      setQueryValue("runKey", runKey);
+      if (isOfflineHost()) {
+        runKey = "offline";
```
scytacki, line 469: Perhaps this is nit picky, but wouldn't it be better if you change changed line 462 to 
`if (preview || isOfflineHost()) {`

But perhaps you wanted the runKey of `offline` to be overridden if some passed a query parameter?

scytacki (approved):
This works. I just have a comment to clarify the logic.

src/portal-api.ts
dougmartin: @scytacki  I went back and forth on this.  I had a helper like `const createRunKey = (fallback: string) => isOfflineHost() ? "offline" : fallback;` and I used that on lines 462 and 468 but then looking back at the call chain I thought if we have an explicit `preview` url parameter we shouldn't pollute any offline data we have saved.

scytacki: Sounds fine to me.

PR #197, opened 2021-03-24
Add service worker version to the bottom of the page [#177458051]

dougmartin:


src/components/app.tsx
```diff
@@ -409,7 +425,10 @@ export class App extends React.PureComponent<IProps, IState> {
             { showOfflineNav && <OfflineNav onOfflineActivities={this.handleShowOfflineActivities} /> }
             { this.renderContent() }
             { this.state.showThemeButtons && <ThemeButtons/>}
-            <div className="version-info" data-cy="version-info">{(window as any).__appVersionInfo || "(No Version Info)"}</div>
+            <div className="version-info" data-cy="version-info">
+              Application: {appVersionInfo || "No Version Info"}
+              {serviceWorkerVersionInfo && ` | Service Worker: ${serviceWorkerVersionInfo} ${mismatchedVersions ? "❌" : ""}`}
```
dougmartin, line 430: I added this red X so we can ask "Is there a red X at the bottom of the page?" instead of asking them to compare the version numbers themselves.  I did have a ✓ if the versions matched but it looked out of place so I removed it.

knowuh (approved):
Looks good to me.

scytacki (changes requested):
I can think of a few cases where this version might not be accurate, but I think that is a different (bigger) PR to add additional info to the footer about the versions of the two possible service workers, controlling and not-yet-controlling, and the status of the not-yet-controlling worker. 

Thinking about the red x. I think we should probably take it out here is why:
In our quest to keep the automatic download size to as small as possible, I think we are going to have to support the case where the service worker gets updated, but the main application is not updated. The reason is because we don't have control when the service worker gets updated, the browser will check for updates on each navigation event. So if a student is on 3G at home, this update check will succeed and the next launch of the APO will activate this new worker.  But we won't want this trigger a full download of the main app since that is currently 7MB. So in that case the student is going to have mismatched versions until they get back to cheap/fast wifi and run the update link.

dougmartin:
@scytacki I've removed the red X in this follow on commit: https://github.com/concord-consortium/activity-player/pull/197/commits/23aaad2957a116b535a1764aaad48b11ca092300

scytacki (approved):
👍

PR #198, opened 2021-03-24
Hide "Show My Work" when we can't run a report.

knowuh:
On the completion page, we gray out/disable the "Show My Work" button unless we can run a report.

This should have been a small change, but I got derailed by a failing test, and then remembered that we were supposed to disable a button instead of removing it.

[#177461358] -- Hide "Show My Work" when we can't run a report

PT Story: https://www.pivotaltracker.com/story/show/177461358

### DIFF
--- a/src/components/activity-completion/completion-page-content.tsx
+++ b/src/components/activity-completion/completion-page-content.tsx
@@ -33,6 +33,7 @@ export const CompletionPageContent: React.FC<IProps> = (props) => {
     onShowSequence } = props;
 
   const [answers, setAnswers] = useState<any>();
+  const [canProvideStudentReport, setCanProvideStudentReport] = useState<boolean>();
 
   const handleExit = () => {
     if (sequence) {
@@ -85,6 +86,7 @@ export const CompletionPageContent: React.FC<IProps> = (props) => {
 
   useEffect(() => {
     const storage = getStorage();
+    setCanProvideStudentReport(storage.canProvideStudentReport());
     storage.watchAllAnswers(answerMetas => {
       setAnswers(answerMetas);
     });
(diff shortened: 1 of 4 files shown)


scytacki (approved):
looks good to me

knowuh:
@scytacki sorry -- I added one more commit.  I forgot that instead of removing the button we were supposed to just disable it.  Attached are the screen shots from the 4 states we could be in:

- Online
- Online but in preview mode
- Offline with tokens
- Offline without Tokens
Screen Shot 2021-03-24 at 2 42 55 PM
Screen Shot 2021-03-24 at 2 42 36 PM
Screen Shot 2021-03-24 at 2 42 20 PM
Screen Shot 2021-03-24 at 2 41 39 PM

src/components/activity-completion/completion-page-content.tsx
```diff
@@ -165,10 +167,25 @@ export const CompletionPageContent: React.FC<IProps> = (props) => {
...
+                      className="button show-my-work disabled"
+                      disabled={true}>
+                        <IconCompletion width={24} height={24} />
+                        Show My Work
+                    </button>
+              }
+            {(!sequence || isLastActivityInSequence) &&
               <div className="exit-button">
-                <span>or</span>
+                { showStudentReportButton && <span>or</span> }
```
scytacki, line 188: Seems like the `or` should be shown all of the time now that the button is just disabled.

scytacki (changes requested):
one issue with the "or" text

knowuh:
I was wondering about that "or" button.  Any way, its back in.

Screen Shot 2021-03-24 at 3 35 21 PM
Screen Shot 2021-03-24 at 3 34 28 PM

scytacki (approved):
Looks good.

PR #199, opened 2021-03-24
try using no-store when loading cache

scytacki:
This seems to fix this issue: https://www.pivotaltracker.com/story/show/177481513
I cannot duplicate it on localhost, so I pushed this branch to test it out. 

It is hard to test remotely as well, but I'm pretty confident that it fixes the problem.
I ran the same steps in the story but used 
https://activity-player-offline.concord.org/branch/test-no-store/?offlineManifest=precipitating-change-v1&confirmOfflineManifestInstall=true&firebaseApp=report-service-pro
for the offline part and it worked.

When testing you have to be careful because different branches share the same `cachedGets` cache. So you need to start completely fresh each time.


dougmartin (approved):
Looks good :+1: 

I tested it locally and it worked.

PR #201, opened 2021-03-25
Upload Log Data to Log Manager [#177288011]

dougmartin:


knowuh (approved):
Looks good, thanks

scytacki:
I'll defer to Noah on this, so I can keep going on my installer work.

I'd note that we are now starting to treat the offline-mode branch as production'ish branch and soon it will be a production branch. So I'm thinking we should start a new `offline-mode-dev` (or something like  that).  And start merging PRs into there before we update offline-mode.

PR #202, opened 2021-03-25
Changed test glossary activity to use AK glossary [#177479101]

dougmartin:


scytacki (approved):
Looks good to me.

I'd note that here is another example where cloudfront is being bypassed.
I added this story: https://www.pivotaltracker.com/story/show/177478486

PR #204, opened 2021-03-26
New install flow

scytacki:
The user needs to load install.html. 
This triggers the installation of the application fines and content files if a `?offlineManifest=[id]` is included.
The user can then go offline and visit index.html and no extra requests should
be made over the network.

PT Story:
https://www.pivotaltracker.com/story/show/177164109


scytacki:
@dougmartin this is ready for a preliminary review. 
I still want to clean up the FIXME and TODO comments.

src/components/install-app.tsx
```diff
@@ -0,0 +1,159 @@
...
+            this.setState({serviceWorkerVersionInfo: "Invalid response!"});
+          }
+        })
+        .catch(() => {
+          this.setState({serviceWorkerVersionInfo: "No response!"});
+        });
+    });
+
+    if (offlineManifestId) {
+      getOfflineManifest(offlineManifestId)
```
dougmartin, line 81: Should this be done in `cacheApplication` so that we know the service worker is ready to cache it or if not can you add a comment about why it is called here instead of there?

src/components/install-app.tsx
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+
+  cacheApplication() {
+    const wbManifest = (window as any).__wbManifest;
+    // const offlineManifest: OfflineManifest = { name: "Application Manifest", activities: [], cacheList: [] };
+
+    // FIXME: this caching is going to happen after the initial page load
+    // that means that the manifest.json and index.html will be cached by chrome as it parses the
+    // manifest.  The index.html request comes from Chrome's new installable dection that makes an offline request
```
dougmartin, line 101: dection?

src/install.html
```diff
@@ -0,0 +1,22 @@
...
+    <link rel="manifest" href="manifest.json">
+    <script>
+      window.__appVersionInfo = "__APP_VERSION_INFO__";
+    </script>
+    <!-- Currently the revision is handled manually, it should be bumped when a
+        new manifest is available. It would be better for webpack to generate this
+        destination name and inject it into the file. However currently the manifest
+        is built during the application webpack config, so the seperate install
+        config can't inject its value -->
+    <script src="app-manifest.js?revision=2"></script>
```
dougmartin, line 17: Maybe good to add a TODO or FIXME here that will show up on a search across the codebase.

src/offline-manifest-api.test.ts
```diff
@@ -28,7 +28,7 @@ describe("offline manifest api", () => {
     });
   });
 
-  it("handles #cacheOfflineManifest", (done) => {
+  it.skip("handles #cacheOfflineManifest", (done) => {
```
dougmartin, line 31: Mark this with a FIXME comment?

dougmartin (commented):
Looks good so far.  I had a few comments about documentation.  Do you also plan to update the readme with the install flow as part of this PR?

src/install.html
scytacki: @dougmartin this is updated now and a hash is added. It isn't perfect, so I left a comment, but didn't tag it with FIXME because I don't think we are going to be fixing it anytime soon.

src/components/install-app.tsx
scytacki: I moved it to cacheApplication and added more comments around it.

src/components/install-app.tsx
scytacki: I updated this comment, might have added more spelling mistakes in the process though. :)

src/offline-manifest-api.test.ts
scytacki: I added a FIXME with more info here.

scytacki:
Definitely could use more tests, but I think this will have to do for now.

dougmartin (approved):
Looks good :+1:

PR #205, opened 2021-03-27
seems like offlineMode should be passed through

scytacki:


dougmartin (approved):
Good catch.  I think I did this when I was updating the tests.

PR #207, opened 2021-03-29
Switch from ik language code to repurposed mr code

dougmartin:
The mr code was already being used for Inuit.  While Inupiaq has a valid language code of ik we will reuse mr to reuse the glossaries.


scytacki (approved):
Seems OK. I don't really understand the history of this, but I don't think I need to.

PR #208, opened 2021-03-29
Rewrite urls inside serialized components of the authored state.

knowuh:
When walking objects rewriting URLs, we now also look inside nested JSON expressions.

[#177163989] -- Support Carousel Interactive
PT:  https://www.pivotaltracker.com/story/show/177163989

### DIFF
--- a/src/utilities/activity-utils.ts
+++ b/src/utilities/activity-utils.ts
@@ -2,6 +2,7 @@ import { Page, Activity, EmbeddableWrapper } from "../types";
 import { SidebarConfiguration } from "../components/page-sidebar/sidebar-wrapper";
 import { isQuestion as isEmbeddableQuestion } from "./embeddable-utils";
 import { runningInCypress } from "./cypress";
+import { isGenerator } from "mobx/dist/internal";
 
 export enum ActivityLayouts {
   MultiplePages = 0,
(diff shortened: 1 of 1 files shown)


scytacki (approved):
Looks good to me. It'd be nice to add a test for this, but it seems better to get this merged first.

scytacki:
I think I'll go ahead and merge this, so I can then update my PR to resolve the small merge conflict this will bring.  @dougmartin if you find issues they'll have to be address in a new PR.

src/utilities/activity-utils.ts
```diff
@@ -2,6 +2,7 @@ import { Page, Activity, EmbeddableWrapper } from "../types";
 import { SidebarConfiguration } from "../components/page-sidebar/sidebar-wrapper";
 import { isQuestion as isEmbeddableQuestion } from "./embeddable-utils";
 import { runningInCypress } from "./cypress";
+import { isGenerator } from "mobx/dist/internal";
```
dougmartin, line 5: Looks like this import isn't being used.

dougmartin (changes requested):
You should remove that unneeded import to reduce the bundle size.

src/utilities/activity-utils.ts
knowuh: Thanks for spotting this.  I am not sure how that got inserted, possibly some eager VS Code autocomplete. Its removed in d23f524

PR #209, opened 2021-03-30
Changed local storage to include resourceUrl when saving plugin states

dougmartin:


src/components/app.tsx
```diff
@@ -378,8 +378,8 @@ export class App extends React.PureComponent<IProps, IState> {
       this.setState(newState as IState);
 
       this.LARA = initializeLara();
-      if (activity) {
-        loadLearnerPluginState(activity, teacherEditionMode).then(() => {
+      if (activity && resourceUrl) {
```
knowuh, line 381: We wont load plugin state if we don't have a resourceUrl. This seems fine, but now I am wondering when we won't have a resourceUrl ... 🤔

src/storage/dexie-storage.ts
```diff
@@ -17,7 +18,7 @@ export class DexieStorage extends Dexie {
   logs: Dexie.Table<LogMessage, number>;
   offlineActivities: Dexie.Table<OfflineActivity, string>;
   answers: Dexie.Table<IIndexedDBAnswer, string>; // number = type of the primary key
-  pluginStates: Dexie.Table<IDexiePluginRecord, number>;
+  savedPluginStates: Dexie.Table<IDexiePluginRecord, number>;
```
knowuh: Change the primary key to a string for savedPluginStates

knowuh (approved):
Looks good. (sync review w/ DM)

PR #211, opened 2021-03-30
Switch glossary to use s3 url

dougmartin:
This was done so that it isn't included in the build and it isn't rewriten to a local models-resources url so the glossary dashboard can load it.


knowuh (approved):
Looks good.

scytacki (approved):
👍

PR #214, opened 2021-03-30
For reference only:  Punt automatic rewrite

knowuh:
@scytacki -- We talked about punting automatic models-resources URL rewrite. This does that.

I haven't tested it extensively.

We would merge this temporarily only if we wanted to manually rewrite URLs in the offline-activity definitions.


scytacki:
Thank you submitting this.

I looked through the actual PC content and there are no images inside of interactives that come from https://models-resources.concord.org, so the issue we ran into isn't a problem in the PC content.
So I think we can just leave this PR open for now.

scytacki:
I'm going to close this. We now only re-write the top level interactive urls in the activity content.

PR #215, opened 2021-03-31
Precipitating Change offline version 3 

scytacki:


scytacki:
@dougmartin and @knowuh you might want to look at this one. In case you have to update the content, the steps of these commits will need to be applied. Hopefully we can "take the pen" on the activities sometime soon so the set of manual updates can be reduced.

PR #217, opened 2021-03-31
Added no url rewrite flag to glossary urls in glossary test activity and v3 PC activities

dougmartin:


knowuh (approved):
Looks good to me.

PR #218, opened 2021-04-01
document the need to update portal's redirect_url

scytacki:
Not updating the redirect_url broke the "Show My Work" button that launches the report from the AP.


dougmartin (approved):
Looks good :+1:

PR #222, opened 2021-04-02
Pc v4 content update

scytacki:


dougmartin (approved):
Looks good :+1:

dougmartin:
@scytacki one thing (not for this PR) - we should probably add the "switch to cloudfront urls" step to the update script.

scytacki:
@dougmartin I changed the cloud urls in LARA itself. Using these scripts: 
https://github.com/concord-consortium/lara/pull/696

PR #223, opened 2021-04-03
improve script for building manifest

scytacki:
- only update models-resource urls for iframes since those are the only ones we need proxy
- pretend to be chrome when fetching so sites like google fonts return the same thing we'd get if running in the browser
- add switch to not fetch the activities when running the script, this is helpful when checking the static analysis since the lara content might change
- simplify the test of url re-writing to make it easier to see what gets re-written to what
- stop re-writing the plugin urls because those are not iframes so do not need to be proxied
- only add slashes when the urls are re-written, that was a case in the content of a image that didn't have a extension, and the added slash broke it
- look into proxied iframe html files to pick up the assets referenced by them
- look for css and script references in the iframe html
- fetch css files and look for absolute urls in the css files
- look for urls in the netlogo html files
- look for urls in the html content of the activity


dougmartin (approved):
Looks good 👍

scytacki:
Note this also fixes an issue with slashes being added to the xhtml text. This happened when the text had a url in it, which then triggered the url re-write code which adds a slash to everything if it doesn't end with an extension.

PR #225, opened 2021-04-03
need to include headers in the return value of plugins

scytacki:
this wasn't clear from the Workbox documentation, but looking at the Workbox code
shows the response is passed down through the chain and onto the next lifecycle
hook.

### DIFF
--- a/src/service-worker.ts
+++ b/src/service-worker.ts
@@ -34,7 +34,13 @@ const stripWbRevision: WorkboxPlugin = {
   cacheKeyWillBeUsed: async ({request, mode, params, event, state}) => {
     const url = new URL(request.url);
     url.searchParams.delete("__WB_REVISION__");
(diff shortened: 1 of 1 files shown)


src/service-worker.ts
```diff
@@ -54,7 +60,9 @@ const cleanIndexHtmlParams: WorkboxPlugin = {
     if (url.origin === location.origin &&
         (url.pathname === rootUrl.pathname ||
          url.pathname === rootIndexHtmlUrl.pathname) ) {
-      return rootIndexHtmlUrl.href;
+      // To be safe we pass the headers through, jut like we do when
```
kswenson, line 63:
```suggestion
      // To be safe we pass the headers through, just like we do when
```

dougmartin (approved):
Looks good :+1:

PR #226, opened 2021-04-04
the jquery-ui css file refers to these icon images

scytacki:
Not all of them seem to be used by the activity player but they are small
so it seems good to be complete.

After this change there isn't a need to manually modify the manifest created by the script.


dougmartin (approved):
Looks good :+1:

PR #229, opened 2021-04-05
Added the hash of the service-worker.ts file to the version info

dougmartin:
This will allow us to see version differences during development as the git commit will not change even if the service worker code changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,10 @@
       "json",
       "node"
     ],
-    "coverageDirectory": "coverage-jest"
+    "coverageDirectory": "coverage-jest",
+    "globals": {
+      "__VERSION_INFO__": "fake version for jest"
+    }
   },
   "nyc": {
     "extends": "@istanbuljs/nyc-config-typescript",
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -409,7 +409,6 @@ export class App extends React.PureComponent<IProps, IState> {
   }
 
   render() {
-    const appVersionInfo = (window as any).__appVersionInfo;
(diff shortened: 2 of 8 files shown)


scytacki:
@dougmartin, did you test this with webpack-dev-server?

I did some testing, and it doesn't seem to work well. It seems like when I change the content of service-worker.ts, the hash from the service worker updates, but not the hash in the installer. I think this is because webpack-dev-server doesn't recompile things unless they change. 

Perhaps it is worth keeping this just so a developer can still test the behavior by shutting down webpack-dev-server and restarting it?

scytacki:
I think I found a way to make this work with the webpack dev server:
https://webpack.js.org/loaders/val-loader/

That can return the hash of the service-worker.ts as the value, and it can also add service-worker.ts as a dependency. So then changes to service-worker.ts should trigger rebuilds.

scytacki:
@dougmartin I played around with the val-loader approach and I think I got it to work:
https://github.com/concord-consortium/activity-player/pull/230

It seems to correctly reload the installer page when changes to the service-worker are made, and the hash is changed on the reload.  The version reported by the service-worker has the old hash, as expected, until the page is reloaded again.

dougmartin:
Closing this due to alternate improved PR: https://github.com/concord-consortium/activity-player/pull/230

scytacki:
I figured out how to do this with the DefinePlugin and using this existing PR seems like the best place to start.

scytacki (approved):
I made several changes to this, so my review isn't really isn't necessary. But Doug took a look and thought it looked good.

PR #230, opened 2021-04-06
using val-loader to add service worker hash

scytacki:
This isn't very pretty, but it seems to work.
Might be better to put all of the version info in the val-loader file,
then at least all of the logic is in the same place.

### DIFF
--- a/src/components/install-app.tsx
+++ b/src/components/install-app.tsx
@@ -7,6 +7,7 @@ import { getOfflineManifest, saveOfflineManifestToOfflineActivities,
 import { OfflineManifest } from "../types";
 import { OfflineManifestLoadingModal } from "./offline-manifest-loading-modal";
 import { queryValue } from "../utilities/url-query";
+import serviceWorkerHash from "../webpack-utils/service-worker-hash";
 
 interface IState {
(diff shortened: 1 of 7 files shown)


src/components/install-app.tsx
```diff
@@ -203,7 +204,7 @@ export class InstallApp extends React.PureComponent<IProps, IState> {
   }
 
   render() {
-    const appVersionInfo = (window as any).__appVersionInfo;
+    const appVersionInfo = (window as any).__appVersionInfo + " hash: " + serviceWorkerHash;
```
dougmartin, line 207: Not critical but it would be nice to put something like this in a utils files to DRY it up so it could be used again below in the service worker

```
export const fullVersionInfo = (appVersionInfo: string, serviceWorkerHash?: string) => {
  return `${appVersionInfo}${serviceWorkerHash ? ` hash: ${serviceWorkerHash` : ""}`
}
```

dougmartin (approved):
Looks good :+1:

scytacki:
I found a better way to do this using the DefinePlugin. I'm closing this and will submit a new one.

PR #232, opened 2021-04-06
Only start installing if the service worker version matches

scytacki:
This prevents the situation where the files start installing before the service worker is updated.
Often when the install page is updated, the service worker needs to be updated too, so this
insures the the two versions match before running the install.


knowuh (approved):
Looks good to me.

PR #237, opened 2021-04-08
remove the offline manifest authoring code

scytacki:
Since we are generating the manifests statically, and this authoring code doesn't currently work with the CacheOnly Strategy used by the app, I think it make sense to remove it.

Perhaps we'll want to bring this back if we can't statically handle
new activities, the README references a tag to indicate the last commit when it was working. 

TODO

- [ ] update the README with more info about the static generation


scytacki:
@dougmartin I'm curious what you think about removing this. I know it was your baby. 

I doubt that static analysis is going to work for everything, but relying on manual navigation through the activity was error prone.

My main goal is to try to clean things up so when we come back to this in a few months (or years?) things are less confusing.

dougmartin:
@scytacki I really don't care if it is removed but I'm just wondering how we discover the runtime assets like the iframe app Javascript and css without this mode.

scytacki:
The script currently finds them by downloading the html files and looking for script and link tags. It also looks for some NetLogo specific code:
https://github.com/concord-consortium/activity-player/blob/0d77921abd711d04282485344d29ff8b7192a3b0/src/utilities/activity-utils.ts#L280-L300

scytacki:
It even goes as far as looking for `url(` patterns in the css referenced by the index file: 
https://github.com/concord-consortium/activity-player/blob/0d77921abd711d04282485344d29ff8b7192a3b0/src/utilities/activity-utils.ts#L399-L410

All of that code doesn't cover all of the cases though. And might be hard to maintain because a fix for one set of activities might cause another set of activities to start pulling down extraneous stuff that it doesn't need. But at least in this first case of the PC content it seemed better to figure out those heuristics rather than running through the activity pages in a browser. 

If we run into a problem of fixing one thing breaks another, perhaps we can add per-manifest configuration options to turn on and off various static analysis patterns.

PR #238, opened 2021-04-08
Fix page scrolling problem on iPads

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177691426

[#177691426]

LARA applies `-webkit-overflow-scrolling: touch;` to `body` as a way to prevent the scrolling problem described in the PT story, so we've added that here.


mklewandowski (approved):
👍 LGTM!  I don't have a great way to test, but this seems worth getting on production.

PR #239, opened 2021-04-09
add support for in place manifest updating

scytacki:
This makes it easier to test script changes. 

If `--bump-version` is not used, then the manifest and activities will be updated in place. 

This PR also refactors the code to use `for (... of ...)` to iterate over arrays. This way `await` can be used inside of the `for of ` without using the more complicated reduce approach.


dougmartin (approved):
Looks good :+1:

PR #240, opened 2021-04-09
Add revisions to cache list

scytacki:
Update the cacheList in the offline manifest so it supports simple url strings as well as entry objects. The entry objects support url, revision, and size fields. 

The update-offline-manfest script was updated to output this format. It looks at the ETag header on request for the resource and uses that at the revision. It ignores 'weak' ETags (they start with `W/`).

The script was also updated to check for simple CORS compliance, and to verify that urls that have ETags (and therefore revisions in the manifest), have an `access-control-expose-headers` header which is required to allow the service worker to see the ETag header in a CORS response.

The service worker was updated to check for resource in the cache when it installs them during its `CACHE_ENTRIES_WITH_PROGRESS` handling.  If the cached resource has an ETag header that matches the revision in the manifest then the service worker does not re-fetch the file.  Additionally a new status message was added so the window which requested the install would know which files have been found in the cache. Currently the window side doesn't do anything special with this other than print out a different message to the console. 

Now that the service worker supports a cacheList with entry objects include a url and revision, we can pass the workbox generated manifest directly to the service worker.

At runtime the service worker no longer sends the url through stripWbRevision which was parsing and stringifying the url. This was necessary because this process was done during the install, and urls that had special chars in them would be escaped during the install. So these same chars needed to be escaped during the runtime cache lookup.  The install no longer does this process, so it isn't necessary to do this same escaping at runtime.

I commented out this stripWbRevision code, because it might be worthwhile to add it back in the future if we need to bust remote caches.

To summarize how these changes might effect the different usages of the system:
- authoring: the update-offline-manifest script is a bit smarter and outputs a new format of manifest
- installing:  
  - the install app now sends the workbox generated wbManifest directly to the service worker. It continues to send the cacheList from offline manifest (content manifest), but now it is format is different. 
  - the service worker now looks at the cache first (see above)
- runtime: when the service worker handles, fetch requests it no longer munges the urls before looking for them in the cache. (see above)


dougmartin (approved):
Looks good :+1:

scytacki:
@dougmartin and @knowuh I think this is ready for final review now. The last series of commits are adding tests and comments plus a little refactoring.

src/offline-manifest-api.test.ts
```diff
@@ -33,55 +41,103 @@ describe("offline manifest api", () => {
...
-    //   ]);
-    //   expect(onUrlCached).toHaveBeenCalledTimes(4);
-    //   expect(onCachingFinished).toHaveBeenCalledTimes(1);
-    //   expect(onUrlCacheFailed).not.toHaveBeenCalled();
-    //   done();
-    // });
+  it("handles #cacheUrlsWithProgress", async () => {
+    const restoreConsole = mockConsole();
+    // We are using NodeJs's message channel. It is basically the same as the browser
+    // message channel. Its ports need to be closed otherwise node won't exist
```
knowuh, line 47: For above documentation, I suggest "Node won't exit"

knowuh:
Looks great, thank you.

knowuh (approved):
Looks great, thank you!

PR #243, opened 2021-04-21
Add Google Analytics

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177857290

[#177857290]

This change adds basic Google Analytics tracking to the Activity Player.


mklewandowski (approved):
👍 Seems good to me, but I'm not too familiar with integrating google analytics.

PR #244, opened 2021-04-28
update the readme after running through process

scytacki:
We might have more updates as we finish up the release. These are the changes I remember us running into so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,8 @@ Other branches are deployed to https://activity-player.concord.org/branch/<name>
 To deploy a production release:
 
 1. Copy CHANGES-template.md to CHANGES.md, and add a list of PT stories related to the release into temporary CHANGES.md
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -74,11 +75,14 @@ To deploy a production release:
...
 4. Push current branch and tag to GitHub
+  - `git push origin master`
+  - `git push origin v<new version>`
 5. Create a GitHub Release
     1. Find the new tag at https://github.com/concord-consortium/activity-player/tags open it, and edit it
     2. Copy the title from CHANGES.md
     3. Copy the content from CHANGES.md
-6. QA the built version at `https://activity-player.concord.org/versions/v[new-version-string]``
+    4. Hit "Publish Release" button
+6. QA the built version at `https://activity-player.concord.org/version/v[new-version-string]`
```
tejal-shah, line 85: This URL works with a slash at the end : `https://activity-player.concord.org/version/v[new-version-string]/` - doesn't work without it.

PR #246, opened 2021-05-03
Makes it so if enable learner state is false, there is no exportable …

eireland:
…answer


scytacki (approved):
This works for me.  It is not as nice for authors, users, and researchers, but it is better than having duplicate answers created.

src/utilities/embeddable-utils.ts
```diff
@@ -61,7 +61,7 @@ export const getAnswerWithMetadata = (
     interactiveState: unknown,
     embeddable: {ref_id: string, authored_state?: string | null},
     oldAnswerMeta?: IExportableAnswerMetadata): IExportableAnswerMetadata => {
-
+      console.log(oldAnswerMeta);
```
scytacki, line 64: This should be removed or the `DEBUG_` system should be used to make it conditional.

scytacki (changes requested):
Looks good, just has a couple of console.logs that should be dealt with.

scytacki (approved):
Looks good to me.

@eireland Can you make sure to label the associated story with the next AP release before you merge this? That next release is currently `2.4.0`

PR #247, opened 2021-05-14
Fix Show My Work button does not handle sequences correctly

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178132009

[#178132009]

Use `getResourceUrl` instead of activity param which may not exist.

### DIFF
--- a/src/utilities/report-utils.ts
+++ b/src/utilities/report-utils.ts
@@ -59,6 +59,7 @@ export const getReportUrl = () => {
             + "?"
             + "runKey=" + runKey
             + "&activity=" + resourceUrl
+            + "&resourceUrl=" + resourceUrl
             + "&firebase-app="+reportFirebaseApp
             + "&sourceKey="+sourceKey
             + "&answersSourceKey="+answerSource;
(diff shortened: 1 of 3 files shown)


src/utilities/report-utils.ts
```diff
@@ -47,18 +48,17 @@ const makeSourceKey = (url: string | null) => {
...
   const answerSource = queryValue("report-source") || getCanonicalHostname();
-  const sourceKey = activityUrl ? makeSourceKey(activityUrl) : getCanonicalHostname();
+  const sourceKey = resourceUrl ? makeSourceKey(resourceUrl) : getCanonicalHostname();
 
   if (runKey) {
     return reportLink
             + "?"
             + "runKey=" + runKey
-            + "&activity=" + activityUrl
+            + "&activity=" + resourceUrl
```
scytacki, line 61: It seems better if we changed this parameter to be resourceUrl. But because that requires changes to the portal-report codebase to support. I'd suggest changing this to:
```
// send the resourceUrl in the activity parameter for backwards compatibility
+ "&activity=" + resourceUrl
+ "&resourceUrl=" + resourceUrl
```

Then once the portal-report is updated to handle the resourceUrl we can remove this activity parameter here.

scytacki (changes requested):
suggested change to make the activity parameter more clear

src/utilities/report-utils.ts
emcelroy: @scytacki OK, makes sense 👍

scytacki (approved):
Looks good to me.

PR #248, opened 2021-05-17
Fix sequence drop-down nav doesn't take you to home page of activity.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177958767

[#177958767]

In `handleSelectActivity` set `currentPage` to 0 so the selected activity always begins at its home page.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -417,7 +417,8 @@ export class App extends React.PureComponent<IProps, IState> {
     this.setState((prevState) =>
       ({ activity: prevState.sequence?.activities[activityNum],
          showSequenceIntro: false,
-         activityIndex: activityNum
+         activityIndex: activityNum,
+         currentPage: 0
       })
     );
   }


mklewandowski (approved):
👍

scytacki (changes requested):
Could you try to add a cypress test for this?

It looks like here is an existing test, that could be extended to visit a couple of different activities:
https://github.com/concord-consortium/activity-player/blob/f74895f15728d7aa72d228653496b43066845bb7/cypress/integration/sequence-page.test.ts#L18

emcelroy:
@scytacki I added a test _like_ that in a different PR (See https://github.com/concord-consortium/activity-player/blob/32a118986132583a7259204542c3a1f7297f9938/cypress/integration/sequence-page.test.ts). Should I just reproduce that in this PR, but add a check for the page number when the activity changes?

PR #250, opened 2021-05-18
Add method for loading a specific activity in a sequence

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178131977

[#178131977]

These changes allow you to load a specific activity within a sequence by adding a `sequence-activity` parameter to an Activity Player sequence's URL. They also add/update a `sequence-activity` param to/in the URL whenever a new activity is selected from the sequence navigation menu or sequence home page.

### DIFF
--- a/src/components/activity-header/sequence-nav.tsx
+++ b/src/components/activity-header/sequence-nav.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { CustomSelect } from "../custom-select";
 import ActivityIcon from "../../assets/svg-icons/activity-icon.svg";
+import { setQueryValue } from "../../utilities/url-query";
 
 import "./sequence-nav.scss";
 
@@ -15,14 +16,13 @@ export class SequenceNav extends React.PureComponent <IProps> {
 
   render() {
     const { activities, fullWidth, currentActivity } = this.props;
-    const indexedActivities = this.createIndexedActivitiesMap();
     return (
       <div className={`sequence-nav ${fullWidth ? "full" : ""}`} data-cy="sequence-nav-header">
         <div className="select-label">Activity:</div>
         { activities &&
           <CustomSelect
             items={activities.map((a, i) => `${i + 1}: ${a}`)}
-            value={indexedActivities.get(currentActivity)}
+            value={currentActivity}
             HeaderIcon={ActivityIcon}
             onSelectItem={this.handleSelect}
           />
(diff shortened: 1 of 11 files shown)


scytacki (changes requested):
Because a URL with this `sequence-activity` parameter is going to be published by LARA to the report-service and then used by the portal-report (and possibly the athena reports). It would be better if an more stable id was supported in addition to the index. That way if a author re-orders the activities existing links will still take the user to the correct activity.

You can can follow the pattern documented by the page parameter, in order to support both options. 

Also we are trying to standardize on camelCase for parameter names. So it should be `sequenceActivity`.

The updating of the URL looks good. I asked on slack to see if we should use a routing library instead of directly setting the url: https://concord-consortium.slack.com/archives/C0303QUNU/p1621375955005500

emcelroy:
@scytacki Aside from the routing library question, do the latest changes look good?

Is it OK that when you navigate between activities in a sequence, the activity's index value will always be used to update the `sequenceActivity` param?

(The related change to LARA that's needed for this to work is in [this PR](https://github.com/concord-consortium/lara/pull/712).)

src/components/activity-header/sequence-nav.tsx
```diff
@@ -42,8 +42,9 @@ export class SequenceNav extends React.PureComponent <IProps> {
     const { activities } = this.props;
     const indexedActivities = this.createIndexedActivitiesMap();
     const selectedActivity = indexedActivities.get(item);
-    const activityIndex = activities?.findIndex((activity) => activity === selectedActivity);
-    activityIndex && this.props.onActivityChange(activityIndex);
+    const activityIndex = activities?.findIndex((activity) => activity === selectedActivity) || 0;
+    activityIndex >= 0 && this.props.onActivityChange(activityIndex);
+    activityIndex >= 0 && setQueryValue("sequenceActivity", activityIndex + 1);
```
scytacki, line 47: I might be mis-reading this, but it doesn't looks like it is using the activityId if it exists.   I see there is some logic below that is is using the activityId if it is exists.  Maybe that can be extracted to a function and reused here.

scytacki (changes requested):
I think I found one case where the logic for the sequenceActivity param is wrong, but I might be missing something.

emcelroy:
@scytacki Yeah, it doesn't use the activity ID. Thanks for catching that. I'm pretty sure though that that particular setting of the `sequenceActivity` param is unnecessary since it is set again in app.tsx when the selected activity loads (in `renderActivity`). So I just removed line 47 from sequence-nav.tsx. Does that make sense?

scytacki (approved):
seems good to me

PR #252, opened 2021-05-25
Fix problem with method for loading specific activity in a sequence

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178131977

[#178131977]

This change fixes an issue where the first activity in a sequence wouldn't load when specified using the new `sequenceActivity` URL param.


src/components/app.tsx
```diff
@@ -116,7 +116,7 @@ export class App extends React.PureComponent<IProps, IState> {
                                     : 0;
       const activityIndex = sequence && sequenceActivityNum ? sequenceActivityNum - 1 : undefined;
       const activityPath = queryValue("activity") || kDefaultActivity;
-      const activity: Activity = sequence != null && activityIndex && activityIndex >= 0
+      const activity: Activity = sequence != null && activityIndex !== undefined && activityIndex >= 0
```
kswenson, line 119:
```suggestion
      const activity: Activity = sequence != null && activityIndex !== null && activityIndex >= 0
```
For consistency with the preexisting code and to catch both null/undefined.

scytacki: I'd suspect that line 117 would also need to be changed in this case to return null instead of undefined.

scytacki (changes requested):
Can you update the cypress tests so it verifies this code is working?

src/components/app.tsx
emcelroy: @kswenson @scytacki Thanks. As I understand it, `activityIndex` can either be undefined or a number. If I change undefined to null here, an error is thrown: `Type 'null' is not assignable to type 'number | undefined'.`

Would it make sense to change line 66 to something like `activityIndex?: null | number;`?

kswenson: Sorry, there was simply a typo in my original suggestion:
```suggestion
      const activity: Activity = sequence != null && activityIndex != null && activityIndex >= 0
```
The point is that `!= null` catches null or undefined and so shouldn't require any other changes and is consistent with the `sequence` test earlier in the same line.

emcelroy: OK, thanks. That change seems to work fine. Still, if I change undefined to null in line 117, the `Type 'null' is not assignable to type 'number | undefined'` error still gets thrown. Should I leave that as undefined? Or would it be best to update the types for `activityIndex` to include null?

kswenson: I don't think you should change line 117 at all.

kswenson (approved):
👍 LGTM

scytacki (approved):
looks good to me

PR #253, opened 2021-05-27
Non-wrapping plugins should not take up space in the header section

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177732961

[#177732961]

This change makes it so an `.embeddable` wrapper DIV is not returned if the `qComponent` it would be wrapping is undefined.

### DIFF
--- a/src/components/activity-page/embeddable.tsx
+++ b/src/components/activity-page/embeddable.tsx
@@ -93,6 +93,11 @@ export const Embeddable: React.ForwardRefExoticComponent<IProps> = forwardRef((p
     qComponent = <div>Content type not supported</div>;
   }
 
+  // The following conditional prevents teacher edition containers from being rendered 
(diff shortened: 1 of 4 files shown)


src/components/activity-page/embeddable.tsx
```diff
@@ -93,6 +93,10 @@ export const Embeddable: React.ForwardRefExoticComponent<IProps> = forwardRef((p
     qComponent = <div>Content type not supported</div>;
   }
 
+  if (qComponent === undefined) {
```
scytacki, line 96: Can you add a comment here about how LARA handles this so there isn't a direct dependency on teacherEditionMode, and include a link to the code in there that uses a mutation observer which handles this.

This will be useful if we ever try to remove the teacher edition coupling.

scytacki (changes requested):
This looks good, I just requested a comment to help us in the future.

scytacki (approved):
Looks great.

PR #254, opened 2021-06-06
Fix activities' project-based customizations not working correctly

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178201284

[#178201284]

These changes allow Activity Player to find an activity or sequence's project information by pulling related data from the activity or sequence's export JSON instead of the Activity Player's `src/utilities/project-utils.ts` file. These changes depend on the updates to LARA in [this PR](https://github.com/concord-consortium/lara/pull/719).

Note that project and partner logo files have been removed from the Activity Player source code. They are now kept in the static.concord.org S3 bucket under projects/logos.


src/components/activity-header/logo.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+    const wrapper = shallow(<Logo logo={CCLogo} url={"https://concord.org/"} />);
+    expect(wrapper.find('[data-cy="project-logo"]').length).toBe(1);
+    expect(wrapper.find(".logo-img").length).toBe(1);
+  });
+  it("renders default logo when no logo is specified", () => {
+    const wrapper = shallow(<Logo logo={undefined} url={undefined} />);
+    expect(wrapper.find('[data-cy="project-logo"]').length).toBe(1);
+    expect(wrapper.find(".icon").length).toBe(1);
+  });
+});
```
mklewandowski, line 18: Probably worth adding this newline for consistency.

src/types.ts
```diff
@@ -135,14 +135,25 @@ export interface Page {
...
+export interface Project {
+  about: string | null;
+  footer: string | null;
+  help: string | null;
+  logo_ap: string | null;
+  logo_lara: string | null;
+  project_key: string;
+  title: string | null;
+  url: string | null;
+}
```
mklewandowski, line 147: I think for most variables, interfaces, etc. in the AP, we tend to use camelCase instead of snake_case.  I would have expected to see `logoAp`, `logoLara`, and `projectKey`.  However, maybe this has drifted or has changed since I last worked on the AP?  I know we've now had a lot of engineers on the project so there might be some mixing and matching of style preferences.

mklewandowski (approved):
👍 I had a question about camelCase vs snake_case for naming conventions, but the PR looks good.  I think you can decide what to do there - I would lean towards camelCase, but perhaps some of the style consistency has changed since I was last working on the AP last year.

One question about the overall scope: if I export an authored staging activity as JSON, I see that the `project_id` field is still there and the `project` field is not.  Similarly if I run an existing activity that specifies a project with this branch, we default to having no project specified (presumably because the `project_id` field is present but the newly required `project` field is not).  For example, this activity specifies HAS as the project, but does not show the proper header icon, etc. when running on this new branch:
https://activity-player.concord.org/branch/fix-activities-project-based-customizations/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21036.json&preview

Is there simply follow-up work here to ensure that the new `project` field gets added to the activity/sequence structure?  Or am I missing something?

src/types.ts
scytacki: These types are for the export structure turned by LARA. That structure has historically used snake_case, so it seems best to stick with it so the format is internally consistent.

scytacki:
@mklewandowski, yes this LARA PR needs to be merged before authoring.staging.concord.org will send down the correct project structure: https://github.com/concord-consortium/lara/pull/719

scytacki (approved):
Looks good to me.

scytacki:
I'm just waiting to merge this until LARA staging is updated with the project export changes.

PR #256, opened 2021-07-08
Add runRemoteEndpoint to initInteractiveMessage

kswenson:
also
- update to `@concord-consortium/lara-interactive-api` `1.0.3`
- fill out `classInfoUrl` and `loggedIn` properties of initInteractiveMessage if possible
- eliminate lint warnings
- update dependencies
- npm audit fix
- reduce reported security vulnerabilities from ~5000 to 9

@scytacki If you're aware of other fields of `initInteractive` that could be filled out let me know.


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -255,16 +257,17 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
...
                   interactiveStateUrl: "",
                   collaboratorUrls: null,
-                  classInfoUrl: "",
+                  classInfoUrl: portalData?.portalJWT?.class_info_url ?? "",
                   interactive: {
                     id: 0,
                     name: ""
                   },
                   authInfo: {
                     provider: "",
```
scytacki, line 266: I dug through the LARA source and configuration. And I think this provider takes the form:
`https://learn.concord.org/`
You should be able to construct this by stripping off the path from from one of the URLs passed to the activity player during launch.

scytacki (approved):
This looks great to me. 

I found one additional field you could add to the initInteractive message, but it isn't necessary to do before merging.

PR #257, opened 2021-07-10
Fix sample activity

tejal-shah:


kswenson (approved):
👍 Changes look good and fix the proximate issue of getting the sample documents to work as expected and so this PR is good to go.

I think we should not lose track of the larger issues raised by the underlying PT stories, however:
1. @eireland was following instructions that did not make it clear that the `?` should be included in the authored content leading to the broken link. Those instructions should be clarified.
2. For the missing image issue, I think it is sufficient to add a PT story to the effect that "Image interactive should show high-res image if regular image is not available." In other words, the right behavior in this case would have been to show the high-res image rather than showing a broken image link. The new PT story should link to this PT story for context. We may never get to that PT story, but at least we won't have dropped the ball on it. 😉

tejal-shah:
> 👍 Changes look good and fix the proximate issue of getting the sample documents to work as expected and so this PR is good to go.
> 
> I think we should not lose track of the larger issues raised by the underlying PT stories, however:
> 
> 1. @eireland was following instructions that did not make it clear that the `?` should be included in the authored content leading to the broken link. Those instructions should be clarified. I added a slack message from @eireland to the PT story for additional context.
> 2. For the missing image issue, I think it is sufficient to add a PT story to the effect that "Image interactive should show high-res image if regular image is not available." In other words, the right behavior in this case would have been to show the high-res image rather than showing a broken image link. The new PT story should link to this PT story for context. We may never get to that PT story, but at least we won't have dropped the ball on it. 😉

@kswenson , 
1. For the instructions that were not clear in the authoring for CBio, I've updated the instructions and they now say that everything from the "?" including the "?" should be copied. The instructions were in LARA Staging for the CBio Library Interactive. In Production, there is no CBio LI; so there are no instructions to update there I suppose.
2. For the missing image issue, I've created the PY story https://www.pivotaltracker.com/story/show/178882818 for the more graceful behaviour.

PR #258, opened 2021-07-16
178817207 hidden pages

mklewandowski:
This PR fixes a bug that caused page content to be displayed incorrectly when there are hidden pages.  Hidden pages are now removed from the nav links at the top and bottom of the page (so if an activity has 3 pages and 1 is hidden, 2 nav page links are shows) and goes to the correct page when navigation links are selected (hidden pages cannot be navigated to).


kswenson (approved):
👍 LGTM

PR #259, opened 2021-07-16
Attachments

kswenson:


src/types.ts
```diff
@@ -215,21 +215,30 @@ export interface IAnonymousMetadataPartial {
...
+  id: string;             // randomly generated id (e.g. uuid)
   type: string;
   answer_text?: string;
   answer?: any;
   submitted: boolean | null;
   report_state: string;
+  attachmentsFolder?: IAttachmentsFolder;
+  // TODO: do we want to store the public url? aren't we transitioning to expiring urls?
+  // should this be stored in some non-persistent place so it'll be looked up next time?
+  attachments?: Record<string, string>;  // name/key => url
```
scytacki, line 241: I was thinking the URL here is the unsigned URL to the object in S3. So when a read operation happens, the client would get the credentials (if it doesn't already have them) from the token service. And then sign this unsigned URL.
The url would be something like:
`https://interactive-attachments.s3.aws.com/attachments/[token-service-record-id]/[session-id]/[name]`

The first part of that URL: 
`https://interactive-attachments.s3.aws.com/attachments/[token-service-record-id]` could be taken from the info retrieved from the token-service. 

So you could just store the remaining path: `[session-id]/[name]`.   Or you could just store the `[session-id]` and then compute the path from that.

Not storing the full url does seem useful. This way we could actually change the attachments bucket, or folder and we wouldn't have to worry  updating all of the interactive state metadata.

src/utilities/attachments-manager.ts
```diff
@@ -0,0 +1,91 @@
...
+
+export class AttachmentsManager {
+  private learnerId?: string;
+  private firebaseJwt?: string;
+  private tokenServiceClient: TokenServiceClient;
+  private resources: Record<string, S3Resource> = {};
+
+  constructor(learnerId: string, firebaseJwt?: string) {
+    this.learnerId = learnerId;
+    this.firebaseJwt = firebaseJwt;
```
scytacki, line 27: The JWT for the token service is different than the JWT for the report service or portal.
I think the code calling this is passing the JWT for the report service here. 

What it should probably do instead if pass the portal JWT, so this variable shouldn't be called `portalJwt`. And then if it has a `portalJwt`, then the attachments manager requests a token service jwt. The token service "app name" in the staging and production portals is `token-service`.   

You can kind of trace similar code here: https://github.com/concord-consortium/report-service/blob/efce8eb9d2b8d1484c35c7c74888e3380b472799/researcher-reports/src/components/app.tsx#L45

In that case it is bit more complex because it has to use OAuth to get the portal JWT. In this case the AP should already have a portal JWT since I think it needed that for other stuff.

scytacki (commented):
This is looking good. I left a few comments.

src/types.ts
kswenson: The value I'm passing in is generated at https://github.com/concord-consortium/activity-player/blob/df7fe8dcc1e439c819829b478c549955e915a6d9/src/portal-api.ts#L412, which seemed like it might be appropriate.

src/utilities/attachments-manager.ts
kswenson: The value I'm passing in is generated at https://github.com/concord-consortium/activity-player/blob/df7fe8dcc1e439c819829b478c549955e915a6d9/src/portal-api.ts#L412, which seemed like it might be appropriate.

src/components/app.tsx
```diff
@@ -171,6 +172,10 @@ export class App extends React.PureComponent<IProps, IState> {
         }
       }
 
+      const _portalData = getPortalData() as IPortalData | IAnonymousPortalData;
+      const learnerId = getUniqueLearnerString(_portalData);
+      initializeAttachmentsManager(learnerId, (_portalData as IPortalData).database?.rawFirebaseJWT);
```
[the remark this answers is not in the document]
kswenson: I did some refactoring and now I'm just passing in the PortalData.

src/types.ts
kswenson: We're going around in circles a bit. As you've/we've said before, the sessionId is insufficient to access an attachment written by a collaborator. I'm now storing the folder id and the publicPath.

src/utilities/attachments-manager.ts
kswenson: I've rewritten this to request a "token-service"-specific JWT from the information in the PortalData.

kswenson:
Latest commit should address the `attachments` and JWT issues. I'm going to leave it here for now and pick it up when I get back.

src/utilities/attachments-manager.ts
```diff
@@ -0,0 +1,105 @@
...
+  public getSessionId() {
+    return this.sessionId;
+  }
+
+  public async createFolder(interactiveId: string): Promise<IWritableAttachmentsFolder> {
+    const folderResource = await this.tokenServiceClient.createResource({
+      tool: kTokenServiceToolName,
+      type: "s3Folder",
+      name: `${this.learnerId}-${interactiveId}`,
+      description: "",
```
scytacki, line 57: The token-service requires a description for the resource.

scytacki:
A few notes for when this is picked up again:
- the resource is required to have a `description` (see inline comment)
- the code is referring to a tool with name `'interactive-attachments'`  that tool will have to be configured in the staging token-service before the code can be tested.

scytacki:
I added the new tool definition (resourceSetting) to Firestore:
image

A test activity shows there is an issue. It seems the readWriteToken is not being added to the request headers for the credentials.

src/utilities/attachments-manager.ts
```diff
@@ -0,0 +1,105 @@
...
+    return folderResource;
+  }
+
+  private getCredentials(folder: IAttachmentsFolder): Promise<Credentials> {
+    const readWriteToken = isWritableAttachmentsFolder(folder) ? folder.readWriteToken : undefined;
+    return this.tokenServiceClient.getCredentials(folder.id, readWriteToken);
+  }
+
+  private async getSignedUrlFromPublicPath(folderResource: S3Resource, publicPath: string, operation: string, expires: number) {
+    const credentials = await this.getCredentials(folderResource);
```
scytacki, line 97: This looks like a bug. The passed in `folderResource` is an `S3Resource` from the token-service. But the argument of the getCredentials function is an IAttachmentsFolder. Looking at runtime inspection (at least in the case I inspected) the actual `folderResource` is an `S3Resource`. It doesn't have `folder.readWriteToken`.

kswenson: Good catch -- fixed.

kswenson:
I'm taking this out of draft status now that it has a reasonable set of tests that makes Codecov happy. In testing @scytacki and I are still seeing failures at runtime but those seem as likely to be token-service/s3 issues as issues with this code.

src/utilities/attachments-manager.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  private async getSignedUrlFromPublicPath(folder: IAttachmentsFolder, publicPath: string, operation: string, expires: number) {
+    const folderResource = await this.getFolderResource(folder);
+    const credentials = await this.getCredentials(folder);
+    const { bucket, region } = folderResource;
+    const { accessKeyId, secretAccessKey, sessionToken } = credentials;
+    const s3 = new AWS.S3({ region, accessKeyId, secretAccessKey, sessionToken });
+    // https://zaccharles.medium.com/s3-uploads-proxies-vs-presigned-urls-vs-presigned-posts-9661e2b37932
+    const s3UrlParams = { Bucket: bucket, Key: publicPath, Expires: expires };
```
scytacki, line 86: My understanding is that the `ContentType` param must be set here so it matches the content type sent in the PUT request. When the presigned url fails AWS shows the content that should be signed and it includes this content type.  For cfm example when it sends data it is: `text/plain;charset=UTF-8`.

To make that possible I think it means the interactive api needs to be changed to require sending this content type when it requests the signed url. But before making that change I'd recommend testing this out locally by hard coding it.

Evidence that it is required:
https://stackoverflow.com/a/32464938
https://github.com/aws/aws-sdk-js/blob/307e82673b48577fce4389e4ce03f95064e8fe0d/test/browser/sample/s3upload.html#L47-L55

Entry point in the source (might be slightly different version of library)
https://github.com/aws/aws-sdk-js/blob/7872257f42daafeba691fccd769922100fea05be/lib/services/s3.js#L924
I tried tracing this through by just looking at the source but got lost and could only find the requests that work with the GET operations.

scytacki: I did also see a note that I lost now how when the POST operation is used so the content type is not required.

scytacki (changes requested):
Need to sort out the presigned url issue.

src/utilities/attachments-manager.ts
scytacki: Here is where the string to sign is computed:
https://github.com/aws/aws-sdk-js/blob/e9371bb4ea0671b8034cef08fcdd69bf434d7a03/lib/signers/s3.js#L66-L86

So it always has a content type or a blank line. And it always has an MD5 hash or a blank line. In the error message we get back from AWS we can see the blank line of the MD5 hash followed by the content type, then the expiration time. So that all gives me more confidence that what we are missing is the ContentType in the presigned url params. 

```
PUT

text/plain;charset=UTF-8
1627651148
...
```

I also found this blog post that makes this pretty explicit:
https://advancedweb.hu/differences-between-put-and-post-s3-signed-urls/#:~:text=for%20put%20urls%20the%20signing

Also the conclusion of that blog post is that POST urls are better. Because they let you use wildcard content types, and you can set a size restriction on the upload.  However if I remember right @kswenson you found a blog post suggesting PUT is better.

kswenson: >So that all gives me more confidence that what we are missing is the ContentType in the presigned url params.

Good sleuthing -- I'll give that a try.

>Also the conclusion of that blog post is that POST urls are better. Because they let you use wildcard content types, and you can set a size restriction on the upload. However if I remember right @kswenson you found a blog post suggesting PUT is better.

I'll take a look at that article; I hadn't seen it before. Actually, the article I had found also suggested that POST URLs are better for the same reasons, but then the examples given were for uploading a File using POST or arbitrary content using PUT and it wasn't immediately obvious how to upload arbitrary content using POST so I went with the more salient example.

scytacki:
@kswenson, Given the LARA PR, I'm guessing this worked locally?

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -175,6 +177,42 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
     setActiveLightbox(null);
   };
 
+  const handleGetAttachmentUrl = async (request: IAttachmentUrlRequest): Promise<IAttachmentUrlResponse> => {
+    const { name, operation, type, expires, requestId } = request as any; // TODO: remove cast once next version of library is published
```
scytacki, line 181: @kswenson perhaps you have this change locally and having pushed it, but I'd expect this `type` to be `contentType` based on the update IAttachmentUrlRequest.

kswenson: Yes, local change to be pushed shortly.

package.json
```diff
@@ -143,8 +143,11 @@
     "webpack-dev-server": "^3.11.2"
   },
   "dependencies": {
-    "@concord-consortium/lara-interactive-api": "^1.0.3",
+    "@aws-sdk/client-s3": "^3.24.0",
+    "@aws-sdk/s3-request-presigner": "^3.24.0",
+    "@concord-consortium/lara-interactive-api": "^1.1.1",
```
scytacki, line 148: Should this be bumped to 1.1.2?

kswenson: Yes, once the new version of the library has been published.

kswenson: Done

PR #260, opened 2021-07-27
Background Images AP

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178990327

[#178990327]

These changes allow Activity Player to use the new `background_image` property in activities and sequences to set a background image for the resources' pages. 

Related LARA changes are at https://github.com/concord-consortium/lara/pull/728

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -236,12 +236,8 @@ export class App extends React.PureComponent<IProps, IState> {
                                  ? activityIndex + 1
                                  : undefined;
     sequenceActivity !== undefined && setQueryValue("sequenceActivity", sequenceActivity);
-    const backgroundImage = sequence?.background_image
-                              ? sequence.background_image
-                              : activity.background_image
-                                ? activity.background_image
-                                : undefined;
-    if (backgroundImage !== undefined) {
+    const backgroundImage = sequence?.background_image || activity.background_image;
+    if (backgroundImage) {
       setAppBackgroundImage(backgroundImage);
     }
 
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -8,7 +8,8 @@ import { IntroductionPageContent } from "./activity-introduction/introduction-pa
 import { Footer } from "./activity-introduction/footer";
 import { ActivityLayouts, PageLayouts, numQuestionsOnPreviousPages,
          enableReportButton, setDocumentTitle, getPagePositionFromQueryValue,
-         getSequenceActivityFromQueryValue, getSequenceActivityId } from "../utilities/activity-utils";
+         getSequenceActivityFromQueryValue, getSequenceActivityId,
+         setAppBackgroundImage } from "../utilities/activity-utils";
 import { getActivityDefinition, getSequenceDefinition } from "../lara-api";
 import { ThemeButtons } from "./theme-buttons";
 import { SinglePageContent } from "./single-page/single-page-content";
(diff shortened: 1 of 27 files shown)


src/components/app.tsx
```diff
@@ -235,6 +236,15 @@ export class App extends React.PureComponent<IProps, IState> {
                                  ? activityIndex + 1
                                  : undefined;
     sequenceActivity !== undefined && setQueryValue("sequenceActivity", sequenceActivity);
+    const backgroundImage = sequence?.background_image
+                              ? sequence.background_image
+                              : activity.background_image
+                                ? activity.background_image
+                                : undefined;
+    if (backgroundImage !== undefined) {
```
mklewandowski, line 244: Safer as `if (backgroundImage) {` since then we'll also skip cases where it is null, empty string, etc.

src/components/sequence-introduction/sequence-introduction.tsx
```diff
@@ -14,6 +15,10 @@ interface IProps {
 export const SequenceIntroduction: React.FC<IProps> = (props) => {
   const { sequence, username, onSelectActivity } = props;
   setQueryValue("sequenceActivity", "0");
+  const backgroundImage = sequence?.background_image ? sequence.background_image : undefined;
+  if (backgroundImage !== undefined) {
```
mklewandowski, line 19: Safer as `if (backgroundImage) {` since then we'll also skip cases where it is null, empty string, etc.

src/components/sequence-introduction/sequence-introduction.tsx
```diff
@@ -14,6 +15,10 @@ interface IProps {
 export const SequenceIntroduction: React.FC<IProps> = (props) => {
   const { sequence, username, onSelectActivity } = props;
   setQueryValue("sequenceActivity", "0");
+  const backgroundImage = sequence?.background_image ? sequence.background_image : undefined;
```
mklewandowski, line 18: Should be safe to just write `const backgroundImage = sequence?.background_image;` with the change below to the next line.

src/components/app.tsx
```diff
@@ -235,6 +236,15 @@ export class App extends React.PureComponent<IProps, IState> {
                                  ? activityIndex + 1
                                  : undefined;
     sequenceActivity !== undefined && setQueryValue("sequenceActivity", sequenceActivity);
+    const backgroundImage = sequence?.background_image
+                              ? sequence.background_image
+                              : activity.background_image
+                                ? activity.background_image
+                                : undefined;
```
mklewandowski, line 243: You can collapse all of this into a single line if you also make the change on the line below.  The above only needs to be `const backgroundImage = sequence?.background_image || activity.background_image;`

mklewandowski (approved):
👍 LGTM!  I added some code suggestions to make the code a little shorter and to make the check for background images a little looser to catch all falsey values (undefined, null, "", etc.).

src/components/app.tsx
emcelroy: Thanks for the great suggestions. I've made all these changes.

scytacki (approved):
looks good to me

src/utilities/activity-utils.ts
```diff
@@ -133,8 +133,10 @@ export const getLinkedPluginEmbeddable = (page: Page, id: string) => {
 
 export const setAppBackgroundImage = (backgroundImageUrl?: string) => {
   const gradient = "linear-gradient(to bottom, #fff, rgba(255,255,255,0), rgba(255,255,255,0))";
-  const el = document.querySelector(".app") as HTMLElement;
-  el?.style.setProperty("background-image", backgroundImageUrl || gradient);
+  const el = document.querySelector("#app") as HTMLElement;
```
emcelroy, line 136: @mklewandowski @scytacki I made another small change to line 136 that I think should be harmless, but please take a quick look and let me know if you see any issues.

When the background image was set on the `.app` element, the image wouldn't always reach the bottom of the browser window.  So I changed it to apply the background to the `#app` element.

mklewandowski (approved):
👍

PR #261, opened 2021-08-09
add plugin support when loading from sequence intro page

mklewandowski:
This PR adds support for plugins (including things like teacher edition content) when the initial loaded content is a sequence without a current activity specified - in this case we load the sequence intro page.  Code changes now search all of the activities within the sequence to determine which plugins, if any, need to be initialized.  Previously no activity content was searched and no plugins were initialized.
Changes includes:
- new sample data
- pass array of activities to `loadLearnerPluginState` and `loadPluginScripts`.  This array is made up of all of the activities in the sequence or the current activity.
- search array of activities to determine if we need to load plugin data

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -186,8 +186,9 @@ export class App extends React.PureComponent<IProps, IState> {
       this.setState(newState as IState);
 
       this.LARA = initializeLara();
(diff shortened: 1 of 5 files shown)


emcelroy (approved):
Looks good to me! 

I did initially think the nested loops in`findUsedPlugins` were maybe getting out of hand, but I can't think of how you'd do that differently.

PR #262, opened 2021-08-25
Update Attachments Manager to create resources with context access rule

pjanik:
[#179181324]

Update Attachments Manager to create resources with context access rule, so teachers have access to students' attachments too.

This PR depends on @concord-consortium/token-service client v2.1.0. The only change in this client is to allow the client app to specify multiple access rules (https://github.com/concord-consortium/token-service/commit/08f3272c7772e26aa126f8fc99b95be8c7e1dafd). I was necessary here. I just published the new client version a moment ago. Note that the token service will automatically add "user" rule if there's only "context" specified, but I think it's better to be explicit and treat this feature as a backup option only.

Also, this means that all the attachments created up to now won't work in the teacher report. We'd need to go over all of them and add "context" access rule in Firestore manually. But since it's only on staging, I guess that's fine. Just a warning in case someone would like to test his older activity in the report (when the implementation there is ready).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6520,9 +6520,9 @@
       }
     },
     "@concord-consortium/token-service": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/@concord-consortium/token-service/-/token-service-2.0.0.tgz",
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍  LGTM!

PR #263, opened 2021-08-26
WIP: Refactor Attachments API to make it more reusable

pjanik:
[#179181324]

@kswenson, @scytacki, I do NOT plan to get that PR merged. This is my proposal of how the new Attachment API host helpers will be organized. It's mostly based on @kswenson work. I'm opening this PR here so all the diff and changes are easier to review + we can see how it's integrated in the host environment in practice. Eventually, most of this code will be in a new NPM package. Also, tests are not updated here, so TravisCI build will fail.

- attachment-manager.ts -> this will be fully moved to a separate repo and NPM package. It includes the updated manager, its init helper, all the necessary interfaces, and a helper function that is used directly by the managed-interactive to handle requests. It'll be divided into more files in the new NPM package, but I just kept it all together here for simplicity.
- app.ts -> the diff shows how the host initializes the attachments manager.
- managed-interactive.ts -> this shows how the host uses the helper in practice to handle requests.
- get-attachments-manager-options.ts -> the host-specific code that obtains all the necessary data like JWT, learner ID, etc.

I've used the same code in portal-report and things seem to be working fine.


pjanik:
I'm closing this PR, as I only wanted to make review easier and receive feedback. I'll open a new PR that creates interactive-api-host NPM package based on this code and AP PR that uses this package to handle getAttachmentUrl requests.

PR #265, opened 2021-09-02
Update logger to include username, correct activity value for sequences, url, and add env selection

pjanik:
I've added a new PT story to track these changes:
https://www.pivotaltracker.com/story/show/179455190
I didn't add a version tag, as I'm not sure which version of AP is coming next.

This PR updates Logger. Changes:
- set username using domain and domain_uid URL params when they're available
- set URL in log events -> This lets researchers use `Only return Teacher Edition events` filter in log-puller
- set correct activity param value for sequences -> sequence JSON path instead of null
- try to guess the expected Log Manager environment based on the report-service env
- small refactor of log puller options -> hash instead of super long argument list

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -49,6 +49,8 @@ const kTimeout = 5 * 60 * 1000; // 5 minutes
 
 const kLearnPortalUrl = "https://learn.concord.org";
 
+const kAnonymousUserName = "Anonymous";
+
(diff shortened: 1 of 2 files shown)


scytacki (approved):
Looks good to me. 

It doesn't have any new tests to verify the username change, but the amount of coverage of these changed files increased so at least most of the new code is being executed.

PR #266, opened 2021-09-09
Update jest and randomly failing test

pjanik:
My recent update broke an unrelated test:
https://github.com/concord-consortium/activity-player/runs/3556741148
Also, it only breaks in the CI env, I couldn't replicate it locally.
So, I tried to update jest library and use a slightly different approach in the problematic test. It seems to help and should be better anyway.


scytacki (approved):
Yay! I think this has been causing problems for a while. hopefully this works better.

PR #267, opened 2021-09-19
Plugin API updates to support non-hardcoded plugins

pjanik:
This PR adds missing Plugin APIs, removes hardcoded references to plugin labels, and tries to refactor or fix some pieces. My attempt to support Sharing Plugin and other non-hardcoded plugins. If there's ever more time to refactor things, I'd try to remove all the hardcoded plugin references. I don't understand why it ended up being implemented that way. But maybe I'm just missing some issues, and this PR that partially removes that, causes some issues I'm not aware of.

I was testing it locally using an activity with Teacher Edition tips, Glossary and Sharing Plugin. It seems to work for me. Note that:
- I've removed references to all the plugin labels like "glossary", "teacherEditionTips", etc. That way multiple plugins types can be loaded. But I don't know why it was implemented before and if I'm not breaking something.
- I've removed checking if we're in the TeacherEdition mode in the AP. Teacher Edition plugin does it itself, so I don't know why AP was doing it. Again, am I breaking something?

There's an issue that I've commented on, but didn't try to fix, as I'm worried it'd break student data:
https://www.pivotaltracker.com/story/show/179634819
If there's no production data yet, I'd recommend fixing that, as it's gonna be more difficult to fix it later.

Also, I think that AP won't support two plugins wrapping the same interactive (for example TE and Sharing):
https://www.pivotaltracker.com/story/show/179634899

### DIFF
--- a/src/components/activity-page/embeddable.tsx
+++ b/src/components/activity-page/embeddable.tsx
@@ -53,14 +53,13 @@ export const Embeddable: React.ForwardRefExoticComponent<IProps> = forwardRef((p
       embeddableContainer: embeddableWrapperDivTarget.current || undefined,
       wrappedEmbeddable: embeddable,
       wrappedEmbeddableContainer: embeddableDivTarget.current || undefined,
-      sendCustomMessage,
-      approvedScriptLabel: "teacherEditionTips"
+      sendCustomMessage
     };
     const validPluginContext = validateEmbeddablePluginContextForWrappedEmbeddable(pluginContext);
-    if (validPluginContext && teacherEditionMode && pluginsLoaded) {
+    if (validPluginContext && pluginsLoaded) {
       initializePlugin(validPluginContext);
     }
-  }, [LARA, linkedPluginEmbeddable, embeddable, teacherEditionMode, pluginsLoaded]);
+  }, [LARA, linkedPluginEmbeddable, embeddable, pluginsLoaded]);
 
   useImperativeHandle(ref, () => ({
     requestInteractiveState: () => {
@@ -93,10 +92,10 @@ export const Embeddable: React.ForwardRefExoticComponent<IProps> = forwardRef((p
     qComponent = <div>Content type not supported</div>;
   }
 
-  // The following conditional prevents teacher edition containers from being rendered 
-  // when not in teacher edition mode. LARA handles this differently by using a mutation observer 
-  // (see https://github.com/concord-consortium/lara/blob/master/app/views/plugins/_show.haml). 
-  // It would be better to do that here as well if we update Activity Player to not use direct 
+  // The following conditional prevents teacher edition containers from being rendered
+  // when not in teacher edition mode. LARA handles this differently by using a mutation observer
+  // (see https://github.com/concord-consortium/lara/blob/master/app/views/plugins/_show.haml).
+  // It would be better to do that here as well if we update Activity Player to not use direct
   // dependencies on teacherEditionMode.
   if (qComponent === undefined) {
     return null;
--- a/src/lara-plugin/plugins/embeddable-runtime-context.spec.ts
+++ b/src/lara-plugin/plugins/embeddable-runtime-context.spec.ts
@@ -1,11 +1,10 @@
 import { generateEmbeddableRuntimeContext } from "./embeddable-runtime-context";
 import { IInteractiveState } from "../plugin-api";
 import { emitInteractiveAvailable, emitInteractiveSupportedFeatures } from "../events";
-// ACTIVITY_PLAYER_CODE:
 import fetch from "jest-fetch-mock";
-// LARA_CODE import * as fetch from "jest-fetch-mock";
-
+import { EmbeddableBase } from "../../types";
 import { IEmbeddableContextOptions } from "./plugin-context";
+
 (window as any).fetch = fetch;
 
 describe("Embeddable runtime context helper", () => {
@@ -19,7 +18,7 @@ describe("Embeddable runtime context helper", () => {
       name: "Test Interactive",
       type: "MwInteractive",
       ref_id: "86-MwInteractive"
-    },
+    } as EmbeddableBase,
     interactiveStateUrl: "http://interactive.state.url",
     interactiveAvailable: true
   };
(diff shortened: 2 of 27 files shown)


scytacki:
I haven't had time to review this yet, but I wanted to confirm you are doing the right thing. The AP should not have hard coded references to specific plugins or modes.

They were added initially just to get things working quickly.

src/components/activity-page/embeddable.tsx
```diff
@@ -93,10 +92,10 @@ export const Embeddable: React.ForwardRefExoticComponent<IProps> = forwardRef((p
...
   }
 
-  // The following conditional prevents teacher edition containers from being rendered 
-  // when not in teacher edition mode. LARA handles this differently by using a mutation observer 
-  // (see https://github.com/concord-consortium/lara/blob/master/app/views/plugins/_show.haml). 
-  // It would be better to do that here as well if we update Activity Player to not use direct 
+  // The following conditional prevents teacher edition containers from being rendered
+  // when not in teacher edition mode. LARA handles this differently by using a mutation observer
+  // (see https://github.com/concord-consortium/lara/blob/master/app/views/plugins/_show.haml).
+  // It would be better to do that here as well if we update Activity Player to not use direct
```
kswenson, line 98:
```suggestion
  // It would be better to do that here as well now that Activity Player no longer has direct
```

src/lara-plugin/plugins/embeddable-runtime-context.spec.ts
```diff
@@ -1,11 +1,10 @@
 import { generateEmbeddableRuntimeContext } from "./embeddable-runtime-context";
 import { IInteractiveState } from "../plugin-api";
 import { emitInteractiveAvailable, emitInteractiveSupportedFeatures } from "../events";
-// ACTIVITY_PLAYER_CODE:
```
kswenson, line 4: @mklewandowski can speak more to this, but as I recall the files in the `lara-plugin` folder were copied directly from LARA and the goal was to make as few changes as possible to prevent them from diverging from their LARA counterparts thus making future merges/synchronizations easier. These `ACTIVITY_PLAYER_CODE` and `LARA_CODE` comments indicated where code changes were required to support the activity player. All of which is just to say that changes made to these files should be made with some consideration to how/whether they will be synchronized with the corresponding code in LARA moving forward. Ultimately, a more robust form of code sharing (e.g. npm module) would clearly be preferable.

kswenson (commented):
I made one suggestion about updating a comment and also added @mklewandowski as a reviewer since I think he's the best person to comment on the question of why things were done the way they were (which in many cases may be because it wasn't clear what the alternatives were).

src/lara-plugin/plugins/embeddable-runtime-context.spec.ts
mklewandowski: This is exactly correct. The LARA plugin code was brought in in its entirety and we tried to mark with comments wherever code needed to diverge in hopes of making future code syncing easier.

mklewandowski:
I don't have internet so I cannot pull the changes or easily look at them on my laptop. I can only access via my phone. That said:
- plugin functionality was added incrementally in the AP. We first added the Teacher Edition (TE) plugin.  We had to conditionally check for the plugin type since that was the ONLY plugin that the AP supported initially. Otherwise we were just loading plugin code for no reason. Also when we first built this, links to the plugin code were NOT in the activity JSON. We had to search the JSON, infer what plugins we're needed, and then load from hard coded URLs that were baked into the AP. This has since changed but this definitely influenced the decision to check the plugin type.  We had to since we needed to match a type to a hard coded plugin js url in the AP. 
- the glossary plugin was then added. We followed the same pattern. This was done with some degree of caution since those of us working on this (mainly me and the others) did not have experience with the plugins, their design goals, or their architecture. 
- we copied all of the LARA plugin code into the AP as needed and tried to make minimal changes and mark those changes with comments per Scott's request. This was in case we needed to merge or sync the two code bases. We could use the comments to more easily identify changes.  
- ideally, yes, we shouldn't need to look at specific plugin types. We should just support them all. However aren't there places where we are making specific layout decisions based on the plugin type? How are you handling this?  Maybe I'm forgetting how it works and this can all be done without looking at the script label?  But surely there are places where we need to make some decision about what sort of plugin content we are dealing with, right? The same is true for the URL param.  How can you, for example, show the teacher edition banner if you don't check the URL param? Maybe I'm simply not familiar with the original integration and design goals. Again, most of us who worked on the initial AP content did not have LARA experience so maybe we overlooked some key design ideas.

src/lara-plugin/plugins/embeddable-runtime-context.spec.ts
mklewandowski: I would vote to not remove these comments. This was requested by @scytacki (unless he is ok removing them).

src/utilities/plugin-utils.ts
```diff
@@ -109,38 +104,48 @@ export const validateEmbeddablePluginContextForWrappedEmbeddable =
...
+  // instance per activity. Fixing this would destroy students data, so I'm not doing this. But it should be handled
+  // if we ever add more plugins that save their state.
+  return await Promise.all(approvedScripts.map(async (approvedScriptInfo) => await getLearnerPluginState(approvedScriptInfo.id)));
 };
 
 export const initializePlugin = (context: IEmbeddablePluginContext) => {
   const { LARA, embeddable, embeddableContainer,
-          wrappedEmbeddable, wrappedEmbeddableContainer, sendCustomMessage, approvedScriptLabel } = context;
-  const usedPlugin = usedPlugins.find(p => p.plugin.approved_script_label === approvedScriptLabel);
-  if (!usedPlugin) return;
```
mklewandowski, line 121: As mentioned, this code existed as the AP initially only handled the teacher edition plugin. So we were intentionally filtering out the other plugins that we did not yet support. If we now have full plugin support, then it seems fine to remove.

pjanik:
Thanks for the explanations. I'm stressing these changes and design changes not to break something accidentally. So, it seems that the main reason was to avoid loading unsupported plugins. That's good news. Now, AP would load them, plugin might fail if it's not supported. But this should not happen on production with real activities. And I think the only untested and possibly unsupported plugin at this point is the model feedback.

> Otherwise we were just loading plugin code for no reason. 

This might happen after merging this PR. For example, TE plugin script will be loaded even in the student runtime. But TE plugin checks itself if it's in the teacher edition mode and doesn't do anything. Other plugins that AP doesn't support yet might just break, but I think none of such activities will be used on production (and it's only model feedback in practice).

> But surely there are places where we need to make some decision about what sort of plugin content we are dealing with, right? 

I think LARA doesn't do it. Maybe the only exception is the TE banner, but I don't remember how it was implemented. I've tried to remove here some references to teacherEditionMode, but perhaps I've left some that were not related to my changes.

src/lara-plugin/plugins/embeddable-runtime-context.spec.ts
pjanik: I'm not sure if that's sustainable long-term. Plugin API is API types + implementation which is host-specific. For example, AP and LARA obtain class info in a very different ways. Or they download firestoreJWT using different methods. It applies to most of the functions provided by the Plugin API. But I don't remember exactly all the pieces of plugin-api. Maybe some things can be reused directly.

The big shared part is the types and interfaces. They're already in `@concord-consortium/plugin-api`, so one day we could try to remove AP copies of the types and use ones from the package.

scytacki: I'm fine with removing these comments now. I think the time has passed when it made sense to try to share the code between the two projects. We are getting much closer to shutting down the LARA runtime. And on top of this when I try to look at the comments now in order to figure out what the differences are it isn't helping me much. 

Sorry for the extra work these comments caused only to be thrown away.

scytacki:
This looks good to me. The one thing I want to check before I approve is:
- [ ] check if window shade TE items cause extra space when not in TE mode. The comment that Kirk revised above refers to this as a potential issue. I looked quickly with Piotr, but forgot we needed to look at window shades instead of question wrappers.

mklewandowski:
> This looks good to me. The one thing I want to check before I approve is:
> 
> * [ ]  check if window shade TE items cause extra space when not in TE mode. The comment that Kirk revised above refers to this as a potential issue. I looked quickly with Piotr, but forgot we needed to look at window shades instead of question wrappers.

I just compared to `master` when running in non-TE mode using an activity that includes TE window shades.  I do NOT see window shade TE items adding any additional space or layout change.

src/components/activity-page/embeddable.tsx
scytacki: I checked this and it turns out not all all direct dependency on teacherEditionMode has been removed. For this specific case line 89 is the key:
`qComponent = teacherEditionMode ? <EmbeddablePlugin embeddable={embeddable} pluginsLoaded={pluginsLoaded} /> : undefined;`

So this comment before Kirks corrections is still correct.

scytacki:
Yes I checked also. And I looked at the code and that makes sense. The code is still only rendering `Embeddable::EmbeddablePlugin`s if they are windowShades and the AP is in teacher edition mode.

scytacki (changes requested):
This looks good. 
My only requested change is to bring up the test coverage.

I'll see if I can take care of that myself.

scytacki:
I'm going to go ahead and merge this as is, because the increased test coverage is going to include several other changes and it will be easier to review that independently.

PR #271, opened 2021-10-04
Added fixed width layout [#179122973]

dougmartin:


scytacki (changes requested):
This looks good. I've got two issues though:

Based on the way LARA does things I would expect the sequence full_width_layout to override any activity specific setting.  Maybe I'm missing it, but it looks like the sequence level setting is being ignored.

Also this doesn't seem to be tested. The new code is covered because it is hit by the sample activities. But there isn't a test that verifies the width is actually changed when a sample activity is loaded. The most useful test would be one that makes sure an interactive has enough width with the 40-60 layout and with the full width layout. It was this need which is why we added this new setting.

dougmartin:
@scytacki I've made a series of commits to address the testing.  The best way to view them is via this compare link:

https://github.com/concord-consortium/activity-player/compare/b1a789f...db30ec2

(I've also fixed all the outstanding lint issues)

PR #272, opened 2021-10-06
Responsive layout

eireland:
Adds sample data for various multiple layout in a page, and multiple column reponsive layout
Fixes styling for responsive layout., two column responsive layout, and multiple different layouts in a page.

### DIFF
--- a/src/utilities/plugin-utils.ts
+++ b/src/utilities/plugin-utils.ts
@@ -41,6 +41,15 @@ export const findUsedApprovedScripts = (activities: Activity[]) => {
           if (!activity.pages[page].sections[section].is_hidden) {
             for (let embeddableNum = 0; embeddableNum < activity.pages[page].sections[section].embeddables.length; embeddableNum++) {
               const embeddable = activity.pages[page].sections[section].embeddables[embeddableNum];
+              // NOTE: TODO this needs to be fixed.
+              //This change might cause there to be extra space when running in non teacher edition mode and viewing a page with
+              // with teacher edition items on it.
+              //The issue is that the teacher edition item wrapper will be added to the page, and then the teacher edition plugin inside
+              //of it will decide not render anything because the app isn't in teacher edition mode.
+              //That behavior by itself is correct because we are trying to not have these "modes" be explicitly part of the
+              //activity player (or lara runtime). But the problem is that the item wrapper usually has some padding or margins.
+              //So even though it has no real content, the padding will still show up in the runtime.
+              //This change was from master when new-sections was rebased on it.
               if (embeddable.type === "Embeddable::EmbeddablePlugin" && embeddable.plugin) {
                 addUsedPlugin(embeddable.plugin);
               }


knowuh (approved):
This looks good to me Evangeline!

Thank you.

I wouldn't mind another set of eyes on it, just because its so big, and I don't completely understand the merge history for this branch.

src/utilities/plugin-utils.ts
```diff
@@ -42,15 +41,14 @@ export const findUsedApprovedScripts = (activities: Activity[]) => {
           if (!activity.pages[page].sections[section].is_hidden) {
             for (let embeddableNum = 0; embeddableNum < activity.pages[page].sections[section].embeddables.length; embeddableNum++) {
               const embeddable = activity.pages[page].sections[section].embeddables[embeddableNum];
-              if (embeddable.type === "Embeddable::EmbeddablePlugin" && embeddable.plugin?.approved_script_label === "teacherEditionTips" && teacherEditionMode) {
+              if (embeddable.type === "Embeddable::EmbeddablePlugin" && embeddable.plugin) {
```
scytacki, line 44: This change might cause there to be extra space when running in non teacher edition mode and viewing a page with with teacher edition items on it.

The issue is that the teacher edition item wrapper (I don't know what this component is actually called) will be added to the page, and then the teacher edition plugin inside of it will decide not render anything because the app isn't in teacher edition mode. That behavior by itself is correct because we are trying to not have these "modes" be explicitly part of the activity player (or lara runtime).  But the problem is that the item wrapper usually has some padding or margins. So even though it has no real content, the padding will still show up in the runtime.

The LARA runtime handles this by observing the content inside of the wrapper and as long as it is empty then no padding or margins are added. The AP doesn't have this observing code yet.

So I'd suggest leaving this code as is for now, perhaps with a comment about what I described above, until something like this observing code is added to the activity player.

src/utilities/plugin-utils.ts
```diff
@@ -110,10 +108,14 @@ export const validateEmbeddablePluginContextForWrappedEmbeddable =
...
-//   return await Promise.all(plugins.map(async (plugin) => await getLearnerPluginState(plugin.id)));
-// };
+export const loadLearnerPluginState = async (activities: Activity[], teacherEditionMode: boolean) => {
+  const approvedScripts = findUsedApprovedScripts(activities);
+  // PJ 09/19/2021: This doesn't seem to make sense. Currently, the state is saved and restored per approved script.
+  // It should be saved and restored per plugin instance. It works with Glossary only because there's one glossary
+  // instance per activity. Fixing this would destroy students data, so I'm not doing this. But it should be handled
+  // if we ever add more plugins that save their state.
+  return await Promise.all(approvedScripts.map(async (approvedScriptInfo) => await getLearnerPluginState(approvedScriptInfo.id)));
+};
```
scytacki, line 118: I'm surprised this is in this PR. It was added to master in a different PR.

scytacki (changes requested):
I think I found an error that would happen with teacher edition items.

However perhaps the change I'm commenting on is actually a change from master that is showing up in this PR like the loadLearnerPluginState. In that case it might be a safe change because I remember checking the changes to master to make sure they didn't cause the "extra space" problem.

src/utilities/plugin-utils.ts
eireland: It got added to new-sections branch in the last rebase to master.

src/utilities/plugin-utils.ts
```diff
@@ -42,15 +41,23 @@ export const findUsedApprovedScripts = (activities: Activity[]) => {
...
-              if (embeddable.type === "Embeddable::EmbeddablePlugin" && embeddable.plugin?.approved_script_label === "teacherEditionTips" && teacherEditionMode) {
+              // NOTE: TODO this needs to be fixed.
+              //This change might cause there to be extra space when running in non teacher edition mode and viewing a page with
+              // with teacher edition items on it.
+              //The issue is that the teacher edition item wrapper will be added to the page, and then the teacher edition plugin inside
+              //of it will decide not render anything because the app isn't in teacher edition mode.
+              //That behavior by itself is correct because we are trying to not have these "modes" be explicitly part of the
+              //activity player (or lara runtime). But the problem is that the item wrapper usually has some padding or margins.
+              //So even though it has no real content, the padding will still show up in the runtime.
+              //This change was from master when new-sections was rebased on it.
```
scytacki, line 52: Sorry now that I understand this better, this comment should not be added. The extra space does not happen in master which also doesn't have this explicit teacher edition check. 

It works because there is another check in this component with a similar comment:
https://github.com/concord-consortium/activity-player/blob/67f77441b5f54773a4984e9fceb6cc52fa9e3e55/src/components/activity-page/embeddable.tsx#L95-L100

scytacki (changes requested):
This looks good. 

However, now that I understand this, the comment should be removed because it will be confusing. There is already a comment in the code about this here: 
https://github.com/concord-consortium/activity-player/blob/67f77441b5f54773a4984e9fceb6cc52fa9e3e55/src/components/activity-page/embeddable.tsx#L95-L100

PR #275, opened 2021-10-18
Pin interactive

eireland:
- Adds pinning of primary column when section is scrolled and secondary column is longer than primary column. use-resize-observer library was added to get the final height of the primary column on render.
- Refactor section and its styling into it's own component
- Puts embeddable styling back in the embeddable.scss
- Fixes overlapping columns when browser window is resized.


src/components/single-page/single-page-content.tsx
```diff
@@ -23,6 +23,10 @@ export const SinglePageContent: React.FC<IProps> = (props) => {
   const renderPageContent = (page: Page, index: number) => {
     // const visibleEmbeddables: VisibleEmbeddables = getVisibleEmbeddablesOnPage(page);
     // const embeddables = [...visibleEmbeddables.headerBlock, ...visibleEmbeddables.interactiveBox, ...visibleEmbeddables.infoAssessment];
+    const onSizeChange = () => {
```
eireland, line 26: Needed to add this function for code to compile. Currently a place holder until single page content is added back in.

src/components/activity-page/section.scss
```diff
@@ -0,0 +1,189 @@
...
+      justify-content: flex-start;
+    }
+
+    div {
+      position: relative;
+      top: 2px;
+      left: 0px;
+      width: 30px;
+      white-space: nowrap;
+    }
```
emcelroy, line 176: Can we change this to

```
div {
  align-self: center;
  position: relative;
  white-space: nowrap;
  width: 30px;
  }
```

That will vertically center the text better. (I think the use of `top` and `left` are a relic of something I didn't do quite right a long time ago.)

eireland:
Removed duplicate commits when branched off responsive-layout

emcelroy (approved):
Looks good to me! Thanks.

PR #276, opened 2021-10-19
Single page layout

eireland:
Adds back the ability to show single page layout with new sections


src/components/single-page/single-page-content.tsx
```diff
@@ -17,39 +15,22 @@ interface IProps {
 
 export const SinglePageContent: React.FC<IProps> = (props) => {
   const { activity, teacherEditionMode, pluginsLoaded } = props;
-  let questionNumber = 0;
-  let embeddableNumber = 0;
-
+  const questionNumber = 0;
```
knowuh, line 18: `questionNumber` is always 0, and then we add it to our embeddableQuestionNumberStart.

It seems like that was something carried over that we can get rid of that completely.

src/components/single-page/single-page-content.tsx
```diff
@@ -17,39 +15,22 @@ interface IProps {
 
 export const SinglePageContent: React.FC<IProps> = (props) => {
   const { activity, teacherEditionMode, pluginsLoaded } = props;
```
knowuh, line 17: I am also seeing TSLint information in github saying that we don't need teacherEditionMode or pluginsLoaded.

src/components/single-page/single-page-content.tsx
```diff
@@ -1,13 +1,11 @@
 import React from "react";
-// import { ActivityLayouts, PageLayouts, isQuestion, VisibleEmbeddables, getVisibleEmbeddablesOnPage,
-//          EmbeddableSections, getLinkedPluginEmbeddable } from "../../utilities/activity-utils";
-import { ActivityLayouts, isQuestion, getLinkedPluginEmbeddable } from "../../utilities/activity-utils";
-import { Embeddable } from "../activity-page/embeddable";
+import { ActivityLayouts, isQuestion, getLinkedPluginEmbeddable, numQuestionsOnPreviousSections } from "../../utilities/activity-utils";
```
knowuh, line 2: TSLint saying we don't need to import `ActivityLayouts`,`isQuestion`, `getLinkedPluginEmbeddable`, or `ActivityLayouts` here.

knowuh (changes requested):
This looks good.

It would be good to get rid of the `questionNumber` constant that is unused.  There are also few places where TSLint is complaining about unused imports and variables.  It would be nice to get rid of those warnings too.

src/components/app.tsx
```diff
@@ -197,7 +193,7 @@ export class App extends React.PureComponent<IProps, IState> {
       this.setState(newState as IState);
 
       this.LARA = initializeLara();
-      const activities: Activity[] = sequence ? sequence.activities : [activity];
+      // const activities: Activity[] = sequence ? sequence.activities : [activity];
```
knowuh, line 196: These are commented out for now until we get plugins working again, correct?

PR #277, opened 2021-10-19
Added collaborators_data_url to answer data if available [#173062938]

dougmartin:


src/types.ts
```diff
@@ -214,6 +214,8 @@ export interface ILTIPartial {
   tool_id: string;
    // This is not an LTI property but it is required in our authenticated answers
   remote_endpoint: string;
+  // This is not an LTI property but it is required to track collorations
```
kswenson, line 217:
```suggestion
  // This is not an LTI property but it is required to track collaborations
```

dougmartin:
@scytacki I added the Cypress test for collaboration launches:  https://github.com/concord-consortium/activity-player/pull/277/commits/2f7c3583c07c8fa9b90c1a963056b7f956e98808

The controller endpoint is protected by a CSRF token - getting that proved to be a bit of a bear but it is done.

cypress/integration/portal-launch.test.ts
```diff
@@ -73,6 +73,53 @@ context("Launch AP From the Portal", () => {
     cy.get("[data-cy=activity-summary]").should("contain", "Single Page Test Activity");
   });
 
+  it("can be launched collaboratively", () => {
```
scytacki, line 76: The before each is also launching an activity. It guess it would be better if the before each just logged in the user.
And then the two non-collaborative lauches are put in their own context block with a before each that launches the activity. 

And this collaborative launch lives outside, or you could give it, its own context block.

PR #282, opened 2021-10-28
Disables Show My Work button when in teacher mode or when in preview without a run key

eireland:
Checks to see if `mode` is `teacher-edition` query param or `preview` are present in the url, and disables Show My Work button from the completion page or the last page of an activity when present.

### DIFF
--- a/src/utilities/report-utils.ts
+++ b/src/utilities/report-utils.ts
@@ -99,7 +99,7 @@ export const getReportUrl = (questionRefId?: string) => {
   }
 };
 
-export const isValidReportLink = getReportUrl() !== null || getReportUrl() !== undefined;
+export const isValidReportLink = getReportUrl() !== null;
 
 export const showReport = () => {
   // Handles not being able to send a null link to window.open
--- a/src/utilities/report-utils.ts
+++ b/src/utilities/report-utils.ts
@@ -72,7 +72,7 @@ export const getReportUrl = (questionRefId?: string) => {
     // We know this is a IPortalData because there is no runKey
     const portalData = getPortalData() as IPortalData;
     // Handles logged in teachers that are previewing a non-assigned resource or are in teacher-edition mode
-    if (!portalData?.offering.id) {
+    if (!portalData?.offering) {
       return null;
     } else {
       const classInfoUrl = portalData?.portalJWT?.class_info_url;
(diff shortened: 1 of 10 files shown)


src/utilities/report-utils.ts
```diff
@@ -67,34 +67,44 @@ export const getReportUrl = (questionRefId?: string) => {
...
+        result += "&iframeQuestionId=" + refIdToAnswersQuestionId(questionRefId);
+      }
+      return result;
+    }
   }
-
-  return result;
 };
 
+export const isValidReportLink = getReportUrl() != null;
```
scytacki, line 102: I think this should probably be a function instead of a static property. `getReportUrl` is based off of the portalData. So by calling getReportUrl statically it means this module is probably going to trigger the first call to getPortalData. That might happen too soon if the logic for how getPortalData is implemented is changed. 

So it seems safer to make this a function and then have it just call getReportUrl every time this function is called. If that is a problem it could be optimized to cache the result since the status of `isValidReportLink` shouldn't change for the same params and portal data.

src/utilities/report-utils.ts
```diff
@@ -67,34 +67,44 @@ export const getReportUrl = (questionRefId?: string) => {
...
-    const offeringUrl = encodeURIComponent(offeringBaseUrl + offeringId);
-    const studentId = portalData?.platformUserId;
+    // Handles logged in teachers that are previewing a non-assigned resource or are in teacher-edition mode
+    if (!portalData?.offering.id) {
+      return null;
+    } else {
+      const classInfoUrl = portalData?.portalJWT?.class_info_url;
+      const authDomainUrl = classInfoUrl?.split("/api")[0];
+      const offeringBaseUrl = classInfoUrl?.split("/classes")[0]+"/offerings/";
+      const offeringId = portalData?.offering.id;
```
scytacki, line 81: In a PR I just merged Doug just changed this to be `portalData?.offering?.id`.  
I think you should merge master into this PR and choose to keep it the way it is here. 
The `if (!portalData.offering.id) {` along with the changes to the tests should take care of this. 

If there isn't an offering we'd want to throw an error so we can track down where things went wrong.

scytacki (changes requested):
Overall looks good.

My two requested changes are:
- merge in master and resolve merge conflicts. Please use merge and don't rebase. If you rebase it will be harder to track this thread of comments in the PR, and harder to look at the choices you made when doing the merge.
- switch to using a function for isValidReportLink

scytacki (approved):
This looks good to me.

scytacki:
This decreases the code coverage slightly, but it is in area that is hard to test. So I think it is OK let it go.

PR #283, opened 2021-11-04
Conversion script

eireland:
Adds ability to converts legacy LARA runtime schema to new-sections schema. Conversion happens on load of resource and shown in the new schema version.
Converted sample documents are available for comparison from legacy to new-sections version.
Majority of the file changes are caused by the addition of sample documents in new schema.
Code changes are in `embeddable.tsx`, `section.tsx`, `convert.ts`, `index.ts`, `lara-api.ts`, and `types.ts`


src/data/new-schema/sample-new-sections-question-interactive-simple.json
```diff
@@ -0,0 +1,114 @@
...
+              "custom_click_to_play": false,
+              "inherit_full_window": true,
+              "custom_full_window": false,
+              "inherit_click_to_play_prompt": true,
+              "custom_click_to_play_prompt": null,
+              "inherit_image_url": true,
+              "custom_image_url": null,
+              "library_interactive": {
+                "hash": "a5d53c7ce6f9ec9a143fd06bff9108b42f3ff7ff",
+                "data": {
```
knowuh, line 69: Just making a note to myself to check if we send this out this way in LARA.

src/lara-api.ts
```diff
@@ -18,10 +19,23 @@ export const getActivityDefinition = (activity: string): Promise<Activity> => {
...
+          return resource;
+        }
+      })
+      .then(resolve);
     } else {
       if (sampleActivities[activity]) {
-        setTimeout(() => resolve(sampleActivities[activity]), 250);
+        if (sampleActivities[activity].version === 1) {
+          const convertedActivityResource = convertLegacyResource(sampleActivities[activity]) as Activity;
+          setTimeout(() => resolve(convertedActivityResource), 250);
```
knowuh, line 35: Do you need to do this using setTimeout? I am surprised, but maybe I don't understand something.

src/types.ts
```diff
@@ -111,9 +111,8 @@ export interface IEmbeddablePlugin extends EmbeddableBase {
 }
 
 export type EmbeddableType = IManagedInteractive | IMwInteractive | IEmbeddableXhtml | IEmbeddablePlugin;
-
 export interface SectionType {
-  secondary_column_display_mode: "stacked" | "carousel";
+  secondary_column_display_mode: "stacked" | "carousel" | string;
```
knowuh, line 115: I wonder what other `display_modes` we are expecting?

knowuh (approved):
This looks good. Thank you.

I had a few questions, but I think that this can be merged.

src/convert-old-lara/convert.ts
```diff
@@ -0,0 +1,226 @@
...
+        break;
+      case null:
+        secondaryBlockEmbeddables.push(embeddableWrapper.embeddable);
+        break;
+    }
+  });
+
+  const headerBlockSection = {
+    "layout": "full-width",
+    "is_hidden": false,
```
scytacki, line 129: This could be based on `show_header`. That way if it is a better representation of the original content. If the sections are rendered even with no visible  embeddables this is another good reason to mark it as hidden.

src/convert-old-lara/convert.ts
```diff
@@ -0,0 +1,226 @@
...
+    "secondary_column_collapsible": false,
+    "secondary_column_display_mode": resourcePage.embeddable_display_mode,
+    "embeddables": getEmbeddablesArray(headerBlockEmbeddables, null, headerBlockHidden)
+  };
+  const splitBlockSection = {
+    "layout": sectionLayout,
+    "is_hidden": false,
```
scytacki, line 136: If both `show_interactive` and `show_info_assessment` is false then it might be good to mark this splitBlockSection as hidden. That way if it is a better representation of the original content.

src/lara-api.ts
scytacki: Perhaps this is to make sample activities behave more like regular activities that require being downloaded? 
But I'd agree that it should be taken out, or a comment added explaining why.

scytacki (changes requested):
Looks good overall, I think I found some logic issues though.

Also I'd rename the `/new-schema/` folder to `/version-2/`

scytacki (changes requested):
This looks good.  I found one last issue with how the header block section is configured, otherwise it is good to go.

scytacki (approved):
Looks good to me

PR #284, opened 2021-11-09
Add support for looking up attachments to linked questions [#179906617]

dougmartin:
TODO:

- update to final lara node package versions

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -70,15 +70,12 @@ interface IProps {
   portalData?: IPortalData;
   answerMetadata?: IExportableAnswerMetadata;
   interactiveInfo?: IInteractiveInfo;
-  questionId: string;
 }
 
 export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef((props, ref) => {
   const { url, id, authoredState, initialInteractiveState, legacyLinkedInteractiveState, setInteractiveState, linkedInteractives, report,
     proposedHeight, containerWidth, setNewHint, getFirebaseJWT, getAttachmentUrl, showModal, closeModal,
-    setSupportedFeatures, setSendCustomMessage, setNavigation, iframeTitle, portalData, answerMetadata, interactiveInfo, questionId } = props;
-  const _idNum = parseInt(id, 10);
-  const idNum = isFinite(_idNum) ? _idNum : 0;
+    setSupportedFeatures, setSendCustomMessage, setNavigation, iframeTitle, portalData, answerMetadata, interactiveInfo } = props;
 
   const [ heightFromInteractive, setHeightFromInteractive ] = useState(0);
   const [ ARFromSupportedFeatures, setARFromSupportedFeatures ] = useState(0);
@@ -279,9 +276,8 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
                   collaboratorUrls: null,
                   classInfoUrl: portalData?.portalJWT?.class_info_url ?? "",
                   interactive: {
-                    id: idNum,
-                    name: "",
-                    questionId
+                    id,
+                    name: ""
                   },
                   authInfo: {
                     provider: "",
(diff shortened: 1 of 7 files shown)


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -279,7 +280,8 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
                   classInfoUrl: portalData?.portalJWT?.class_info_url ?? "",
                   interactive: {
                     id: idNum,
-                    name: ""
+                    name: "",
+                    questionId
```
scytacki, line 284: To match my comments in the LARA PR, the id field here should be changed from `id: idNum` to just `id`. So it picks up the id prop.

I'd note that in the case of the AP these id values will be the old format `123-ManagedInteractive` instead of `managed_interactive_123`.  But as long as they are consistent in the AP answers the code shouldn't care.  It is unfortunate that we couldn't get them unified before the AP starting being used: https://docs.google.com/document/d/1vhRfcrxPRMtJRkHDqYbDPXKPDCqZV5JjjKoSmqB5zMg/edit# 
Implemented.

src/firebase-db.ts
```diff
@@ -477,7 +477,10 @@ export const getLegacyLinkedInteractiveInfo = (embeddableRefId: string, laraData
           activityName: linkedRef?.activity.name,
           interactiveState: answers[index]?.interactiveState || null,
           updatedAt: answers[index]?.meta.created,  // created is same as updated as it is set on each write
-          externalReportUrl: getReportUrl(linkedRefId) || undefined
+          externalReportUrl: getReportUrl(linkedRefId) || undefined,
+          interactive: {
+            questionId: linkedRefId
```
scytacki, line 482: should switch to `id` if possible

src/types.ts
```diff
@@ -301,4 +301,9 @@ export interface ILegacyLinkedInteractiveState {
   linkedState?: any;
   allLinkedStates?: IInteractiveStateProps[];
   externalReportUrl?: string;
+  interactive?: {
+    id: number;
+    name: string;
+    questionId: string;
+  }
```
scytacki, line 308: should be updated to match other changes.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -70,12 +70,13 @@ interface IProps {
...
+  questionId: string;
 }
 
 export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef((props, ref) => {
   const { url, id, authoredState, initialInteractiveState, legacyLinkedInteractiveState, setInteractiveState, linkedInteractives, report,
     proposedHeight, containerWidth, setNewHint, getFirebaseJWT, getAttachmentUrl, showModal, closeModal,
-    setSupportedFeatures, setSendCustomMessage, setNavigation, iframeTitle, portalData, answerMetadata, interactiveInfo } = props;
+    setSupportedFeatures, setSendCustomMessage, setNavigation, iframeTitle, portalData, answerMetadata, interactiveInfo, questionId } = props;
   const _idNum = parseInt(id, 10);
   const idNum = isFinite(_idNum) ? _idNum : 0;
```
scytacki, line 81: I think the questionId prop can go away and you can just use the id property.  The id property is set to the interactiveId which is set to embeddable.ref_id. So it has the same value as the questionId property.

The idNum stuff can go away if we change the id field to be a string and just pass the id prop. See the comments in the LARA PR for more details.

scytacki (changes requested):
Overall looks good. 

My requested change is to remove the `questionId` field and update the `id` field to be a string with a "typed" value. This is part of the plan I created a while ago here: https://docs.google.com/document/d/1vhRfcrxPRMtJRkHDqYbDPXKPDCqZV5JjjKoSmqB5zMg/edit#  

It might be best to look at the LARA PR comments first.

scytacki (approved):
Looks good to me.

For the record: 
I was concerned about the two different id formats being used here. The `embeddable.ref_id` will be in the format used by the lara export. Which currently is `123-ManagedInteractive`. But the report service needs `managed_interactive_123`.
I checked and the `getAnswer` method used automatically converts from one to the other so we are safe.

PR #288, opened 2021-11-17
Add tests

eireland:
Adds back jest and cypress tests during continuous integration.
portal-launch.test in cypress integration is skipped for now, and should be fixed when new-sections is merged to master.


src/components/activity-page/section.test.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import { Section } from "./section";
+import { configure, fireEvent, render } from "@testing-library/react";
```
knowuh, line 3: minor linting error here: `fireEvent` is never used. Same with `getByText` below.

src/data/sample-activity-1.json
```diff
@@ -4,198 +4,198 @@
   "layout": 0,
   "name": "Single Page Test Activity",
   "notes": "",
-  "project": null,
+  "project": {
```
knowuh, line 7: I am wondering if this `sample-activity-1.json` file is in the new format, or the old format, and whether or not that matters.

I noticed that a number of the tests were changed to load data from the `../data/version-2/` path, which made sense. But now I am wondering, was this activity changed/updated in place by accident?

knowuh (approved):
This looks good to me. 

I have one important question about whether one of the original sample activity was updated-in-place as by accident. 

I also highlighted some unused vars pointed out by the linter, which should probably addressed, but are not critical.

src/components/activity-page/section.test.tsx
eireland: I was hoping to keep the collapsible header test in the unit test. Decided to move it to Cypress test instead.

src/data/sample-activity-1.json
eireland: Since we can now convert documents on the fly, I reverted sample-activity-1 to its original form (legacy LARA).
Jest tests point to the version-2 activities because the conversion doesn't happen on the fly.

knowuh (approved):
Great, looks good! Thanks!

knowuh (approved):
looks good to me, thank you.

PR #290, opened 2021-11-30
fixes bug when placing embeddables on left or right

eireland:


knowuh (approved):
Looks good, thank you!

PR #292, opened 2021-12-09
Updating portal-report version number that is used in Show My Work to…

tejal-shah:
… 4.6.0


scytacki:
When I looked at this actions log: https://github.com/concord-consortium/activity-player/runs/4474566483?check_suite_focus=true 
It seems like it failed on a different test than the one that you disabled. 

The one you disabled is a finicky one, that launches the AP from learn staging and relies on previous data saved in the report-service to be able to test the lara sharing plugin. So yes it is fine to disable it for now.

PR #293, opened 2021-12-14
Browser reload should load the same page

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180560270

[#180560270]

- Adds a function for determining a page's ID by its position value. 
- Updates `page` URL param when user navigates to a new page.


eireland (approved):
LGTM

PR #295, opened 2021-12-16
Fix linked interactive preview [#180446636]

dougmartin:
This adds the created field to anonymous answers which is used by the CFM to determine when to show the preview of the two linked states.


scytacki (approved):
looks good to me

PR #297, opened 2021-12-17
Fixes extra space to the right of content

eireland:
This removes the empty space to the right of the embeddables.
image
image


emcelroy (approved):
This looks good. I only wonder if lines 65 and 85 in sections.scss should be removed instead of commented out?

eireland:
Good catch, I thought I got them all

PR #300, opened 2022-01-04
Page nav and Show My Work button issues when hidden pages are present

eireland:
Fixes making Show My Work button when there are hidden pages. 
Disables forward arrow when on last page and there are hidden pages


emcelroy (approved):
Looks great and works as expected locally -- thanks 👍

PR #302, opened 2022-01-06
Shows hidden pages with a banner when in author-preview

eireland:
This goes with work in new-sections LARA. This allows user to preview all pages regardless of whether they are hidden or not.
When `author-preview=true` query param exists, all pages will be rendered including hidden pages. Hiddenn pages will have a purple banner warning author that page is hidden.


src/components/activity-page/activity-page-content.tsx
```diff
@@ -38,15 +38,18 @@ export class ActivityPageContent extends React.PureComponent <IProps> {
...
-        <div className="name">{ pageTitle }</div>
-        {this.renderSections(sections, totalPreviousQuestions)}
-        { enableReportButton &&
-          <BottomButtons
-            onGenerateReport={this.handleReport}
-          />
-        }
-      </div>
+      <>
+      { page.is_hidden && this.renderHiddenWarningBanner()}
```
emcelroy, line 42: Remove the space after the initial { here?

eireland:
@emcelroy Added the hidden page icon to match the LARA nav pages as you suggested.

emcelroy (approved):
Looks good 👍 I had thought the icon would go in the banner but that works, too. It's probably even better.

PR #303, opened 2022-01-07
Fixes page nav and question numbering when there are hidden pages

eireland:
Fixes numbering in page nav when there are hidden pages to match authoring page nav

Fixes question numbering when there are hidden pages


emcelroy (approved):
Looks good to me and works when tested locally 👍

PR #306, opened 2022-01-11
Fixes author preview tab title

eireland:
This is to fix the previous PR on tab titles. It didn't take into account author preview mode, and so AP would show blank pages when author tried to preview an activity that had hidden pages.

### DIFF
--- a/src/utilities/activity-utils.ts
+++ b/src/utilities/activity-utils.ts
@@ -163,10 +163,16 @@ export const setAppBackgroundImage = (backgroundImageUrl?: string) => {
 
 export const setDocumentTitle = (activity: Activity | undefined, pageNumber: number) => {
   if (activity) {
-    const visiblePages = activity.pages.filter(p => !p.is_hidden);
-    document.title = pageNumber === 0
+    if(queryValue("author-preview")) {
+      document.title = pageNumber === 0
       ? activity.name
-      : `Page ${pageNumber} ${visiblePages[pageNumber - 1].name || activity.name}`;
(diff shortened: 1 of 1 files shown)


emcelroy (approved):
This looks good to me 👍

I'm not sure but wonder if something more concise like the below would achieve the same thing?

`const visiblePages = activity.pages.filter(p => !p.is_hidden);`
`const pageTitle = queryValue("author-preview")`
`  ? `Page ${pageNumber} ${activity.pages[pageNumber - 1].name || activity.name}`
`  : `Page ${pageNumber} ${visiblePages[pageNumber - 1].name || activity.name}`
`document.title = pageNumber === 0 ? activity.name : pageTitle;`

src/utilities/activity-utils.ts
```diff
@@ -162,16 +162,18 @@ export const setAppBackgroundImage = (backgroundImageUrl?: string) => {
...
+    const visiblePages = activity.pages.filter(p => !p.is_hidden);
+    if (queryValue("author-preview")) {
+      setTabTitle(activity.pages);
     } else {
-      const visiblePages = activity.pages.filter(p => !p.is_hidden);
-      document.title = pageNumber === 0
-        ? activity.name
-        : `Page ${pageNumber} ${visiblePages[pageNumber - 1].name || activity.name}`;
+      setTabTitle(visiblePages);
     }
```
kswenson, line 177: Could simplify this to something like:
```typescript
    setTabTitle(queryValue("author-preview") ? activity.pages : visiblePages);
```

PR #307, opened 2022-01-12
173913078 interactive sizing

eireland:
Fixes question interactive sizing to respect aspect ratio setting.
Can be tested [here](https://activity-player.concord.org/branch/interactive-sizing/index.html?activity=interactive-sizing-demo-question-interactive).
Behavior can be compared to this [activity](https://authoring.staging.concord.org/activities/21411) Description of desired behavior is in the intro.


emcelroy (commented):
I think this looks good and it pretty much works as expected when I test it in a browser. The only issue I saw in the example activity was on page 18. When you collapse the left-hand column, there's a relatively large gap between the collapsed column and the interactive. That gap doesn't exist in the similar examples on pages 17 and 19.

image

Should that be fixed? Or am I misunderstanding how the sizing should behave there?

Also, it looks like test coverage could be improved, especially for src/components/activity-page/managed-interactive/managed-interactive.tsx. Was adding tests problematic? Or is that still in the works?

PR #308, opened 2022-01-14
Footer page nav does not enable when all questions submitted

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177098917

[#177098917]

Adds a unique key to the ActivityNav component to ensure that both instances on a page are updated when all required questions are submitted.

### DIFF
--- a/cypress/integration/workspace.test.ts
+++ b/cypress/integration/workspace.test.ts
@@ -40,14 +40,14 @@ context("Test the overall app", () => {
         cy.get("[data-cy=version-info]").scrollIntoView().should("be.visible");
     });
   });
-describe("accessibility",()=>{
+  describe("accessibility",()=>{
     it("go to correct page when tabbed and keydown enter from page navigation header",()=>{
-      activityPage.getNavPage(2).type("{enter}");
+      activityPage.getNavPage(2).type("{enter}", { force: true });
       activityPage.getSidebarTab().should("be.visible");
     });
 
     it("go to correct page when tabbed and keydown enter from page list",()=>{
-      cy.get("[data-cy=nav-pages] button").eq(1).type("{enter}");
+      cy.get("[data-cy=nav-pages] button").eq(1).type("{enter}", { force: true });
       cy.get("[data-cy=intro-page-content]").should("be.visible");
       cy.get(".page-item").eq(1).focus();
       cy.focused().type("{enter}");
--- a/cypress/integration/workspace.test.ts
+++ b/cypress/integration/workspace.test.ts
@@ -42,12 +42,12 @@ context("Test the overall app", () => {
   });
   describe("accessibility",()=>{
     it("go to correct page when tabbed and keydown enter from page navigation header",()=>{
-      activityPage.getNavPage(2).type("{enter}", { force: true });
+      activityPage.getNavPage(2).type("{enter}");
       activityPage.getSidebarTab().should("be.visible");
     });
 
     it("go to correct page when tabbed and keydown enter from page list",()=>{
-      cy.get("[data-cy=nav-pages] button").eq(1).type("{enter}", { force: true });
+      cy.get("[data-cy=nav-pages] button").eq(1).type("{enter}");
       cy.get("[data-cy=intro-page-content]").should("be.visible");
       cy.get(".page-item").eq(1).focus();
       cy.focused().type("{enter}");
--- a/src/components/activity-header/activity-nav.tsx
+++ b/src/components/activity-header/activity-nav.tsx
@@ -16,9 +16,8 @@ interface IProps {
 export class ActivityNav extends React.PureComponent <IProps> {
   render() {
     const { activityPages, currentPage, fullWidth, lockForwardNav, onPageChange, singlePage } = this.props;
-    const uniqueKey = `activity-nav-${new Date().getTime()}`;
     return (
-      <div key={uniqueKey} className={`activity-nav ${fullWidth ? "full" : ""}`} data-cy="activity-nav-header">
+      <div className={`activity-nav ${fullWidth ? "full" : ""}`} data-cy="activity-nav-header">
         { !singlePage &&
           <NavPages
             pages={activityPages}
--- a/src/components/activity-header/activity-nav.tsx
+++ b/src/components/activity-header/activity-nav.tsx
@@ -16,8 +16,9 @@ interface IProps {
 export class ActivityNav extends React.PureComponent <IProps> {
   render() {
     const { activityPages, currentPage, fullWidth, lockForwardNav, onPageChange, singlePage } = this.props;
+    const uniqueKey = `activity-nav-${new Date().getTime()}`;
     return (
-      <div className={`activity-nav ${fullWidth ? "full" : ""}`} data-cy="activity-nav-header">
+      <div key={uniqueKey} className={`activity-nav ${fullWidth ? "full" : ""}`} data-cy="activity-nav-header">
         { !singlePage &&
           <NavPages
             pages={activityPages}
--- a/src/components/activity-header/nav-pages.tsx
+++ b/src/components/activity-header/nav-pages.tsx
@@ -143,7 +143,8 @@ export class NavPages extends React.Component <IProps, IState> {
 
   private handleChangePage = (page: number) => () => {
     if (!this.state.pageChangeInProgress) {
-      this.setState({ pageChangeInProgress: true }, () => {
+      const allowPageChange = !this.props.lockForwardNav;
+      this.setState({ pageChangeInProgress: allowPageChange }, () => {
         this.props.onPageChange(page);
       });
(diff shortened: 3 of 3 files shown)


emcelroy:
@eireland @mklewandowski I remembered you both mentioning the importance of using unique keys for components when doing work on Activity Player a while ago. I thought that may be the solution for this bug since the ActivityNav component is used twice on each activity page, but only one instance is disabled when it shouldn't be. 

After adding the unique key, the bug no longer occurs when I test the behavior in a browser. However, that change broke a couple tests in cypress/integration/workspace.test.ts. Somehow Cypress thinks the nav buttons aren't clickable even though if I run the tests with the `--headed` flag, I can see the the buttons being clicked by the automated test and the correct navigation occurring.

The Cypress errors look like this:

image

As I understand it, it's saying it can't type "enter" on the button because it's not a valid typeable element, but then it says button is a typeable element.

I added `force: true` to the type commands and that made the tests pass. The behavior looks correct when I watch the tests run in a browser. It still seems kind of strange and wrong to me, but it's the only thing I could do to get the tests to pass. Do you think that's problematic? Or do you maybe see a flaw in how or where I added the key?

mklewandowski (commented):
This use of a timestamp in the key makes me nervous.  It seems volatile, hard to interpret, and could have unexpected consequences.  I think we would rather use something that is constant for the given activity like an activity id, page number, etc.

kswenson:
>This use of a timestamp in the key makes me nervous. It seems volatile, hard to interpret, and could have unexpected consequences. I think we would rather use something that is constant for the given activity like an activity id, page number, etc.

In general I agree. The result will be to guarantee a re-render whenever anything changes whether it affects the rendering or not. That said, its use here is apparently to make sure a re-render happens under some circumstance in which one should happen but isn't currently happening, in which case the use of a static value like activity id or page number probably won't have the desired effect. Ultimately, the right solution is to figure out what change _should_ trigger a re-render that currently isn't and figure out how to make it trigger one. As a simple for-instance, perhaps if the `<ActivityNav>` component took the number of answered questions as a prop that might have the desired result of triggering a re-render when required questions are answered.

emcelroy:
@mklewandowski @kswenson  Thanks. Looks like I misunderstood the applicability of unique keys for this problem. I also see the point that using a timestamp could be problematic.

As far as I could tell the two instances of `renderNav` in app.tsx were 100% identical. And in both cases a prop related to the number of questions answered -- `lockForwardNav` -- was already being passed down to `ActivityNav` and `NavPages`.

After taking another look, I realized the problem was with the `pageChangeInProgress` state variable in `NavPages`. It's used as one of a few conditions to determine if the nav buttons should be disabled. When you click a disabled nav button, it triggers `handleChangePage` which was setting `pageChangeInProgress` to `true` and then calling `onPageChange` which in the case that you haven't answered all required questions, doesn't change the page. It instead puts up a modal telling you to answer all the questions first. 

The latest change prevents `pageChangeInProgress` from erroneously being set to `true`. Does that look better? I know using the `allowPageChange` variable isn't necessary, but thought it could help clarify a page change may not actually happen. I wonder though if maybe `handleChangePage` should be something more like `handleNavButtonClick`?

kswenson:
>The latest change prevents `pageChangeInProgress` from erroneously being set to `true`. Does that look better? I know using the `allowPageChange` variable isn't necessary, but thought it could help clarify a page change may not actually happen.

This is definitely an improvement. This construct:
```typescript
      const allowPageChange = !this.props.lockForwardNav;
      this.setState({ pageChangeInProgress: allowPageChange }, () => {
        this.props.onPageChange(page);
      });
```
is a little odd in that we're calling `onPageChange()` even when `allowPageChange` is false. One could possibly handle that with name changes but in the interest of expediency a comment may be sufficient.

>I wonder though if maybe `handleChangePage` should be something more like `handleNavButtonClick`?

Or perhaps `handlePageChangeRequest()`?

kswenson:
>I added `force: true` to the type commands and that made the tests pass. The behavior looks correct when I watch the tests run in a browser. It still seems kind of strange and wrong to me, but it's the only thing I could do to get the tests to pass. Do you think that's problematic? Or do you maybe see a flaw in how or where I added the key?

I don't think anyone ever answered the cypress part of your question. I'm not necessarily the best person to do so, but my sense is that cypress can be weird about such things sometimes. There's probably a logical explanation but it's unlikely to be worth the effort to track it down. If `force: true` causes the tests to pass then I'd say run with it. The CLUE cypress tests have a bunch of `force: true`s. 🤷

mklewandowski:
I'm looking at this now as well.  I think I'm reproducing the bug and want to confirm that that is it since the PT story is vague:
- go to page with required questions
- click on one of the disabled nav buttons
- a popup appears indicating that you cannot navigate to the page until answering all questions
- all nav buttons in the nav bar are disabled (which is already a bug since previous pages should not be disabled)
- answer all questions
- nav buttons are still disabled

If this is the bug, then it is indeed tied to the `pageChangeInProgress` state variable.  It is never being set to `false` after the popup is closed.  Even after answering all questions, it is still `false` and all nav buttons remain disabled.

@emcelroy I haven't looked at your solution yet, but from the description, it looks like you are on the right path.

mklewandowski:
Outside the scope of this PR, but I wish we said something better than "Alert" in the popup header.  Even the text likely needs improvement.  It should probably say something more like "Please submit an answer for all required questions first." or something that is clear that there might be more than one required question. 

image

kswenson:
> Outside the scope of this PR, but I wish we said something better than "Alert" in the popup header. Even the text likely needs improvement. It should probably say something more like "Please submit an answer for all required questions first." or something that is clear that there might be more than one required question.

Or even "You must submit an answer for all required questions before advancing to another page."

mklewandowski:
Here is my concern about the current solution.  I know I worked on some of this code originally, but I am struggling to remember why we disabled all of the nav buttons when the page change was in progress.  There must have been something that we were protecting against.  With this solution we're losing that protection when the forward navigation is locked.  When the forward navigation is locked, the backwards navigation (navigating to previous pages) is NOT locked.  So, in theory, we are opening up a path to some bug that we were protecting against by disabling the nav buttons when a page change is in progress.  In other words...
Scenario 1:
- we're on a normal page with all nav buttons enabled
- user presses nav button to navigate to new page
- `pageChangeInProgress` is set true and all nav buttons are disabled while we do whatever we need to do to prepare and display the next page (so in theory we are protecting against the user quickly pressing a different nav button and messing things up)

Scenario 2:
- we're on a page with required questions and all forward nav buttons are disabled
- user CAN, however, press a nav button to navigate to a previous page... when the user does press an enabled nav button...
- `pageChangeInProgress` is NOT changed and remains false (since `lockForwardNav` is true).  Nav buttons are NOT disabled while we do whatever we need to do to prepare and display the previous page.  User is able to press a different enabled nav button and mess things up

So I think we are potentially opening up some other bug.  I think we are on the path to the ideal approach - properly set `pageChangeInProgress`.  However, if we are navigating back or to a previous page, then we still want to set `pageChangeInProgress` to `true` EVEN when `lockForward.Nav` is true

mklewandowski:
@emcelroy @kswenson and if you want to ignore my rambling above, I think we just need make this tweak when setting `allowPageChange `:
```
const allowPageChange = page < this.props.currentPage || !this.props.lockForwardNav;
```

This is of course not taking into consideration Kirk's comments about naming or adding a comment.

mklewandowski:
And if we want to make the buttons all behave the same, then this should be removed from the CSS in line 76 of nav-pages.scss:
```
      &.arrow-button {
        pointer-events: none;
      }
```

It seems wrong that we let the user click on the numeric page buttons that are disabled and show them a message but if the user clicks on the next arrow button we do NOT similarly see the same message.  I think all of these buttons should behave the same and show the message.  If it isn't a hassle, I would remove the above CSS.

mklewandowski (changes requested):
Looking good.  Final things:
- `allowPageChange` should always be true if navigating to a previous page as follows: `const allowPageChange = page < this.props.currentPage || !this.props.lockForwardNav;`
- if it isn't a hassle, we should remove the CSS in `nav-pages.scss` that prevents the popup from showing on the next arrow button.  See my comment.
- if it isn't a hassle, we should tweak the message on the popup. I like Kirk's idea so in `app.tsx` we want `const kDefaultIncompleteMessage = "You must submit an answer for all required questions before advancing to another page.";`
- I agree with Kirk's comment on the Cypress test and his comment on adding a comment or tweaking the name

emcelroy:
Thanks again, @kswenson and @mklewandowski. Though the changes we settled on were simple, I had some weird issues with my copy of the AP code and decided to make a [fresh branch](https://github.com/concord-consortium/activity-player/pull/315).

Regarding the cypress tests, luckily the different approach we're taking made it so `force: true` is no longer needed.

PR #310, opened 2022-01-19
Pins collapsible header when collapsed and interactive is taller

eireland:
Demo [here](https://activity-player.concord.org/branch/pin-collapsible-header/?activity=interactive-sizing-demo-question-interactive&page=page_139538). 
Make the window wide so the interactive is taller than collapsed secondary column. Scroll down, and collapsed header should be pinned.


emcelroy (approved):
Looks good!

PR #311, opened 2022-01-19
fixes bug with horizontal scroll not scrolling the entire width

eireland:
Demo [here] (https://activity-player.concord.org/branch/horizontal-scroll/?activity=interactive-sizing-demo-question-interactive&page=page_126082)
Make the browser narrow so that horizontal scroll bar appears on the bottom

### DIFF
--- a/src/components/app.scss
+++ b/src/components/app.scss
@@ -14,8 +14,7 @@ body {
   display: flex;
   justify-content: center;
   background-color: var(--theme-primary-color);
-  overflow-y: auto;
-  overflow-x: auto;
+  overflow: auto;
 }
 
 .app {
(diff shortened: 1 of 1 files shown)


src/components/app.scss
```diff
@@ -15,6 +15,7 @@ body {
   justify-content: center;
   background-color: var(--theme-primary-color);
   overflow-y: auto;
+  overflow-x: auto;
```
emcelroy, line 18: You could merge these two lines to make things more concise.

```suggestion
  overflow: auto;
```

emcelroy (approved):
Looks good and works as expected. I made one small suggestion for a possible change, but it's not necessary to do.

PR #313, opened 2022-01-21
Only pins the primary interactive if its height is less than the browser height

eireland:
Demo [here](https://activity-player.concord.org/branch/no-pinning/?activity=interactive-sizing-demo-question-interactive&mode=teacher-edition)
On Page 3, if you change the height of the interactive to be larger than the browser window, the interactive should jump to the top of the div and not be pinned. If you change the height of the browser to be shorter than the interactive, the interactive should not be pinned as you scroll.
On page 16, If you open click on the teacher tip and the teacher tip makes the primary column taller, the interactive should not be pinned.
On page 14, it uses the total height of both interactives to determine whether to pin or not.

### DIFF
--- a/src/components/activity-page/section.scss
+++ b/src/components/activity-page/section.scss
@@ -137,9 +137,11 @@
     position: relative;
     height: auto;
     .embeddableWrapper {
-      position: sticky;
       top: 10px;
       align-self: flex-start;
+      &.pinned {
+        position: sticky;
+      }
     }
   }
 
--- a/src/components/activity-page/section.tsx
+++ b/src/components/activity-page/section.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import classNames from "classnames";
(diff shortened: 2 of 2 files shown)


emcelroy (approved):
This looks good to me and the demo works as expected. A few super picky things:

1. I think the indentation for lines 33 and 39-53 should be adjusted to align with the rest of the code.

2. There's a tiny typo in the comment on line 33 ("fwhen")

3. On page 14 of the demo when there are two interactives, if you have a tall browser window and the interactives are pinned while you scroll way down the page, if you then decrease the height of the browser window so its less than the height of the two interactives, the interactives instantly jump back up to the top of their column. It makes it seem like they just disappeared. I don't think it's critical to fix that and I can't think of a straightforward solution for the problem, either. But if you can think of a good solution and it's easy to implement, it'd be nice if that didn't happen.

eireland:
For #3: A similar thing happens in LARA. It doesn't disappear immediately on browser resource, but as soon as you scroll, the interactives disappears.

PR #314, opened 2022-01-21
Fixes styling so there is no gap between collapsed column header and primary column

eireland:
Current behavior is there is a gap between the primary and secondary column when you collapsed the secondary column.
image

This changes the 
Demo [here](https://activity-player.concord.org/branch/collapsed-column-gap/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21423.json)


emcelroy (approved):
Looks good, thanks!

PR #315, opened 2022-01-21
Footer page nav does not enable when all questions submitted.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177098917

[#177098917]

These changes refactor the nav button click handler to better deal with cases when a page change is not allowed. Most importantly, `pageChangeInProgress` is only set to `true` when page changing is allowed. This prevents `pageChangeInProgress` from inadvertently disabling nav buttons when they should be enabled.

### DIFF
--- a/src/components/activity-header/nav-pages.scss
+++ b/src/components/activity-header/nav-pages.scss
@@ -73,9 +73,6 @@
       outline: none;
       cursor: auto;
       opacity: .35;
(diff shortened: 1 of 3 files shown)


mklewandowski (approved):
👍 LGTM!  Thanks for making some of those minor changes too.  I think you managed to fix/improve three issues in this PR!

PR #316, opened 2022-01-21
180968689 pin collapsible header

eireland:
Pins Hide/Show column header regardless of its state.
The original branch only pinned the header when it is collapsed. This makes the header pinned regardless or whether the collapsed or not.


emcelroy (approved):
Looks good 👍

PR #318, opened 2022-01-24
Fix get attachment url for new interactives [#180916194]

dougmartin:
The interactiveId is optional and should only be present for linked interactives.  Before this fix the interactiveId was required and this caused writes of new interactives to fail as they had no existing metadata.

This aligns AP with the Lara code.


scytacki (changes requested):
Do I understand correctly that the `interactiveId` is only present/required when an interactive is requesting an attachment URL for an interactive other than itself? 

If I have that right, can you update or add a comment which clarifies this?

scytacki (approved):
works for me.

PR #320, opened 2022-02-01
Fix all sequences have the same wrong HTML title value

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178676995

[#178676995]

These changes set `document.title` to the sequence title when viewing a sequence's home page. Previously, when viewing a sequence home page, `document.title` would be inadvertently set to "Sample Layout Types".


mklewandowski (approved):
👍 LGTM!

PR #321, opened 2022-02-02
Disables pointer events when page nav is disabled

eireland:
Fixes bug when clicking on a disabled page nav button disables the entire page nav. 
(#177501334)[https://www.pivotaltracker.com/story/show/177501334]
Added a separate class for arrow buttons that sets pointer-events to none. I couldn't just add it to the `disabled` class because we still need the pointer-events for when there is a required question to be submitted. The user is still be able to click on the arrow or the page number after the blocker page, and a warning modal appears that an answer needs to be submitted.

### DIFF
--- a/cypress/integration/activity-page.test.ts
+++ b/cypress/integration/activity-page.test.ts
@@ -61,7 +61,7 @@ context("Test the overall app", () => {
       activityPage.getHomeButton().eq(1).click();
       cy.wait(1000);
       activityPage.getPreviousPageButton().should("have.class", "last-page");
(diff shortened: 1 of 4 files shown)


emcelroy (approved):
This looks good. I was a little confused at first because I assumed `last-page` referred to the actual last page in the sequence, then I thought "last" was being used as a synonym for "previous", then I finally figured it out. I can't really think of a better term to use, though. `limit-page`? It's probably fine as is and I'd just leave it as is if you think `last-page` is best.

PR #322, opened 2022-02-02
Page scrolls to the top when user navigates to a different page

eireland:


src/components/app.tsx
```diff
@@ -459,6 +459,15 @@ export class App extends React.PureComponent<IProps, IState> {
     window.location.href = this.portalUrl;
   }
 
+  private scrollToTop = () => {
+    const app = document.getElementById("app");
+    app?.style.setProperty("position", "relative");
+    window.scroll(0,0);
+    if(app?.clientTop === 0) {
```
emcelroy, line 466: I don't think `clientTop` is what we want to check here. See https://developer.mozilla.org/en-US/docs/Web/API/Element/clientTop

I think `scrollTop` is probably what we should use instead. See https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTop, although I do wonder if this check to see if the page is scrolled all the way up is necessary. Did you find that it was in testing this? Or am I misunderstanding what this is for?

Also, being super picky, there should be a space after the if.

emcelroy (changes requested):
This is looking good, but see my inline comment for some changes I think should be made.

src/components/app.tsx
emcelroy: Sorry but I still think changes need to be made. I don't think this is checking what you think it's checking. `clientTop` is the "width of the top border of an element in pixels." So this is checking if `#app` has a top border of zero, which it always does. 

If you want to check if the scroll is complete, I'm pretty sure you would want to use `scrollTop`. Even then though, the fact that this works with checking `clientTop` makes me think the check isn't really necessary. I'm not knowledgable enough to know for sure, but I suspect what's happening is the if statement is adding _just enough_ more time that the scroll gets to finish before the position is set back to fixed. 

When testing this, I do see that the scroll doesn't always work when the if statement is removed. I suspect there's a better way to make sure the scroll is complete before changing the position back, but I don't know what it is. Sorry. It doesn't look like there's a callback option for `scroll` unfortunately.

If you disagree with that or can't find a different way to make this work, we should get another opinion before leaving it as is.

src/components/app.tsx
```diff
@@ -471,7 +480,8 @@ export class App extends React.PureComponent<IProps, IState> {
     } else if (page >= 0 && (activity && page <= activity.pages.length)) {
       const navigateAway = () => {
         this.setState({ currentPage: page, incompleteQuestions: [] });
-        setDocumentTitle({activity, pageNumber: page});
+        setDocumentTitle(activity, page);
```
emcelroy, line 483: This line should be changed back to `setDocumentTitle({activity, pageNumber: page});`.

eireland:
Ending this PR because of conflicts to master

PR #323, opened 2022-02-02
Scrolls page to the top when there is a page change

eireland:
Adds function to scroll page to the top when user changes page.


emcelroy (approved):
This looks great -- thanks!

PR #326, opened 2022-02-10
AP Runtime supports new footer fields in Project form

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181059679

[#181059679]

These changes add support for new Project fields in LARA (see https://github.com/concord-consortium/lara/pull/901). Basically, when values for `project.copyright` and/or `project.collaborators` are present, those values along with optional image URLs will be used to build the footer content. If values for those properties are absent, the deprecated LARA runtime footer content (if present) will be used instead.

Note that I removed a couple sets of style rules from src/components/activity-introduction/footer.scss. They weren't being used anywhere as far as I can tell, and seemed simple enough that they could easily be reconstructed if they end up being needed in the future. I also didn't see anything that seemed related in the AP specs, so thought it was safe to remove them.


mklewandowski (approved):
👍 This looks good to me.  Probably worth making sure Evangeline has a chance to look too since I'm a little out of the loop on the Activity Player.  But overall, this looks good!

eireland (approved):
LGTM!

PR #329, opened 2022-02-14
Fix state saving on page change

pjanik:
AP should request all the interactives state on page change. This was initially implemented here: https://github.com/concord-consortium/activity-player/pull/103.

However, this feature got broken when sections support was added and the `Section` component was inserted between `ActivityPageContent` and `Embeddable` components.

This PR also updates Cypress tests related to data saving. We had tests that could actually detect the problem - they were saving user data and changing pages, but... there were explicit `cy.wait` commands that prevented that from being actually tested and detected (also, comments suggested like this feature never existed). So, I've updated these tests to focus on two separate cases:
- automatic saving after a delay
- automatic saving on page change

I tried to comment out my recent updates and the Cypress test failed as expected.

### DIFF
--- a/cypress/integration/data-saving-anonymous.ts
+++ b/cypress/integration/data-saving-anonymous.ts
@@ -19,23 +19,21 @@ context("Saving and loading data as an anonymous user", () => {
     });
   });
 
-  describe("Saving and loading data", () => {
+  describe("Data saving", () => {
     const runKey = uuidv4();
     const activityUrl = "?activity=sample-activity-1&enableFirestorePersistence";
 
-    it("we can use a runKey to retrieve data previously persisted", () => {
+    it("happens automatically after a small delay", () => {
+      // Data is being saved in two ways:
+      // 1. After a small delay (to prevent spamming the database)
+      // 2. When user changes the activity pages using header navigation.
+      // This test focuses on 1. case.
       const activityUrlWithRunKey = activityUrl + "&runKey=" + runKey;
-
       cy.visit(activityUrlWithRunKey + "&clearFirestorePersistence");
       activityPage.getNavPage(2).click();
       getInIframe("body", "[data-cy=choices-container] input").eq(1).click({force: true});
 
-      // unfortunately we have to wait for the data to be posted to firestore, which happens after a
-      // delay to prevent spamming the database.
-      // The fastest and most reliable way to kick it off seems to be 1. blur, 2. wait, 3. navigate away
-      cy.get("[data-cy=account-owner").click({force: true});
       cy.wait(3000);
-      activityPage.getNavPage(1).click();
 
       // We are essentially reloading the page here but in this case we are not clearing the persistance first
       // this should force the activity player to load the data back in from the runKey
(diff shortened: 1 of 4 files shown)


pjanik:
To test that manually, you can first open this AP activity:
https://activity-player.concord.org/?activity=https%3A%2F%2Fauthoring.concord.org%2Fapi%2Fv1%2Factivities%2F12565.json
Then, go to Page 1, type anything in Open Response text area, and immediately click "Page 2" button (without losing focus on the text area). Then, go back to Page 1. Your answer is most likely lost, as it's the production branch.

If you repeat these steps here (PR branch):
https://activity-player.concord.org/branch/fix-state-saving-on-page-change/index.html?activity=https%3A%2F%2Fauthoring.concord.org%2Fapi%2Fv1%2Factivities%2F12565.json
Your data should not be lost.

scytacki (approved):
Looks good to me. 
As I understand it, the issue was that the old code was looking for embeddable refs in the embeddableRefs map. But that map was not being populated with any refers. 

The new code delegates this to the section component, which now records the embeddable refs for all of the embeddables it is rendering. 

I don't understand why we didn't see errors about the unused properties. VSCode shows this when hovering over the variable, but it doesn't report it as a warning:
image

We could use https://www.typescriptlang.org/tsconfig#noUnusedLocals, to make this an error. I tried turning it on on the code in mater, and it only picked up these errors and couple in plugins.ts.
I'd guess that there are legitimate cases where you want locally unused properties, so by default TS is setup to just inform about them and not warn about them.

scytacki:
Because this is causing lost learner data I'm going to go ahead and merge it so we can start to test it.

pjanik:
> Looks good to me. As I understand it, the issue was that the old code was looking for embeddable refs in the embeddableRefs map. But that map was not being populated with any refers.

Yes. embeddableRefs map used to be populated when `ActivityPageContent` was rendering `Embeddable`s directly. However, `Embeddable`s rendering has been moved to `Section`. Section was populating `embeddableRefs` similarly to `ActivityPageContent`, but it was done incorrectly (missing useRef in functional component), and actually this hash was never used by anything.

At the same time, `ActivityPageContent` was trying to use its own, empty `embeddableRefs` hash. Because it was actually used and read, I don't think that any linting would catch that problem. 

Generally, ActivityPageContent is requesting a particular action (interactive state request) from its children, and this chain of calls needs to go down to `IframeRuntime`. The whole chain and logic were broken by `Section` component that was inserted in the middle and didn't try to forward the call.

> The new code delegates this to the section component, which now records the embeddable refs for all of the embeddables it is rendering.

Yes. It does what `ActivityPageContent` used to do. And it also exposes imperative API that is now used by the parent component (`ActivityPageContent`).


> I don't understand why we didn't see errors about the unused properties. VSCode shows this when hovering over the variable, but it doesn't report it as a warning: image
> 
> We could use https://www.typescriptlang.org/tsconfig#noUnusedLocals, to make this an error. I tried turning it on on the code in mater, and it only picked up these errors and couple in plugins.ts. I'd guess that there are legitimate cases where you want locally unused properties, so by default TS is setup to just inform about them and not warn about them.

I've played with that too. It seems there are two separate things:

1. Our `"@typescript-eslint/no-unused-vars"` confing in `.eslint.rc` doesn't seem to even warn us about unused vars. When I changed the config to `"@typescript-eslint/no-unused-vars": ["warn", { "vars": "all", "args": "after-used", "ignoreRestSiblings": false }],`, it fixes the problem. I'd also set that to "error" in `.eslintrc.build.js`. "warn" is useful in development / debugging, as error breaks compilation and it's annoying. But I think there's no reason for unused vars in the final code.

2. `"noUnusedLocals": true` is defined in `tsconfig.json` and it works for me. It lists issues that are listed in VSCode. It's a different set of errors (class properties). I guess it might be worth turning on too.

But as I said, none of these settings would prevent us from this error. `embeddableRefs` was actually used and read, but just never populated. I don't think linters could catch that.

The best thing that could catch this error would be the integration test. We almost had it, but the `cy.wait` let AP save answer using different code path.

PR #330, opened 2022-02-17
Added interactive.id to report state when saving to Firebase [#181103123]

dougmartin:
This is needed for the CFM interactive api to process and load the interactive state when viewing the interactive in the portal-report.


src/utilities/embeddable-utils.ts
```diff
@@ -74,6 +74,9 @@ export const getAnswerWithMetadata = (
     // LARA also seems to be storing the interactiveState as a JSON string and that
     // JSON string is what is used in the reportState
     interactiveState: JSON.stringify(interactiveState),
+    interactive: {
+      id: embeddable.ref_id
```
scytacki, line 78: In the other place in the state the id is included it uses `refIdToAnswersQuestionId` to convert the id if necessary.
Can you look into this to understand what format the id needs to be in for things to work in the report?

If this id needs to be a in the `123-MangedInteractive` format instead of `managed_interactive_123`, then please update the interactive ids document. It currently refers to `adding an answer to the Report Service` but doesn't specify that the id actually appears in 2 places in the answers. https://docs.google.com/document/d/1vhRfcrxPRMtJRkHDqYbDPXKPDCqZV5JjjKoSmqB5zMg/edit

scytacki (changes requested):
I think the format of the id needs to be converted so it matches other interactive ids in the report service. If the id format doesn't need to be converted then it at least should be documented, and perhaps we need to figure out if we can use the converted form to make the id format unification easier in the future.

scytacki (changes requested):
I haven't looked at this closely, but I'm pretty sure the bare id is wrong. 

In the id migration document, using bare id is described as a bug. This is because there can be conflicts between the iframe and managed interactives. 

In the PR associated with the linked commit, I wrote: 

> But if you want to leave this as is, that is fine. This structure is not one that will last after the LARA runtime shutdown.

and 

> If the interactive_id is changed to be the typed id, then this could go away. And the type of the interactive_id would have to be changed to string.
>
> But again this object format is temporary so it'd be fine to leave it alone.

In this case the changes are the AP which will be permanent, and they are being stored in the report-service which would mean we would have to migrate them.

Why did you want to use the bare id here instead of the typed form `managed_interactive_123`?

scytacki (changes requested):
Doug sorry to do this to you, but I've changed my mind. 

As I was updating the [id document](https://docs.google.com/document/d/1vhRfcrxPRMtJRkHDqYbDPXKPDCqZV5JjjKoSmqB5zMg/edit#heading=h.gztpyrmuf9gp) I realized it would be better to store the snake_case version of the id even though it is inconsistent with the runtime id format.

If we store the `123-ManagedInteractive` style id in the report initMessage this id will not be compatible with the ids that the portal-report is using in other places. In the portal-report it does not request the lara export of the activity. It only works with the structure published to the report-service. And that structure only uses the `managed_interactive_123`.  So if we are ever to support linked interactives in portal-report that will be broken unless the `interactive.id` is the new `managed_interactive_123` format. Or the portal-report does some conversion in some parts. Since the AP is already doing this conversion I'd prefer to try to limit the conversion to fewer systems. 

So even though it is inconsistent with the runtime, I think we should use the `managed_interactive_123` format in the report initInteractive message sent by the AP to firestore.

What I didn't look at is how the portal-report handles attachments. I seem to remember that the attachments api uses the interactive.id. But I didn't look at it to see if I'm remembering right, or how it is handling that.  Up until this PR the portal-report would not be sending the the interactive id to the interactive anyway though. So that attachment code must handle the case where there is no id. Now that there will be an id, which I think will not match, I wonder what will happen.

Can you look into that? Or just try using the `refIdToAnswersQuestionId ` conversion function so the report initMessage stored by the AP is the snake case of the id, and then see if attachments work in the report?

dougmartin:
@scytacki I updated to the snake case but now attachment saves are not completing, with no errors in the console.  I'll have to look at this later today.

scytacki (approved):
looks good to me

PR #332, opened 2022-02-25
Add reset button for interactives

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181169398

[#181169398]

These changes add the missing "Undo all my work" button for interactives. The button has been relabelled "Clear & start over." 

- The button should appear directly below interactives when the option is checked in the authoring form for iframe interactives or is set in the library interactive for a managed interactive. 
- When the button is clicked, the iframe should reload and be reset to its original state. (Note that this is different than LARA runtime where clicking the button causes the entire page to reload.) 
- Any linked interactives should remain linked to the reloaded interactive.


dougmartin (approved):
Looks good - I like the use of the increment to the key to reload the iframe.

PR #334, opened 2022-03-08
Added unloading option [#181169438]

dougmartin:
This adds the new `unloading` option to the `getInteractiveState` message which lets the interactive know it is being removed due to a page change.  With [this CFM PR](https://github.com/concord-consortium/cloud-file-manager/pull/260) SageModeler/CODAP interactives will now have their final state saved instead of the last state from the autoSave timer.

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -8,7 +8,7 @@ import {
   IGetInteractiveSnapshotResponse, IInitInteractive, ILinkedInteractive, IReportInitInteractive,
   ISupportedFeatures, ServerMessage, IShowModal, ICloseModal, INavigationOptions, ILinkedInteractiveStateResponse,
   IAddLinkedInteractiveStateListenerRequest, IRemoveLinkedInteractiveStateListenerRequest, IDecoratedContentEvent,
-  ITextDecorationInfo, ITextDecorationHandlerInfo, IAttachmentUrlRequest, IAttachmentUrlResponse
+  ITextDecorationInfo, ITextDecorationHandlerInfo, IAttachmentUrlRequest, IAttachmentUrlResponse, IGetInteractiveState
 } from "@concord-consortium/lara-interactive-api";
 import Shutterbug from "shutterbug";
 import { Logger } from "../../../lib/logger";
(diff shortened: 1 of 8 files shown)


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -20,7 +20,7 @@ import { getReportUrl } from "../../../utilities/report-utils";
 
 const kDefaultHeight = 300;
 
-const kInteractiveStateRequestTimeout = 2000; // ms
+const kInteractiveStateRequestTimeout = 20000; // ms
```
scytacki, line 23: This seems too long to me. If a student is waiting 20s without any UI feedback after hitting the next button they are going to think something is broken. 2s seems much better. 

I know why this is bumped up, but I think we have to add some kind of feedback. I added a new story for this: https://www.pivotaltracker.com/story/show/181511767 so it doesn't need to hold up this PR.

scytacki (approved):
This looks good to me. I just added a new story about providing feedback now that the timeout is 20s instead of just 2s.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
dougmartin: Just for documentation - this was changed to align it with the other two 20 second timeouts:

- the firebase request tracker: https://github.com/concord-consortium/activity-player/blob/master/src%2Ffirebase-db.ts#L71
- the ajax request tracker: https://github.com/concord-consortium/activity-player/blob/master/src%2Futilities%2Frequest-tracker.ts#L1 (this is the default, the callers use the Firebase 20 second timeout)

If these are not aligned the user can switch pages and unload the interactive before the request finishes which then causes an unrelated error message to show on the new page when the 20 second timeout fails.

kswenson: > Just for documentation - this was changed to align it with the other two 20 second timeouts:
> 
> * the firebase request tracker: https://github.com/concord-consortium/activity-player/blob/master/src%2Ffirebase-db.ts#L71
> * the ajax request tracker: https://github.com/concord-consortium/activity-player/blob/master/src%2Futilities%2Frequest-tracker.ts#L1 (this is the default, the callers use the Firebase 20 second timeout)
> 
> If these are not aligned the user can switch pages and unload the interactive before the request finishes which then causes an unrelated error message to show on the new page when the 20 second timeout fails.

Seems like that comment should be in the code.

PR #335, opened 2022-03-15
Record Open Response in runtime

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181223914

[#181223914]

### DIFF
--- a/src/components/activity-page/managed-interactive/managed-interactive.test.tsx
+++ b/src/components/activity-page/managed-interactive/managed-interactive.test.tsx
@@ -79,6 +79,7 @@ jest.mock("@concord-consortium/interactive-api-host", () => ({
 const mockWatchAnswer = jest.fn((id: string, callback: (answer: any) => void) => callback({ meta: {} }));
 jest.mock("../../../firebase-db", () => ({
   watchAnswer: (id: string, callback: (answer: any) => void) => mockWatchAnswer(id, callback),
+  getAnswer: () => { return { answerType: "multiple_choice_answer", selectedChoiceIds: []}; },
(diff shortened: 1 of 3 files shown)


src/components/activity-page/managed-interactive/managed-interactive.test.tsx
```diff
@@ -210,4 +212,87 @@ describe("ManagedInteractive component", () => {
...
+              questionNumber={1}
+              setSupportedFeatures={mockSetSupportedFeatures}
+              setSendCustomMessage={mockSetSendCustomMessage}
+              setNavigation={mockSetNavigation}
+              />);
+    expect(screen.getByTestId("managed-interactive")).toBeInTheDocument();
+    // allow initialization to complete
+    jest.runAllTimers();
+
+    act(() => {
```
dougmartin, line 282: I was just curious after @scytacki remarked about `act` in one of my PRs and it looks like it is not needed when we use (like we do in this repo) React Testing Library (@testing-library/react) as it is baked in to the helpers.

I don't think anything needs to be done for this PR but we should all be looking out for it in new tests.

More info: 

- https://reactjs.org/docs/testing-recipes.html#act
- https://javascript.plainenglish.io/you-probably-dont-need-act-in-your-react-tests-2a0bcd2ad65c

> You might find using act() directly a bit too verbose. To avoid some of the boilerplate, you could use a library like [React Testing Library](https://testing-library.com/react), whose helpers are wrapped with act().

dougmartin (approved):
Looks good :+1:

scytacki (approved):
looks good to me

PR #336, opened 2022-03-21
Fix interactive reset button functionality.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181169398

[#181169398]

With the work done in https://github.com/concord-consortium/activity-player/pull/332 the reset handler was resetting interactives to the state they had when the page loaded. That would include any work that had been saved in previous sessions. These changes correct that so the interactive is fully reset to its default state, with no saved work at all.

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -85,12 +85,13 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
 
   const [ heightFromInteractive, setHeightFromInteractive ] = useState(0);
   const [ ARFromSupportedFeatures, setARFromSupportedFeatures ] = useState(0);
-  const [resetInteractive, setResetInteractive] = useState(false);
   const [reloadCount, setReloadCount] = useState<number>(0);
+  const iframePhoneTimeout = useRef<number|undefined>(undefined);
   const iframeRef = useRef<HTMLIFrameElement>(null);
   const phoneRef = useRef<IframePhone>();
   const setInteractiveStateRef = useRef<((state: any) => void)>(setInteractiveState);
   setInteractiveStateRef.current = setInteractiveState;
+  const interactiveStateRef = useRef(initialInteractiveState);
   const linkedInteractivesRef = useRef(linkedInteractives?.length ? { linkedInteractives } : { linkedInteractives: [] });
   const interactiveStateRequest = {
     promise: useRef<Promise<void>>(),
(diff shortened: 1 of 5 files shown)


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -85,12 +85,13 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
 
   const [ heightFromInteractive, setHeightFromInteractive ] = useState(0);
   const [ ARFromSupportedFeatures, setARFromSupportedFeatures ] = useState(0);
-  const [resetInteractive, setResetInteractive] = useState(false);
   const [reloadCount, setReloadCount] = useState<number>(0);
+  const [iframePhoneTimeout, setIframePhoneTimeout] = useState<NodeJS.Timeout|undefined>(undefined);
```
kswenson, line 89:
```suggestion
  const [iframePhoneTimeout, setIframePhoneTimeout] = useState<number|undefined>(undefined);
```
since we're running in the browser

emcelroy: Thanks, @kswenson. Using `number` resulted in an error:

`Argument of type 'Timeout' is not assignable to parameter of type 'SetStateAction<number | undefined>'.`

Does the latest change to line 89 seem like the right solution?

kswenson: It looks better because it's not explicitly wrong, but it's still implicitly wrong. If you hover over the value you can still see that it has type `NodeJS.Timeout | undefined`, but at runtime we know that it's the browser version of `setTimeout` that will be called and that the return value will be a number, so using `Node.JS.Timeout` is just telling the TypeScript compiler to type-check against the wrong type. The optimal solution would be to figure out what's introducing the node.js types into the application (I tried removing `node` from the `types` in the `tsconfig.json` and it didn't solve the problem, which suggests the types are being inserted somewhere _in addition_ to the `tsconfig.json`). I think the next-best solution would be to use `number` and then use `window.setTimeout` at the point of call, which is inelegant but at least correct.

emcelroy: OK, I see. I've made the changes you suggested.

knowuh (approved):
Looks good to me.

We might consider using `useRef` for the value set by `setTimeout`.

I am not too worried about coverage issues on this one.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -85,12 +85,13 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
 
   const [ heightFromInteractive, setHeightFromInteractive ] = useState(0);
   const [ ARFromSupportedFeatures, setARFromSupportedFeatures ] = useState(0);
-  const [resetInteractive, setResetInteractive] = useState(false);
   const [reloadCount, setReloadCount] = useState<number>(0);
+  const [iframePhoneTimeout, setIframePhoneTimeout] = useState<number|undefined>(undefined);
```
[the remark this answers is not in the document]
dougmartin: I agree that `useRef` would be better here.

dougmartin (approved):
Looks good as is but you could replace the `useState` with `useRef`.

PR #337, opened 2022-03-25
Skip saving iframe runtime interactive state when there is no change [#181175814]

dougmartin:
This avoids updating the timestamp that is used when comparing linked interactive states.


scytacki (approved):
Looks good to me.

I'm not sure all of those `act(` are necessary, but they were there before.

PR #338, opened 2022-03-29
DEBUG/NOMERGE: Added debugging info to save interactive state

dougmartin:
This is for testing on staging to see why state is being saved when there apparently is no change.


dougmartin:
Closing to cleanup PR list.

PR #341, opened 2022-03-30
Changing version used by AP portal-report to 4.10.0 which is used by …

tejal-shah:
…"Show my work"


scytacki (approved):
Looks good to me.

PR #342, opened 2022-04-04
Uncouple Show "Clear & start over" button from Enable save state setting

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181775027

[#181775027]

These changes ensure that the Clear & Start Over button is displayed only when `enable_learner_state` is true _and_ `show_delete_data_button` is true. Previously, only the `show_delete_data_button` property was being used to determine if the Clear & Start Over button should be shown. That resulted in the button being present even when an embeddable doesn't save state.


src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -274,15 +274,23 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
     }
   }));
 
+  const setShowDeleteDataButton = () => {
+    return embeddable.type === "MwInteractive"
+             && embeddable.enable_learner_state
+             && embeddable.show_delete_data_button
+           || embeddable.type === "ManagedInteractive"
+             && embeddable?.library_interactive?.data.enable_learner_state
+             && embeddable?.library_interactive?.data.show_delete_data_button;
```
scytacki, line 283: because you are already assuming `embeddable` is not undefined in the earlier lines I thin you can remove the `?` from the end. So then it would be 
`embeddable.library_interactive?.data.enable_learner_state`

Other than this, it seems good to me.

scytacki (approved):
I have one suggestion, but otherwise it looks good.
You don't need to wait for me to review it again.

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -274,15 +274,23 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
     }
   }));
 
+  const setShowDeleteDataButton = () => {
```
kswenson, line 277: I see this has already been merged, but `setShowDeleteDataButton` implies that the function sets something when it actually just returns a value, so `getShowDeleteDataButton` or `computeShowDeleteDataButton` might be more appropriate.

emcelroy: @kswenson Thanks for the suggestion. I'll keep my eye out for a chance to change that in the future. I vacillated between `setShowDeleteDataButton` and `getShowDeleteDataButton` but didn't think either was quite right. I went with `setShowDeleteDataButton` but I think `computeShowDeleteDataButton` would be better.

kswenson: Another option would be `shouldShowDeleteDataButton()`.

emcelroy: Yeah, slightly but definitely even better.

PR #343, opened 2022-04-06
Add attachments to initInteractive [#181729937]

dougmartin:


scytacki (changes requested):
Looks good, but it should have some test coverage.

PR #347, opened 2022-05-02
Fixed glossary popup persisting on time out and page change. [#176434143] [#181727884]

lublagg:


src/lara-plugin/plugin-api/popup.ts
```diff
@@ -110,6 +112,7 @@ export const addPopup = (_options: IPopupOptions): IPopupController => {
     }
     $dialog.remove();
```
dougmartin, line 113: You are missing the code after the remove call to update the list, something like this:

`removeList = removeList.filter(fn => fn !== remove);`

src/components/app.tsx
```diff
@@ -23,7 +23,7 @@ import { IPortalData, IPortalDataUnion } from "../portal-types";
 import { signInWithToken, initializeDB, setPortalData, initializeAnonymousDB,
          onFirestoreSaveTimeout, onFirestoreSaveAfterTimeout, getPortalData } from "../firebase-db";
 import { Activity, IEmbeddablePlugin, Sequence } from "../types";
-import { initializeLara, LaraGlobalType } from "../lara-plugin/index";
+import { initializeLara, LaraGlobalType, PluginAPI_V3 } from "../lara-plugin/index";
```
dougmartin, line 26: It might be better to directly import the `__closeAllPopups()` call so that it doesn't look like it is part of the API, something like:

`import { __closeAllPopups } from "../lara-plugin/plugin-api/popup.ts";`

dougmartin (changes requested):
Just a couple small changes needed.

src/lara-plugin/plugin-api/popup.ts
```diff
@@ -104,12 +106,17 @@ export const addPopup = (_options: IPopupOptions): IPopupController => {
...
+
   const remove = () => {
     if (options.onRemove) {
       options.onRemove();
     }
     $dialog.remove();
   };
+
+  removeList.push(remove);
+  removeList = removeList.filter((fn) => fn === remove);
```
dougmartin, line 118: This needs to be inside the remove method so that it removes itself from the list when called, like this:

```
  const remove = () => {
    if (options.onRemove) {
      options.onRemove();
    }
    $dialog.remove();
    removeList = removeList.filter((fn) => fn === remove);
  };
```

dougmartin (changes requested):
Almost right... :)

src/lara-plugin/plugin-api/popup.ts
lublagg: So close! Thank you for your patience!

dougmartin (approved):
Looks good :+1:

PR #355, opened 2022-05-18
Added click to play support [#181169226]

dougmartin:
Added support for click to play for interactives and their plugins.


src/components/activity-page/managed-interactive/click-to-play.tsx
```diff
@@ -0,0 +1,25 @@
...
+  prompt = prompt || "Click here to start the interactive.";
+
+  const handleOnClick = (e: React.MouseEvent<HTMLDivElement>) => {
+    e.preventDefault();
+    onClick();
+  };
+
+  return (
+    <div className="click-to-play" onClick={handleOnClick} data-cy="click-to-play">
+      {imageUrl && <img src={imageUrl} />}
```
scytacki, line 21: Was LARA doing anything to scale this image if it was too big?

src/utilities/plugin-utils.ts
```diff
@@ -136,11 +136,18 @@ export const initializePlugin = (context: IEmbeddablePluginContext) => {
 
   let embeddableContext: IEmbeddableContextOptions | null = null;
   if (wrappedEmbeddable && wrappedEmbeddableContainer) {
+    let clickToPlay = false;
+    if (wrappedEmbeddable.type === "ManagedInteractive") {
+      clickToPlay = !!wrappedEmbeddable.library_interactive?.data.click_to_play;
```
scytacki, line 141: Will this `!!wrappedEmbeddable.library_interactive?.data.click_to_play` ever change to true if the interactive has click_to_play enabled?

I would expect the `interactiveAvailable` property of the emeddableContext to switch from false to true when a click to play interactive is clicked. But maybe LARA didn't work that way?

dougmartin: LARA, and this code, uses the `LARA.Events.emitInteractiveAvailable` method to let the plugins know that the interactive is available after the click to play.

dougmartin:
> Can you add a test for an iframe (MW) interactive too? Ie. not a library interactive. I'm not sure if your current code supports that, and I didn't see a test for it. Dan typically will use iframe interactives for COAP and Sage instead of library interactives.
> 
> I also think I found an issue where the context passed to the plugin is not being updated, but I'm not sure how this was actually handled in LARA.

I didn't add a test because they are both handled by the same code:

```
if (embeddable.type === "MwInteractive" || (embeddable.type === "ManagedInteractive" && embeddable.library_interactive)) {
    qComponent = <ManagedInteractive
	@@ -83,7 +92,9 @@ export const Embeddable: React.ForwardRefExoticComponent<IProps> = forwardRef((p
                    questionNumber={questionNumber}
                    setSupportedFeatures={handleSetSupportedFeatures}
                    setSendCustomMessage={setSendCustomMessage}
                    setNavigation={handleSetNavigation}
                    emitInteractiveAvailable={handleEmitInteractiveAvailable}
                 />;
  }
```

src/components/activity-page/managed-interactive/managed-interactive.test.tsx
```diff
@@ -295,4 +296,68 @@ describe("ManagedInteractive component", () => {
...
+      inherit_image_url: true,
+      custom_image_url: null,
+      library_interactive: {
+        hash: "af5d1860b2d4a037ae01e3d86931a30886fcb42d",
+        data: {
+          aspect_ratio_method: "DEFAULT",
+          authoring_guidance: "",
+          base_url: "https://models-resources.concord.org/question-interactives/branch/master/multiple-choice/",
+          click_to_play: true,
+          click_to_play_prompt: "Click here to play!!!",
```
scytacki, line 329: I've looked at the code now to verify that this will work even these properties are up 2 levels on a MwInteractive. But I'd still like a duplicate of this test to make sure that keeps working in the future. Especially since that MwInteractive case is the more common one.

scytacki (changes requested):
I'd still like the MwInteractive specific test since that is the more common use case.

src/components/activity-page/managed-interactive/managed-interactive.test.tsx
dougmartin: I've added a test of the plain MwInteractive in this new commit: https://github.com/concord-consortium/activity-player/pull/355/commits/ac9544eb1a4b60d75df90ec580064ac60d546feb

scytacki (approved):
Looks good.

PR #356, opened 2022-05-19
Added ap_runs collection to save which page the user last viewed per activity [#181175850]

dougmartin:
NOTE: there is a lot moved around in the `componentDidMount` method in order to get the ap_run info from Firebase before using the sequence/activity/page info.


src/firebase-db.ts
```diff
@@ -494,3 +497,72 @@ export const getLegacyLinkedInteractiveInfo = (embeddableRefId: string, laraData
...
+  } else {
+    apRun = {
+      type: "anonymous",
+      run_key: portalData.runKey,
+      ...common
+    };
+  }
+
+  const firestoreSetPromise = app.firestore()
+    .doc(apRunsPath(existingApRun?.id))
```
dougmartin, line 562: Now that I have it on GitHub looking at the diffs this might be buggy - it will overwrite an existing document with a new sequence_activity if that is the first time the sequence_activity is set.

scytacki:
Looking at the code I'm not sure if this will save the page of activities that are run outside of sequences.

Also, did you figure out if the LARA Runtime remembers pages within each activity of a sequence? Ie if you go to Act 1 and then navigate to page 3, then go to Act 2 and navigate to page 2. Then you switch back to Act 1.  At this point are you on Act 1, page 1 or page 3?

dougmartin:
@scytacki - could you review this again.  I updated some tests and merged master back in to resolve a merge conflict.

scytacki (changes requested):
I found a few things I had question about.

Also it seems like the current activity is not being saved when running a sequence?
Did I miss that somewhere, or maybe that wasn't requested?
I'm pretty sure that LARA will take you into activity you last opened when you launch a sequence the second time.

dougmartin:
@scytacki I forgot about this limbo PR - I noticed it this morning when reviewing my assigned and created PRs.  I've added a commit (https://github.com/concord-consortium/activity-player/pull/356/commits/a9ee7215064ec9ca25f5d878e01c1e9f83d75846) to address your review comments.

src/firebase-db.ts
```diff
@@ -494,3 +497,82 @@ export const getLegacyLinkedInteractiveInfo = (embeddableRefId: string, laraData
...
+  if (sequenceActivity) {
+    query = query.where("sequence_activity", "==", sequenceActivity);
+  }
+
+  const doc = await query.get();
+  if (doc.empty) {
+    return null;
+  }
+
+  return {id: doc.docs[0].id, data: doc.docs[0].data() as IApRun};
```
scytacki, line 527: I don't see any sorting here. It seems to just be using the first document that it finds. So if if the activities in the sequence are visited in order lets say 1, 2, 3. Then the user goes back to activity 2. And then they quit. Now if they launch the sequence again, I'd expect it won't open activity 2. I tested it with this URL: 
https://activity-player.concord.org/branch/save-page-position/?sequence=sample-sequence
And as expected it opens activity 1 instead of activity 2.

Can you add a test for this kind of navigation and reloading? It seems like that part of the code isn't covered by the existing tests in this PR.

scytacki (changes requested):
Most of the changes look good, but it doesn't look like it remembers the activity of the sequence correctly unless I'm doing something wrong.

src/components/app.tsx
```diff
@@ -519,17 +537,37 @@ export class App extends React.PureComponent<IProps, IState> {
...
+    let currentPage = 0;
+    const {sequence} = this.state;
+    const sequenceActivity = sequence !== undefined ? getSequenceActivityId(sequence, activityNum) : undefined;
+    if (sequenceActivity && sequence) {
+      const activity = sequence.activities[activityNum];
+      const apRun = await getApRun(sequenceActivity);
+      if (apRun) {
+        currentPage = getPagePositionFromQueryValue(activity, `page_${apRun.data.page_id}`);
+      }
+      setQueryValue("sequenceActivity", sequenceActivity);
```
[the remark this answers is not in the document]
dougmartin: I added this last week here: https://github.com/concord-consortium/activity-player/pull/356/commits/73dae368f88caae0257e6b41c4ccb0d25945b9da

dougmartin:
@scytacki could you re-review?  I believe I have addressed all the comments except for one.  I've been fighting with Cypress today to try to get a test where it reloads and goes back to the saved page.  It works when manually testing but not during the automated tests.

scytacki (approved):
Looks great to me.

PR #357, opened 2022-05-26
Added page change  notifications [#181511767]

dougmartin:
The page change notification displays a please wait message after 500ms on a page change along with an error with a 3000ms delay.  Previously no please wait message was shown and any error was shown in a Javascript alert.


scytacki (approved):
Looks good to me.

The error case in app.tsx is not covered by tests, but I'd assume that would a bit annoying to cover. We'd need to use some time modification library so it doesn't need to wait for the timeout.

Maybe @tejal-shah could cover this with cypress.

kswenson:
> The error case in app.tsx is not covered by tests, but I'd assume that would a bit annoying to cover. We'd need to use some time modification library so it doesn't need to wait for the timeout.

Jest has support for fake timers which should eliminate the wait, although error cases can be hard to test for other reasons as well.

PR #360, opened 2022-06-03
Fixed check of answers in completion page [#182143506]

dougmartin:
This uses code adapted from portal-report to determine if an answer has a response or is empty.

### DIFF
--- a/src/components/activity-completion/completion-page-content.tsx
+++ b/src/components/activity-completion/completion-page-content.tsx
@@ -5,9 +5,9 @@ import IconUnfinishedCheck from "../../assets/svg-icons/icon-unfinished-check-ci
 import { isValidReportLink, showReport } from "../../utilities/report-utils";
 import { Sequence, Activity, EmbeddableType, Page } from "../../types";
 import { renderHTML } from "../../utilities/render-html";
-import { watchAllAnswers } from "../../firebase-db";
+import { watchAllAnswers, WrappedDBAnswer } from "../../firebase-db";
 import { isQuestion } from "../../utilities/activity-utils";
-import { refIdToAnswersQuestionId } from "../../utilities/embeddable-utils";
+import { answerHasResponse, refIdToAnswersQuestionId } from "../../utilities/embeddable-utils";
 import { SummaryTable, IQuestionStatus } from "./summary-table";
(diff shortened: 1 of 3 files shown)


dougmartin:
@scytacki / @emcelroy - I tagged both of you for review since you both were involved in updating or reviewing the similar portal report code.

scytacki (changes requested):
As discussed in the meeting today, this should check if the answer is required it should only be marked as completed if the answer has been submitted.

dougmartin:
I don't think we can support the equivalent of this code in portal-report in AP:

```
if (question.get("required") && !answer.get("submitted")) {
    return false;
  }
```

Looking between Lara and AP there seems to be no way to know if a question is required.  In LARA we export `required` based on the `is_prediction` value of the base embeddable types but not those types handled by AP which are:

`IManagedInteractive | IMwInteractive | IEmbeddableXhtml | IEmbeddablePlugin`

We also don't reference either `required` or `is_prediction` anywhere in the AP code.

The closest we come is in the navigation component where we maintain a state of incomplete questions via this but that is used at runtime on each interactive in response to the `navigation` interactive iframe api handler:

```  
private handleSetNavigation = (refId: string, options: INavigationOptions) => {
    const { incompleteQuestions } = this.state;
    const qIndex = incompleteQuestions.findIndex((q: IncompleteQuestion) => q.refId === refId);
    const updatedIncompleteQuestions = [...incompleteQuestions];
    if (qIndex >= 0 && options.enableForwardNav) {
      updatedIncompleteQuestions.splice(qIndex, 1);
      this.setState({ incompleteQuestions: updatedIncompleteQuestions });
    } else if (qIndex < 0 && !options.enableForwardNav) {
      const newIncompleteQuestion: IncompleteQuestion = { refId, navOptions: options };
      updatedIncompleteQuestions.push(newIncompleteQuestion);
      this.setState({ incompleteQuestions: updatedIncompleteQuestions });
    }
  }
```

scytacki (changes requested):
As we discussed in slack, there is a way to extract the `required` field from the interactive state. So it seems best (but ugly) to go ahead and do that. It is what LARA does in order to put the required field into the report service for interactive questions.

scytacki (approved):
Looks good to me.

PR #361, opened 2022-06-15
Added conversion of production glossary plugin urls [#182392985]

dougmartin:
This change converts the s3Url in the glossary plugin data to use the new glossary api endpoint for matching production glossaries.


dougmartin:
@scytacki I forgot to assign this to you last week for review.

PR #362, opened 2022-06-15
WIP: Added await to setInteractiveState to force write to Firebase [#182039050]

dougmartin:
This hopefully will force a write of the answer to Firebase as currently there is sometimes a delay of a few seconds.


dougmartin:
Not needed.

PR #363, opened 2022-06-27
Handle activities and sequences with defunct flag

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182465001

[#182465001]

When an activity or sequence has its `defunct` property set to `true`, display a banner at the top of every page warning that it contains obsolete elements.

The `defunct` property is added in [this LARA PR](https://github.com/concord-consortium/lara/pull/971).

This is how the banner looks in the browser:

image

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -16,6 +16,7 @@ import { getActivityDefinition, getSequenceDefinition } from "../lara-api";
 import { ThemeButtons } from "./theme-buttons";
 import { SinglePageContent } from "./single-page/single-page-content";
 import { WarningBanner } from "./warning-banner";
+import { DefunctBanner } from "./defunct-banner";
 import { CompletionPageContent } from "./activity-completion/completion-page-content";
 import { deleteQueryValue, queryValue, queryValueBoolean, setQueryValue } from "../utilities/url-query";
 import { fetchPortalData, firebaseAppName } from "../portal-api";
@@ -70,6 +71,7 @@ interface IState {
   currentPage: number;
   teacherEditionMode?: boolean;
   showThemeButtons?: boolean;
+  showDefunctBanner: boolean;
   showWarning: boolean;
   username: string;
   portalData?: IPortalData;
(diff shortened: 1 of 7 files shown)


emcelroy:
There was no spec for the banner, so I made up the text.

I'm wondering if I should just drop the "may"? So it would say "This resource contains obsolete features that no longer work."

kswenson:
> There was no spec for the banner, so I made up the text.
> 
> I'm wondering if I should just drop the "may"? So it would say "This resource contains obsolete features that no longer work."

Or perhaps "... that are no longer supported." and then you don't have to specify whether they may not or won't work.

src/components/app.tsx
```diff
@@ -226,10 +232,12 @@ export class App extends React.PureComponent<IProps, IState> {
       }
 
       const showThemeButtons = queryValueBoolean("themeButtons");
+      // Show a warning about obsolute features if activity/sequences is marked as defunct
+      const showDefunctBanner = activity.defunct || (sequence?.defunct);
```
dougmartin, line 236: Just a style thing but the parens about the sequence defunct check are not needed.

src/components/app.tsx
```diff
@@ -279,11 +287,14 @@ export class App extends React.PureComponent<IProps, IState> {
   }
 
   render() {
+    // tslint:disable-next-line
+    console.log("state", this.state);
```
dougmartin, line 291: Again, is this for debugging or meant to be left in?

dougmartin (approved):
Looks good :+1:   Just a couple of questions about if the two console.log statements that were added are meant to stay.

I also agree with @kswenson that it might be better to use `that are no longer supported`.  

Either way I think this is going to generate some webmaster questions.  Maybe we should also include a link to a static page in the activity player that explains what it means that things are no longer supported and then we could change add a "more info" link after the not supported statement.

scytacki (approved):
I'm didn't look closely, but it seems good to me.

PR #370, opened 2022-07-29
Fixed zero showing at top of sequence pages when defunct flag is off [#182853964]

dougmartin:
This also adds a sample-sequence-defunct.json file to view when the flag is enabled.


emcelroy (approved):
Looks good -- thanks for fixing that!

PR #371, opened 2022-08-01
Fix some embedded images look disproportionately big

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182873531

[#182873531]

This change makes it so images embedded in text blocks will conform to any height and width values specified in authoring provided the specified width is not wider than the image's containing column. In that case, the image will be constrained to the width of the column, and its height will be automatically adjusted proportionally.


eireland (approved):
Looks good!

PR #373, opened 2022-08-15
Fixed interactiveAvailable message passing with sharing plugin [#182921431]

dougmartin:
The issue was that the interactiveAvailable message being sent when click to play was clicked was not being routed to the wrapped embeddable the sharing plugin was wrapping.  There was a check to see if the message was generated by the container of the wrapped embeddable and this has been changed to be that plus any parent container.


scytacki (approved):
This seems fine.

I'm not sure how often this is called, but it seems like it will be inefficient when it doesn't find a match. Perhaps if there are lots of interactives on the page this could become a problem, but I doubt it.

PR #374, opened 2022-08-16
Fix iframes in text boxes and sidebars not being rendered

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182908653

[#182908653]

Adds `iframe` to the default set of allowed HTML tags, and adds some related attributes to the set of allowed attributes.


scytacki (approved):
looks good to me.

I don't remember about the border on iframes. I'll ask Dan in the story...

scytacki:
Oh I see that Kiley just said to go without borders. 👍

PR #375, opened 2022-08-17
Added deprecated "loadInteractive" message [#182875673]

dougmartin:
This mirrors what LARA does so that legacy interactives that do not use the new initInteractive API can continue to work in AP.


scytacki (approved):
Seems fine to me.

It would be nice if we could use versioning so we don't send this to "new" interactives. But that would be a bigger change, so this is good as it is.

dougmartin:
@scytacki Turns out when testing with the table interactives I missed a conditional that LARA has - it only sends the loadInterative message when there is existing interactive state.  I've added that to this commit: https://github.com/concord-consortium/activity-player/pull/375/commits/e5270c53af0512b961547f03e7f25ea75ee8fc9a

dougmartin:
@scytacki This now fixes the table interactives, however the table data is only saved when using the navigation - it doesn't ping ever 5 seconds asking for the interactive state like LARA does.

PR #377, opened 2022-08-19
Added domain to host features in initInteractive message [#182916725]

dougmartin:
Also added `bluetooth; clipboard-read; clipboard-write` to the allow list of the iframe.


scytacki (approved):
looks good to me

PR #378, opened 2022-08-22
Add OAuth support and set teacher logging username using Portal JWT

pjanik:
[#181821831]

This PR adds OAuth support to Activity Player. Implementation is based on the Portal Report one.

Also, when the Portal JWT can be obtained, but the user type is not "learner", AP will set the logging username to `user_id@portal_domain` (eg 123@learn.concord.org). This matches what LARA used to do, what Dashboard does, and what AP does when launched in Teacher Edition mode straight from Portal.

It'll let researchers match Dashboard log events with ones coming from AP Teacher Edition, even if AP Teacher Edition is opened directly from Dashboard (not from Portal).

@scytacki, note that OAuth path cannot be used by students who launch AP from Portal at the moment without additional changes. Portal JWT API returns a very limited set of information when OAuth token is used to get Portal JWT. 

When students use short-lived token, JWT includes more information. I believe mostly because Portal saves learner data in the AccessGrant here: https://github.com/concord-consortium/rigse/blob/6d75a19f6d1a45c60c0d9681245d5cdedae61449/rails/app/models/external_report.rb#L37-L44
and this doesn't happen in the OAuth path.

When teacher uses OAuth, we don't have much information in JWT too. Not even user_type. But there was "uid" so it was enough for me to construct a logging username and avoid changing Portal. And it doesn't seem like we need anything more now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6730,6 +6730,11 @@
         "raf-schd": "^4.0.2"
       }
     },
+    "@servie/events": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/@servie/events/-/events-1.0.0.tgz",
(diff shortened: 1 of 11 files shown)


scytacki (approved):
This looks great to me.

PR #379, opened 2022-08-24
Fix loading partially invalid plugins [#183065524]

dougmartin:
LARA is exporting some plugins with approved_script as null, probably due to deleted data.

This will need to be fixed in LARA but this is a quick bandaid until that is done.


scytacki (approved):
seems fine to me

PR #381, opened 2022-09-27
[#182895285]: Interactive is pinned when set to "use all available space."

lublagg:
Note: this does not solve interactive pinning when the interactive is wrapped in a Teacher Edition Question Wrapper.

### DIFF
--- a/src/components/activity-page/embeddable.tsx
+++ b/src/components/activity-page/embeddable.tsx
@@ -136,8 +136,8 @@ export const Embeddable: React.ForwardRefExoticComponent<IProps> = forwardRef((p
       key={embeddable.ref_id}
       ref={targetDiv}
     >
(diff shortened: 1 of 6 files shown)


scytacki (changes requested):
Looks good, just the one issue with `"DEFAULT"`

scytacki:
Sorry I didn't see that you made the change I suggested. If you re-request a review after making addressing my PR comments then I'm more likely to see you've updated the PR. You can re-request the view by clicking the 
image
icon up in the reviewers section of the PR.

Note to future self: I've found it useful to compare this series of PRs against what the code looked like before them. To do that after I merge this you can compare with a tag I just made: `pre-interactive-sizing-updates`.  That version tag is just before this PR branch. It should get deployed to: https://activity-player.concord.org/version/pre-interactive-sizing-updates/

scytacki (approved):
Looks good to me.

PR #384, opened 2022-10-07
[#183457143]: Fixed completion page crashing when "Clear and start over" button is clicked.

lublagg:


scytacki (approved):
Looks good to me.
I'm surprised this didn't cause a problem sooner.

PR #388, opened 2022-11-16
[#183748003]: Fix line-height affecting visibility of text.

lublagg:
Before with line-height on embeddableWrapper set to 0:
Screen Shot 2022-11-16 at 10 52 37 AM
Screen Shot 2022-11-16 at 10 51 51 AM

After removing line-height set to 0:
Screen Shot 2022-11-16 at 10 52 01 AM
Screen Shot 2022-11-16 at 10 52 10 AM


pjanik (approved):
Looks good!

PR #389, opened 2022-12-05
Update logger url [#175337534]

dougmartin:


dougmartin:
**DO NOT MERGE**: need to wait until production log-ingester is adding ids to the logs.

src/lib/logger.ts
```diff
@@ -5,8 +5,8 @@ import { LaraGlobalType } from "../lara-plugin/index";
 type LoggerEnvironment = "dev" | "production";
 
 const logManagerUrl: Record<LoggerEnvironment, string> = {
-  dev: "//cc-log-manager-dev.herokuapp.com/api/logs",
-  production: "//cc-log-manager.herokuapp.com/api/logs"
+  dev: "//logger.concordqa.org/logs",
+  production: "//logger.concord.org/logs"
```
scytacki, line 9: Will http work with these URLs? Perhaps it would be better to hard code them to https?
The only place this will really mater is in dev. If someone is running without https locally I'd guess they will get a bunch of errors or redirects which might or might not save.

scytacki (changes requested):
Maybe hard code to https?

scytacki (changes requested):
This seems fine to me.

However, there is a failing test. It looked random, but I tried re-running it and it failed again in the same place:
```
 1) Launch AP From the Portal
       non-collaboratively
         can open a sample activity:
     AssertionError: Timed out retrying after 8000ms: Expected to find element: `[data-cy=activity-summary]`, but never found it.
      at Context.eval (http://localhost:8080/__cypress/tests?p=cypress/e2e/portal-launch.test.ts:167:50)
```

I just re-ran a master build to see if there is something in a remote service that has caused this to start failing:
https://github.com/concord-consortium/activity-player/actions/runs/3485386979

If master fails the same way, then I'd be fine merging this PR, and working on the failing test in a new PR. Adding a story to track that would be helpful.

dougmartin:
@scytacki It is also failing on master so I added a PT story to fix it:

https://www.pivotaltracker.com/story/show/184047920

scytacki (approved):
Sounds good.

PR #391, opened 2022-12-16
Fix failing cypress test causing deployment failures

tejal-shah:


dougmartin (approved):
Looks good :+1:   Thanks for doing this!

PR #397, opened 2023-01-31
fix: Clear & start over button should not be activated when tabbing away

pjanik:
[[#184349849]](https://www.pivotaltracker.com/story/show/184349849)


emcelroy (approved):
Looks good 👍

PR #399, opened 2023-02-17
feat: Added read aloud toggle with saving in session storage (PT-184466103) (PT-184466599) (PT-184466624)

dougmartin:
- Added new top of page read aloud toggle (PT-184466103)
- Added saving of read aloud value in session storage (PT-184466599)
- Added check for text to speech support with message (PT-184466624)

### DIFF
--- a/src/components/activity-page/activity-page-content.scss
+++ b/src/components/activity-page/activity-page-content.scss
@@ -10,11 +10,18 @@
     width: 100%;
   }
 
-  .name {
-    font-size: 24px;
-    font-weight: bold;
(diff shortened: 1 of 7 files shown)


dougmartin:
@scytacki I'm not worried about the lack of test code coverage right now - all this code is going to be moved around to other repos and we can add tests when that is done.

dougmartin:
@scytacki I'm going to close this and the other interim read aloud PRs and move/split all the code using the design we discussed yesterday.  No need to clutter the git history with these commits.

PR #401, opened 2023-02-18
build: Fixed build by pinning to Node 16

dougmartin:
This also monkeypatches the app test to make crypto.getRandomValues() available at runtime.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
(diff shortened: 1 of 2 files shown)


dougmartin:
@scytacki I first tried updating to Node 18 (the version in `ubuntu-latest` in GitHub actions) and upgrading the cli to webpack 5 but it caused a bunch of issues.  

I ended up with this getting the builds to work:

1. Pin to Node 16
2. Monkeypatch around an error in the app test thrown by the `uuid` dependency in the `lara-api-host`

If you could review this I can get it merged and then move my feature branch and PT branches to use it so I can get them to build for demo links.

scytacki (approved):
seems fine to me

PR #406, opened 2023-02-22
feat: Added read aloud toggle with saving in session storage (PT-184466103) (PT-184466599) (PT-184466624)

dougmartin:
- Added new top of page read aloud toggle ([PT-184466103](https://www.pivotaltracker.com/story/show/184466103))
- Added saving of read aloud value in session storage ([PT-184466599](https://www.pivotaltracker.com/story/show/184466599))
- Added check for text to speech support with message ([PT-184466624](https://www.pivotaltracker.com/story/show/184466624))


scytacki:
@dougmartin can you have someone else look at this, I'm a bit swamped. 
@pjanik maybe you could take a look.

dougmartin:
Branch demo url: https://activity-player.concord.org/branch/add-read-aloud-toggle

src/components/app.tsx
```diff
@@ -45,6 +45,7 @@ import { IPageChangeNotification, PageChangeNotificationErrorTimeout, PageChange
 import { getBearerToken } from "../utilities/auth-utils";
 
 import "./app.scss";
+import { isReadAloudAvailable, loadReadAloudSetting, saveReadAloudSetting } from "../lib/read-aloud-code-to-refactor";
```
dougmartin, line 48: I'll fix the placement of this import in the next PR.

src/components/activity-page/activity-page-content.tsx
```diff
@@ -20,6 +21,9 @@ interface IProps {
   setNavigation: (refId: string, options: INavigationOptions) => void;
   pluginsLoaded: boolean;
   pageChangeNotification?: IPageChangeNotification;
+  readAloud?: boolean;
+  setReadAloud?: (readAloud: boolean) => void;
+  readAloudDisabled?: boolean;
```
pjanik, line 26: These 3 properties are added to multiple interfaces, so there are quite a few copies of this code.
Wouldn't it make sense to create some base interface that would include them?
I guess all these page-related components might have some elements in common, like this new toggle.

src/components/activity-page/activity-page-content.tsx
```diff
@@ -45,7 +49,14 @@ export class ActivityPageContent extends React.PureComponent <IProps> {
...
         {this.renderPageChangeNotification()}
         <div className={`page-content full ${isResponsiveLayout ? "responsive" : ""}`} data-cy="page-content">
-          <div className="name">{ pageTitle }</div>
+          <div className="header">
+            <div className="name">{ pageTitle }</div>
+            {this.props.setReadAloud && <ReadAloudToggle
+              disabled={this.props.readAloudDisabled}
+              isChecked={this.props.readAloud || false}
+              onChange={this.props.setReadAloud}
+            />}
```
pjanik, line 58: These few lines are also copied and pasted into multiple comments. Maybe some helper function that could take props could be useful?

It feels to me that MobX or another library that lets us avoid passing all the properties from top to bottom could simplify AP quite a bit. But that's obviously another task.

pjanik (approved):
Looks good. I've left two inline comments about the code that is repeated in multiple spots. But it's just a suggestion or question, not a merge blocker.

src/components/activity-page/activity-page-content.tsx
kswenson: Seems like React context would make sense in this context, since I believe it's a top-level setting that affects all of the components within it.

dougmartin: We are using a context for the read aloud component (now called "dynamic text" to allow for future merge with the glossary highlighter and possibly text translation).  I have move this code into its own provider when I do that work today.

PR #407, opened 2023-02-24
feat: Add read aloud using dynamic text library (PT-184475161)

dougmartin:


src/components/app.tsx
```diff
@@ -61,6 +62,22 @@ const kAnonymousUserName = "Anonymous";
 
 const kDefaultFixedWidthLayout = "1100px";
 
+// this is exported to that the TextBox component can pass it directly to DynamicText
```
scytacki, line 65:
```suggestion
// this is exported so the TextBox component can pass it directly to DynamicText
```

scytacki (approved):
Looks good. I found one grammar mistake in a comment, that was all.

PR #408, opened 2023-02-27
build: Remove s3_deploy.sh and s3_website.yml

dougmartin:
These are no longer needed as the ci.yml s3-deploy task now uses the s3-deploy-action


scytacki (approved):
looks good

PR #409, opened 2023-02-27
Added log events for read aloud (PT-184466607)

dougmartin:
- Logs toggle of "Tap text to listen" component
- Logs start of read aloud including text read
- Logs cancel of read aloud including text read


scytacki (approved):
looks good

PR #404, opened 2023-02-21
build: Updated s3-deploy to use s3-deploy-action

dougmartin:

PR #412, opened 2023-03-03
ci: Rename 'workflows/release-production.yml.yml' to 'workflows/release_production.yml'

433eros:
Saw this after viewing GH-404


dougmartin (approved):
Looks good :+1:

PR #415, opened 2023-03-08
fix: Fixes from read aloud QA (multiple stories)

dougmartin:
- [PT-184631349] Remove read aloud for sequence link
- [PT-184631109] Fix z-index of read aloud toggle to not overlap sidebar content
- [PT-184631417] Fix unregistering of iframed components

Also removed separate require from dynamic-text/component to avoid duplicate code as it also exists in the root dynamic-text module.

### DIFF
--- a/src/components/activity-page/managed-interactive/iframe-runtime.tsx
+++ b/src/components/activity-page/managed-interactive/iframe-runtime.tsx
@@ -102,6 +102,7 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
   const currentInteractiveState = useRef<any>(initialInteractiveState);
 
   const dynamicText = useDynamicTextContext();
+  const dynamicTextComponentIds = useRef<Set<string>>(new Set());
 
   useEffect(() => {
     const initInteractive = () => {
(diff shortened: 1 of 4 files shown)


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -368,6 +372,15 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
 
     // Cleanup.
     return () => {
+      // Unregister all components that were proxied through this iframe handler so
+      // we don't leave dangling registered components and so the current component is
+      // silenced if it is being read when the iframe unloads
+      // disable warning about dynamicTextComponentIds as it is not referencing a DOM node
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
pjanik, line 379: You can avoid disabling eslint if you add something like that **above** the cleanup function:

```
const currentDynamicTextComponentIds = dynamicTextComponentIds.current;
```

pjanik (approved):
Looks good. 👍 
I think you can avoid disabling eslint in one place - I left an inline comment.

PR #416, opened 2023-03-09
Added read aloud toggle to sequence homepage [PT-184662647]

dougmartin:
Also refactored code to use a context for the read aloud settings instead of drilling them down as props everywhere.


dougmartin:
@pjanik / @kswenson You'll be happy to see I refactored to a context for the read aloud settings in this PR after having to add the read aloud toggle to yet another page level component in AP.

kswenson (approved):
👍 Looks good!

PR #418, opened 2023-03-15
feat: Added support for hide read aloud setting [PT-184688393]

dougmartin:
When the hide read aloud setting is enabled for the activity or sequence the toggle button is not displayed and the read aloud is disabled without persisting the setting to local storage.

If the sequence hide read aloud setting is enabled this overrides the activity hide read aloud setting.

NOTE: there are two new samples, one for an activity with hide read aloud enabled and a sequence with hide read aloud enabled.  This will allow testing before the setting is made available in LARA authoring.


pjanik (approved):
Looks good 👍

PR #419, opened 2023-03-17
fix: Update hide read aloud logic so sequence always overrides [PT-184688393]

dougmartin:
Before the activity in a sequence could hide the read aloud even if the sequence was set to show the read aloud controls.  Now, if there is a sequence only that setting is used.


lublagg (approved):
Looks good :)

PR #420, opened 2023-03-24
bug: Fixed canceling read aloud on iOS when navigating away [PT-184631473]

dougmartin:
Added canceling of read aloud on pagehide as Safari on iOS doesn't handle beforeunload well.


lublagg (approved):
Looks good to me :)

pjanik (approved):
Looks good 👍

PR #421, opened 2023-03-29
bug: Fixed read aloud toggle not appearing on single page activities [PT-184823019]

dougmartin:
Added read aloud toggle to single page activity content.  Since the toggle needs to be right aligned but may be hidden an enclosing div to align it was not used, instead a style prop was added to left align the toggle itself in this instance.


pjanik (approved):
Looks good 👍

PR #422, opened 2023-04-11
feat: Converted all font sizes to use rems with pxToRem helper [PT-184910439]

dougmartin:
This is the base work to enable large type.

*NOTE*: there is at least one font size issue found when AP was tested at 20px but since then a 22px large font size was selected so this should be merged as is and follow on PRs used to update padding and container sizes once it is available for testing on staging.


scytacki (approved):
Looks good to me.

PR #424, opened 2023-04-21
feat: Added accessibility to initInteractive [PT-184981017]

dougmartin:
This reads the font size setting from either the sequence or activity and then creates an accessibility object that is passed to all iframe interactives as part of the init message.


scytacki (approved):
Looks good to me.

PR #425, opened 2023-04-21
fix: Removed width for estimated time to complete [PT-184981592]

dougmartin:
This also adds estimated times to complete to two of the sample activities so that the normal vs. large font can be seen.

Note: the estimated time display in the sequence homepage was already set to auto to override the fixed width so that was also removed.


scytacki (approved):
seems fine to me

PR #428, opened 2023-04-25
fix: Added font_size to legacy resource import

dougmartin:
This is missing for the sequences so the large font sequence sample was not rendering correctly as the font_size was not being passed in.


dougmartin:
Merging without review for demo today.

PR #429, opened 2023-05-02
fix: Changed sequence thumbs to use min-height to avoid cropping images in large fonts [PT-185079448]

dougmartin:
Also fixed layout issues by switching to using `gap` instead of margins.

Before:

image

After: 

image


scytacki (approved):
This seems OK.

I think the thumbnail is used in a few places though. So by cropping it we were forcing authors to use a specific size so that it would fit. I don't remember the details though, so I'm fine with letting go of that nuance.

PR #430, opened 2023-05-02
fix: Activity summary title alignment [PT-185079432]

dougmartin:
Added `align-items: center` to `.activity-title` to ensure the thumbnail image vertically aligns with the activity title text when in large font mode.


scytacki (approved):
seems fine to me.

PR #431, opened 2023-05-03
fix: Font size issues found during QA [PT-184981533]

dougmartin:
- Removed set of body font size so only html font size is set, just like question-interactives
- Fixed modal dialog to not have fixed height so it lays out in larger font and changed modal classes to more specific names to prevent conflicts
- Added body class for font size, same as the useAccessibility() hook used in question-interactives
- Reduced the section margin in large fonts (using body class selector above) to prevent border clipping


scytacki (approved):
Seems fine.

PR #437, opened 2023-05-31
feat: Force large font size for notebook layout [PT-185296765]

dougmartin:
This forces any activity that uses the notebook layout to use the large font setting.


lublagg (approved):
Looks good

PR #438, opened 2023-05-31
fix: Updated plugin popups to use rem based font sizes [PT-184981622]

dougmartin:
This allows the glossary popup to use the large font setting if enabled.

This is what the glossary popup now looks like in the large font setting:

image


lublagg (approved):
👍

PR #439, opened 2023-05-31
fix: Add fallback for question name to empty string

dougmartin:
This was removed by accident in the notebook reskin.


lublagg (approved):
Looks good 👍

PR #441, opened 2023-06-01
fix: Large font issues for sidebar icon container [PT-184981622]

dougmartin:
Changed to rem based units from px based units for sidebar container icon so that it changes dynamically with the font size setting.


lublagg (approved):
👍

PR #444, opened 2023-06-07
Fixes to layout and arrows when using collapsible 70-30 or 60-40 layout. (PT-185354960)

lublagg:
This PR fixes two issues:

- When the collapsible header was on the right side of the layout, the wrong arrows were showing for Show and Hide modes.
- When the collapsible header was on the right side of the layout, the collapsed column was taking up too much space.


dougmartin (approved):
Looks good :+1:

PR #446, opened 2023-06-08
fix: Hidden sections are no longer rendered in notebook layout [PT-185314409]

dougmartin:
Note: this also fixes a check for hidden sections with responsive layouts.


lublagg (approved):
Looks good 👍

PR #449, opened 2023-06-13
chore: Remove forced notebook layout [PT-185302905]

dougmartin:
This was initially added to make dev/testing/demoing easier but is no longer needed as there is an authoring setting to set the notebook layout now.


lublagg (approved):
👍

PR #451, opened 2023-06-13
fix: Lab book interactive only displays properly on Tab 1 of the Notebook layout [PT-185055230]

dougmartin:
This changes hidden tabs to render with visibility hidden with a height of 0 to ensure the layout with the correct width at startup.

In order to do this the collapsible header transition needs to be disabled in the notebook layout so it doesn't transition in when the section visibility changes when a tab is selected.


lublagg (approved):
Looks good.

PR #452, opened 2023-06-13
fix: Missing $ in string interpolation when generating question headers [PT-185302905]

dougmartin:
This was introduced in the notebook spike and wasn't seen until the forcing of the activity layout to the notebook was removed in the notebook feature branch.


lublagg (approved):
Looks good!

PR #453, opened 2023-06-13
fix: Added space after colon in question headers [PT-185302905]

dougmartin:
This also removes the colon when there is no question name to match the current master.


lublagg (approved):
👍

PR #455, opened 2023-06-15
feat: Added fontType to accessiblility [PT-185336244]

dougmartin:
This adds the fontType to the accessibility object in the init interactive message.

It also goes ahead and adds the new section layout override check for the notebook font (but not the layout itself as that is another PT story).


lublagg (approved):
Looks good :)

PR #456, opened 2023-06-16
Give iframe top border if there is no question header. (PT-185395490)

lublagg:
*NOTE FROM DOUG*: I merged the `notebook` branch back in as a branch merged before this updated the package-lock.json format to v2.  I then had to downgrade the build to use node 14 to use a lower npm version that doesn't error out on a peer dependency check for React between AP and the dynamic-text package.  Upgrading back to 16 will be addressed later this week.


src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -102,6 +103,7 @@ export const IframeRuntime: React.ForwardRefExoticComponent<IProps> = forwardRef
     resolveAndCleanup: useRef<() => void>(),
   };
   const currentInteractiveState = useRef<any>(initialInteractiveState);
+  const hasHeaderClass = hasHeader ? "iframe-has-header" : "iframe";
```
pjanik, line 106: 1. hasHeader is a number, but clearly used as a boolean. Note that if `quesitonNumber` is 0, it might not work as expected (it might never not equal to 0, but who knows where do we increase index to start from 1).
2. "iframe" class doesn't seem to be used anywhere.
3. Note that we have a `classname` lib in AP. You could use it like `classnames({ "has-header": hasHeader });`.

pjanik (changes requested):
Generally, it's a small PR and looks good, but it'd be nice to clean up a few details.

Also, I'd appreciate some PR descriptions. Even if it's that small, if you review it out of context, it isn't easy to guess what happens and why we fix it that way. Sometimes it's easier to just read the PT story/bug description than rewrite it in PR description, but in this case, a PT story link in the PR description would be great too. Or if there are plenty of comments in the code, then we could also skip it.

While it's pretty logical that we need a header when there's a question number, it's a bit unclear to me why this logic was added just now and how it worked before. Not a blocker, I trust it's necessary, but just to explain what I mean. :wink:

src/components/activity-page/managed-interactive/iframe-runtime.tsx
```diff
@@ -81,13 +81,14 @@ interface IProps {
   showDeleteDataButton?: boolean;
   setAspectRatio: (aspectRatio: number) => void;
   setHeightFromInteractive: (heightFromInteractive: number) => void;
+  hasHeader: number|undefined;
```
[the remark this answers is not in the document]
dougmartin: Fixed in https://github.com/concord-consortium/activity-player/pull/456/commits/53deeea5f1ee17a543e67021592b9293620da51b.

src/components/activity-page/managed-interactive/iframe-runtime.tsx
dougmartin: Fix in follow on PRs.

pjanik (approved):
Thanks, looks good 👍

PR #458, opened 2023-06-20
fix: Border radius for iframes with no question header [PT-185395490]

dougmartin:
This keeps the original bottom border radius when the question has a header and sets the default border radius to be on all four corners.


lublagg (approved):
Looks good 👍

PR #460, opened 2023-06-20
feat: Add sequence layout for notebook [PT#185073491]

dougmartin:


src/notebook.scss
```diff
@@ -3,6 +3,8 @@
 $notebook_font_family: 'Andika';
 $notebook_font_size: 1rem;  // this uses the new font size setting, forced to "large" in notebooks
 
+$notebookHeaderZIndex: 100;
```
lublagg, line 6: If it's alright, let's change the noteBookHeaderZIndex to less than 100. The update to the LARA Sharing Plugin gives the plugin a z-index of 100, and we want it to be visible over the header.

lublagg (changes requested):

src/notebook.scss
dougmartin: @lublagg No problem - changed in https://github.com/concord-consortium/activity-player/pull/460/commits/de3e6d73855a36e162b2ae2a45dcf306016cb74b.

lublagg (approved):
Looks good!

PR #461, opened 2023-06-20
Fix overlapping arrows with text and check for responsive 2 column layout (PT-185354960)

lublagg:
This addresses an issue where when the activity font is set to large, the text and collapse/show arrows were colliding.
It was tested in both cases of text being set to normal and large, and there are no regressions if text is set to normal.
Also added a check for responsive 2 column layout -- the column is on the left in that case, and the proper arrows need to be displayed.


dougmartin:
@lublagg You have some failing Cypress tests for this PR.

dougmartin (approved):
Looks good :+1:

PR #462, opened 2023-06-20
fix: Sidebars not displaying in notebook layout [PT-185395591]

dougmartin:
This was code left over from the spike that was not removed.


lublagg (approved):
Looks good

PR #463, opened 2023-06-20
Fix tabs and notebook header z-index issues. (PT-185395585)

lublagg:
This PR changes the section tabs to have the same z-index as the notebook header (99) so that they don't appear over the sharing plugin, which has a z-index of 100.

It also adds an absolute position to the expandable-container div, because otherwise the stacking order was interfering with the correct application of the z-indices.


dougmartin (approved):
Looks good :+1:

PR #466, opened 2023-06-25
fix: Added key to renderActivity() [PT-185073491]

dougmartin:
With no key React sometimes does not update the activity homepage text when activities switch within sequences.


lublagg (approved):
Nice catch :)

PR #468, opened 2023-06-26
fix: Updated plugin rendering to use notebook font [PT-185471695]

dougmartin:
This removes the last of the specific Lato overrides in AP and refactors the font family in the notebook.scss to use a sass variable.


lublagg (approved):
Looks good!

PR #469, opened 2023-06-27
fix: Update logic for rendering top and bottom activity nav [PT-185316649]

dougmartin:
This cleans up the logic for rendering the bottom and top nav elements.


lublagg (approved):
Nice, looks good 👍

PR #470, opened 2023-06-28
fix: Re-enable Cypress tests on notebook branch [PT-185302126]

dougmartin:
This PR makes some changes to the tests to look for components with new class names.
I'm not sure why the Scorebot classnames were different from the ones that were previously in the tests, but they were, so I updated them.


pjanik (approved):
Looks good, thanks. 👍

PR #471, opened 2023-06-28
fix: Digging deeper label in notebook layout [PT-185471695]

dougmartin:
This fixes the font family for the digging deeper glossary label in notebook layout


lublagg (approved):
Looks good 👍

PR #473, opened 2023-07-10
Updated dependencies and build settings

dougmartin:
- Reset build_test to node 16 in ci.yml
- Reset package.json engines back to node 16, npm 8
- Updated @concord-consortium/dynamic-text to 1.0.4
- Updated @concord-consortium/lara-interactive-api to 1.9.4


pjanik (approved):
Looks good. 👍 

I've just pushed one commit that updates dynamic-text to v1.0.5 so we can have one less dependency conflict (v1.0.5 updated its required react version). But there are many, many more... So, --legacy-peer-deps/force is still necessary.

PR #476, opened 2023-08-24
SPIKE (DO NOT MERGE): Hide question numbers [PT-185893147]

dougmartin:
This spike does the following:

1. ALWAYS hides the question number in the interactive iframe header
2. IF "hideHeaderIfEmpty" is somewhere in the query parameters (no matter the value) AND the question has no name or hint the header itself is hidden so that it isn't a blank bar.
3. Disables the Cypress tests as they test the existence of the question name


dougmartin:
Closing - only needed for testing out functionality.

PR #478, opened 2023-10-27
185400438 notebook footer space

eireland:
Adds some css that removes the extra space below sections in notebook layout with tabs


src/notebook.scss
```diff
@@ -239,6 +239,9 @@ body.notebook {
           height: 0;
           margin: 0 !important;
           padding: 0 !important;
+          position: relative;
```
dougmartin, line 242: This is a nice pragmatic solution!  I think it might also be good to set it to a larger negative number like `-10000px` and also set `opacity: 0` in the OFF chance that someone creates a extremely long tab - that way it won't overlap.

dougmartin (changes requested):
Please try out the `-10000px` change and add `opacity: 0` to `.hidden-tab`.

dougmartin (approved):
Looks good :+1:

lublagg (approved):
LGTM 👍

PR #481, opened 2024-01-05
fix: Hide question numbers QA issues [PT-186748231]

dougmartin:
- Added back showing header even when question numbers are disabled and the question has no name when the activity is in notebook layout, the question has a hint or the question has a plugin.
- Fixed question numbers not being hidden in the single page layout


eireland (approved):
👍 LGTM

PR #490, opened 2024-09-16
build: Disable ScoreBot Cypress tests

dougmartin:
The ScoreBot tests are broken because the ScoreBot api endpoint is not sending correct responses.  These tests should be fixed if and when ScoreBot is fixed.

The Image Question tests are broken on GitHub but run fine locally.  This is probably due to the GitHub Actions platform updating some dependency.  These tests should be fixed the next time AP is worked on.


scytacki (approved):
Looks good to me.

PR #496, opened 2024-10-30
fix: Added missing inherit_hide_question_number in hide question number sample

dougmartin:


nstclair-cc:
Nice! I was wondering what was going on in the hide question number sample activity. I'll go ahead and do the merge. Thanks, @dougmartin!

nstclair-cc (approved):

PR #499, opened 2025-01-17
feat: gather teacher feedback (PT-188673566)

emcelroy:
[#188673566](https://www.pivotaltracker.com/story/show/188673566)

### DIFF
--- a/src/components/activity-introduction/introduction-page-content.tsx
+++ b/src/components/activity-introduction/introduction-page-content.tsx
@@ -9,20 +9,18 @@ import "./introduction-page-content.scss";
 
 interface IProps {
   activity: Activity;
-  isInSequence?: boolean;
   onPageChange: (page: number) => void;
 }
 
 export const IntroductionPageContent: React.FC<IProps> = (props) => {
-  const { activity, onPageChange, isInSequence } = props;
+  const { activity, onPageChange } = props;
   const [feedback, setFeedback] = useState<TeacherFeedback | null>(null);
 
   useEffect(() => {
-    if (!activity.id) return;
     const unsubscribe = watchActivityLevelFeedback(fb => setFeedback(fb));
 
     return () => unsubscribe();
-  }, [activity.id, isInSequence]);
+  }, []);
 
   return (
     <div className="intro-content" data-cy="intro-page-content">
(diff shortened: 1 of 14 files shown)


src/components/activity-introduction/introduction-page-content.tsx
```diff
@@ -1,17 +1,27 @@
...
 }
 
 export const IntroductionPageContent: React.FC<IProps> = (props) => {
-  const { activity, onPageChange } = props;
+  const { activity, onPageChange, isInSequence } = props;
+  const [feedback, setFeedback] = useState<TeacherFeedback | null>(null);
+
+  useEffect(() => {
+    if (activity.id) {
+      return watchActivityLevelFeedback(fb => setFeedback(fb));
```
dougmartin, line 22: can we return a function here to stop watching activity level feedback when the page unmounts?

src/components/activity-page/managed-interactive/managed-interactive.tsx
```diff
@@ -388,6 +389,18 @@ export const ManagedInteractive: React.ForwardRefExoticComponent<IProps> = forwa
...
+    async function watchFeedback() {
+      const answer = await getAnswerMetadata(interactiveId);
+      if (answer?.id) {
+        return watchQuestionLevelFeedback(answer.id, (fb) => {
+          setFeedback(fb);
+        });
+      }
+    }
+    watchFeedback();
+  }, [interactiveId]);
```
dougmartin, line 402: Same question here - can we return a function that "unwatches" the feedback when it unmounts?

src/firebase-db.ts
```diff
@@ -271,6 +282,121 @@ export const watchAllAnswers = (callback: (wrappedAnswer: WrappedDBAnswer[]) =>
...
+    throw new Error("Must set portal data first");
+  }
+  // Only logged-in students will have feedback.
+  if (portalData.userType !== "learner" || portalData.type === "anonymous") return;
+
+  let query: firebase.firestore.Query = app.firestore().collection(teacherFeedbackPath("question"));
+
+  if (portalData.type === "authenticated") {
+    query = query
+      .where("answerId", "==", answerId)
```
dougmartin, line 316: Did you need to leave in the answerId?

dougmartin (changes requested):
I think we need to unmount watching feedback in the two components I commented on, unless that is taken care of elsewhere.

src/firebase-db.ts
emcelroy: No. Thanks for catching that. It's gone now.

emcelroy:
> I think we need to unmount watching feedback in the two components I commented on, unless that is taken care of elsewhere.

@dougmartin What do you think of the latest changes for that?

dougmartin (approved):

dougmartin:
@emcelroy / @lublagg the unsubscribe changes look good!

PR #500, opened 2025-01-21
Sequence Homepage Feedback (PT-188732083)

lublagg:
Makes the following changes:
- Sequence homepage displays feedback banner below the Time to Complete banner when feedback is present on one or more activities at either the activity level or question level (or both)
- Sequence homepage displays feedback badges on Activity buttons when feedback is present on those activities at either the activity-level or question-level (or both)


dougmartin (approved):
Looks good :+1:

PR #502, opened 2025-01-22
feat: activity homepage displays activity-level feedback (PT-188732078)

emcelroy:
[#188732078](https://www.pivotaltracker.com/story/show/188732078)

When there is activity-level feedback, a banner is displayed at the bottom of the activity home page if the activity does not have a completion page. If it does have a completion page, the activity-level feedback will be displayed on the completion page instead. Adding feedback to the completion page is a [separate story](https://www.pivotaltracker.com/story/show/188732086), but both feedback-display elements will share the new `ActivityLevelFeedbackBanner` component.

Also, we decided to consolidate all feedback-related component files into a new `teacher-feedback` directory.


dougmartin (approved):
Looks good :+1:

PR #503, opened 2025-01-22
fix: gather teacher feedback

emcelroy:
The `answerId` is actually required in `getQuestionLevelFeedbackDocsQuery`. Otherwise, feedback for any question in an activity will be applied to all questions in that activity.


dougmartin (approved):
Looks good :+1:

lublagg (approved):
LGTM 👍

PR #507, opened 2025-01-24
fix: page nav & TOC feedback badges (PT-188732073)

emcelroy:
[188732073](https://www.pivotaltracker.com/story/show/188732073)

This should fix two problems identified during testing:

1) The home page navbar button wasn't displaying a feedback badge when appropriate. It should display one when there is no completion page, and there is activity-level feedback or at least one question-level feedback.

2) Question-level feedback wasn't properly being identified in standalone activities because we were only building the `questionToActivityMap` for sequences.

### DIFF
--- a/src/components/activity-header/activity-nav.tsx
+++ b/src/components/activity-header/activity-nav.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Page, QuestionToActivityMap } from "../../types";
+import { Page, QuestionMap } from "../../types";
 import { NavPages } from "./nav-pages";
 
 import "./activity-nav.scss";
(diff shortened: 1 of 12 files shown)


emcelroy:
> 2. Question-level feedback wasn't properly being identified in standalone activities because we were only building the `questionToActivityMap` for sequences.

@dougmartin @lublagg I just realized how odd it sounds to build a `questionToActivityMap` for a single activity. Maybe we should rename it `questionToPageMap` and it'd be understood that activity info is included? 

Or maybe simply `questionMap`?

dougmartin (approved):
Looks good :+1:

dougmartin:
> > 2. Question-level feedback wasn't properly being identified in standalone activities because we were only building the `questionToActivityMap` for sequences.
> 
> @dougmartin @lublagg I just realized how odd it sounds to build a `questionToActivityMap` for a single activity. Maybe we should rename it `questionToPageMap` and it'd be understood that activity info is included?
> 
> Or maybe simply `questionMap`?

I think `questionMap` is a good name.

PR #509, opened 2025-01-27
feat: Page question clickable link in summary table (PT-188732105)

lublagg:
[#188732105](https://www.pivotaltracker.com/story/show/188732105)

Question and page number text in the summary table is now a clickable link. Clicking a link takes you to the target page and scrolls the target question into view. Link clicks are logged to student action report in addition to triggering a page-change log message. The latter is an existing part of the `handleChangePage` function in `App` that we modified for this feature.

These changes include some refactoring:

- Adds a new context for `questionMap` and removes related props that are no longer needed. The context is also used for the new `scrollToQuestionId`.
- Reorganizes `ManagedInteractive` a bit, and moves some pieces into the new `ManagedInteractiveHeader` and `ManagedInteractiveHint` components and a new `useSizeAndAspectRatio` hook.


dougmartin (approved):
Looks good :+1: 

I like the new question context that cleans up a lot of the prop drilling.

PR #512, opened 2025-01-30
feat: Show teacher feedback button (PT-188732115)

lublagg:
[#188732115](https://www.pivotaltracker.com/story/show/188732115)

- When feedback is present on a question, the "Show Teacher Feedback" button appears at the bottom of the container.
- Clicking button reveals feedback comment. 
- If the teacher has updated their feedback, this button returns to its default state (comment is again hidden).
- State of the feedback does not need to be saved, i.e., button returns on page reload.
- Button clicks should be logged in the student actions report


dougmartin (approved):
Looks good :+1:

PR #515, opened 2025-02-04
feat: support for `showFeedback` param (PT-188732128)

emcelroy:
[#188732128](https://www.pivotaltracker.com/story/show/188732128)

When the `showFeedback` URL query param is set:

- Activities with completion page will automatically load the completion page
- Activities without the completion page will automatically load the home page
- Sequences will automatically load the sequence home page


dougmartin (approved):
Looks good :+1:

PR #516, opened 2025-02-04
fix: don't show rubric when rubric hidden (PT-188780572)

emcelroy:
[#188780572](https://www.pivotaltracker.com/story/show/188780572)

Fixes an issue where rubrics were displayed to students even when the rubrics' "Hide rubric from students" option was checked.


lublagg (approved):
Looks good to me! 👍

PR #517, opened 2025-02-06
fix: summary table feedback margin (PT-188859980)

emcelroy:
[#188859980](https://www.pivotaltracker.com/story/show/188859980)

Ensures that when a question has no prompt, there is no top margin to question-level feedback in the summary table.


dougmartin (approved):
Looks good :+1:

PR #521, opened 2025-05-05
fix: Filter out hidden page questions when building question map [AP-36]

dougmartin:
Questions on hidden pages should not be included in the question map for the summary table.


emcelroy (approved):
Looks good -- thanks!

PR #523, opened 2025-05-13
fix: Out of bounds page read in setDocumentTitle [AP-38]

dougmartin:
Fixes an error when the showFeedbackPage query parameter is present.  The completion page page number was not being correctly calculated when there were hidden pages and setDocumentTitle was being passed a page number outside the bounds of the page array.

This also adds a sanity check to ensure setDocumentTitle uses a page number within the range of pages to avoid a page crash.

### DIFF
--- a/src/components/app.tsx
+++ b/src/components/app.tsx
@@ -184,6 +184,12 @@ export class App extends React.PureComponent<IProps, IState> {
     }
   }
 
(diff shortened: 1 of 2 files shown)


emcelroy (approved):
Looks good 👍 I left some comments about the new `getVisiblePages` but I don't think they should block merging this in if you think it's necessary and good as is.

PR #525, opened 2025-05-19
fix: Removes null callback parameter for feedback watchers (AP-40)

dougmartin:
This changes the null callback parameter to an empty array and updates the callback sites to remove the typing that masked that null was being passed.


dougmartin:
@emcelroy I removed your review and converted back to draft due to all the Cypress failures.  I'll work on this next sprint (it was assigned for the next one anyway)

emcelroy (approved):
Looks good -- thanks 👍

PR #527, opened 2025-07-22
fix: Changed lock check to be at student level [AP-42]

dougmartin:
This was missed when we updated the offering lock behavior to allow for locked offerings with an override at the student level.


emcelroy (approved):
Looks good 👍

PR #549, opened 2026-06-02
AP-108: Cross-origin interactive focus-trap design spec

scytacki:
## Summary

High-level design for adding focus-trap support to Activity Player overlays (dialog, lightbox) that contain cross-origin interactives, plus the shared host-side and protocol pieces this needs in `accessibility-tools`, `lara-interactive-api`, and a new `interactive-api-host` layer.

**This is a cross-repository design spec, not an implementation plan.** Per-repository implementation specs will be written separately, in each affected repo, as follow-ups.

## For the reviewer

A short ["For reviewers"](https://github.com/concord-consortium/activity-player/blob/AP-208-interactive-focus-api/specs/AP-208-focus-traps-cross-origin-interactives.md#for-reviewers) section at the top of the spec lists what to focus on and what to skim. In short:

**Please weigh in on the architectural decisions:**
- The cooperating / non-cooperating split and the protocol shape (sections: _Background: why a protocol is unavoidable_, _The focus-coordination protocol_, _Path 1_, _Path 2_).
- The library layering across repos and what lives where (sections: _Architecture: layers_, _Host code sharing across apps_, _Components by repository_).
- The focus-category model (section: _Interactive focus categories & visual focus indication_).

**Feel free to skim or trust these — they are background or deferred:**
- _Background: how focus works with iframes in browsers_ — empirical cross-browser findings (verified on macOS Chrome / Firefox / Safari) that confirm feasibility. The design only depends on the matrix in one place: `tabIndex={-1}` reliably skips the iframe.
- _Host-configured key forwarding_ — not used by AP in this version; included because the shared host layer is the right home for it (CODAP is the immediate consumer).
- _Open questions / risks_, _Out of scope_, _Future direction_ — known unknowns, scope boundaries, and explicitly deferred follow-ups.

## Test plan

This PR adds a design document only — no code changes.

- [ ] Spec renders correctly on GitHub
- [ ] Internal anchor links in the "For reviewers" section navigate to the right sections
- [ ] Reviewer approves the architectural decisions listed above, OR raises concerns to resolve before per-repo implementation specs begin


dougmartin (approved):
I read through it all and it seems very well thought out.  The one thing I didn't see was click handling both to focus into and out of an iframe into another iframe or into AP.  Do you see this library handing a user clicking into an iframe and then tabbing around and then clicking into another iframe on the page or clicking on AP host UI?

scytacki:
@dougmartin I don't think that we have to do anything special to handle clicks. In all browsers but Safari, the clicked on element will be focused by the browser. And in Safari it has some internal state that remembers it is the focused element (not window.activeElement).

So then tabbing afterward will goes to the next element in native tab order. Or if tab is being handled by a11y-tools then it would find the current focused element, get the list of other focusable elements and go to the next one. The Safari behavior will break a11y-tools tab handling though. But this is an issue independent of whether iframes are being used or not.

You can experiment with this on the test site: https://models-resources.concord.org/demos/branch/iframe-focus/

PR #562, opened 2026-06-25
AP-116: Host-side cooperating focus protocol

scytacki:
## Summary

Wires the **host half of the cooperating focus protocol** so an interactive that speaks the protocol (advertises `focusProtocol: true`) can coordinate focus with the Activity Player in the **dialog** modal — the host sends `focusEnter`, the interactive sends `focusExit`. Interactives that don't speak the protocol are unchanged (AP-110 sentinel/native fallback).

Jira: https://concord-consortium.atlassian.net/browse/AP-116
Design: `specs/AP-116-cooperating-focus-protocol.md` (builds on AP-108 / AP-110).

## What's here

- **Build & surface the transport** — `iframe-runtime` constructs a `FocusManager` alongside the iframe-phone endpoint and surfaces `focusManager.transport` via an optional `onFocusTransportReady` callback. Built unconditionally; only the dialog consumes it.
- **Consume it in the dialog** — `dialog-overlay` holds the transport in state and passes it into `useIframeSlot`, so the cooperative flow engages when the interactive cooperates.
- **Forward the `focusProtocol` capability to the FocusManager** — iframe-phone delivers a single listener per message type and `iframe-runtime` already owns `"supportedFeatures"`, so the FocusManager can't register its own without clobbering. `iframe-runtime` now forwards the capability via `FocusManager.notifyCapability()` from its existing listener.
- **Library upgrade** — uses the published pre-release libs that ship the protocol (accessibility-tools `0.2.0-pre.3` with late-`transport` subscription + late-capability upgrade; interactive-api-host `0.12.0-pre.0` with `FocusManager.notifyCapability`; lara-interactive-api `1.14.0-pre.0`).

## Manual Testing

This link uses an updated question-interactives with an image-question that supports the focus protocol:
https://activity-player.concord.org/branch/cooperating-focus-protocol/index.html?activity=sample-activity-all-question-interactives&override.qi=image-focus-protocol

Page 9 has the image question on it. You should notice that focus lands on the toolbar in the drawing tool immediately when the "make drawing" button is clicked via a keyboard or mouse.

You can compare it with the master version of the question-interactives with this link:
https://activity-player.concord.org/branch/master/index.html?activity=sample-activity-all-question-interactives
In this case the image-question is not "cooperating" so the AP shows a "Press tab to enter the interactive" hint.

### Punted issues

- even in the cooperating case the hint flashes quickly and then disappears
- when the mouse is used to open the dialog by clicking on the "make drawing" button the hint should not be shown. The hint is specific to keyboard users, so it should only be shown when the keyboard is used to open the dialog.
- there shouldn't be a focus ring on the drawing toolbar when the mouse is used to open the cooperating dialog

These issues are punted because I've run out of time. The solution to (2) and (3) requires a new option in the openDialog interactive api. This way the interactive can tell its host if the mouse or keyboard was used to open the dialog. Support for hiding the hint is provided in the accessibility-tools library that the AP is using to manage all of this.

The solution to (1) is less clear. It happens because the iframe takes some time to load and let the AP know that it supports the focus protocol. So the AP starts by just showing the hint and then when the interactive says it supports the focus protocol it hides the hint. One approach would be to add some delay before showing the hint to give more time to the interactive to load. Another would be for the openDialog api dialog to include a flag indicating that the interactive being opened supports the focusProtocol so the host (AP) should wait and not show the hint.

## Testing

- Unit tests for transport surfacing, FocusManager teardown, and capability forwarding (`iframe-runtime.test.tsx`, `managed-interactive.test.tsx`).
- Manually verified end-to-end against the lara `focus-interactive-coop` example interactive: opening a (no-close) dialog hands focus cooperatively into the interactive with no sentinel hint; the non-cooperating fallback and closeable-dialog focus are unchanged.

## Notes

- `master` is merged into this branch.
- **Not** included: `focusExit("escape")` → close-the-dialog wiring (the dialog still passes no `onRequestExit`); a one-line follow-up when desired.
- No bespoke localhost sample activity is added; cooperating-path manual testing uses the lara example interactive for now.


kswenson (approved):
👍 Looks good -- the following review was developed in conjunction with Claude Code 🤖.

### What it does

Wires the **host half** of a cooperating focus protocol. An interactive that advertises `focusProtocol: true` can now coordinate focus with the Activity Player inside the **dialog** modal: the host sends `focusEnter`, the interactive replies `focusExit`. Interactives that don't speak the protocol are unchanged — they keep the AP-110 sentinel/native-fallback hint. Builds on AP-108/AP-110.

### The approach

"Option A — build the transport everywhere, consume it only in the dialog," with zero new conditional branches:

- **`iframe-runtime.tsx`** — beside the existing `new iframePhone.ParentEndpoint(...)` in the `[reloadCount, url]` effect, constructs `new FocusManager(phone)`, stores it in a ref, and surfaces `focusManager.transport` via a new optional `onFocusTransportReady?` prop. Cleanup calls `destroy()` and re-fires the callback with `undefined`. Because iframe-phone allows only one listener per message and iframe-runtime already owns the `"supportedFeatures"` listener, it forwards the capability via `notifyCapability(!!features.focusProtocol)` from that existing listener instead of letting FocusManager register a competing one.
- **`dialog-overlay.tsx`** — holds the transport in `useState`, passes `onFocusTransportReady={setTransport}` into `IframeRuntime`, and threads `transport` into the existing `useIframeSlot({...})` call.
- **`managed-interactive.tsx`** (inline) — render untouched; transport built but unused (FocusManager is passive).
- **`types.ts`** — adds `removeListener` to the `IframePhone` interface.
- **Library upgrade** — accessibility-tools `0.2.0-pre.3`, interactive-api-host `0.12.0-pre.1`, lara-interactive-api `1.14.0-pre.0`.

### Assessment

Clean, minimal, well-reasoned integration. Verified locally:

- **Runtime API matches.** `interactive-api-host@0.12.0-pre.1` `FocusManager` exposes exactly `transport`, `notifyCapability(boolean)`, and `destroy()` as used; the library's own d.ts doc-comments independently confirm the one-listener-per-message rationale behind `notifyCapability`.
- **`removeListener` is real.** The actual iframe-phone `ParentEndpoint` exposes `removeListener`, so `FocusManager.destroy()` is safe — not just a type assertion.
- **Lifecycle is correct.** FocusManager built per phone build, destroyed before `phone.disconnect()` in cleanup; rebuilt and re-surfaced on reload/url change. Late-subscribe ordering is handled by the library caching/replaying the last capability.
- **Fallback unchanged.** Existing AP-110 `dialog-overlay.test.tsx` still passes.
- **Tests:** all 21 tests across the 4 changed/added suites pass; `src` tsc clean; lint clean on changed sources; full CI (Build+Jest, all Cypress shards) green.

### Issues

No blocking concerns. Punted items are explicitly documented in the PR description and reasonably deferred (they need a new openDialog API option on the interactive side).

Minor (non-blocking): `notifyCapability(!!features.focusProtocol)` fires on every `supportedFeatures` message, re-emitting `capability:false` when absent. Harmless (library caches/replays), just noting it.

scytacki:
Thanks for the review. 

I'm going to leave this one "ready to merge" until the other PRs have also been approved. Then hopefully I can remove all of the pre-release versions before merging.
Other PRs:
- https://github.com/concord-consortium/accessibility-tools/pull/7
- https://github.com/concord-consortium/question-interactives/pull/510
- https://github.com/concord-consortium/lara/pull/1260

PR #571, opened 2026-06-29
fix(AP-98,100,101,102): accessible "Did you know?" sidebar (semantic buttons + contrast)

kswenson:
## Summary

Accessibility fixes for the "Did you know?" page sidebar (the orange tab + slide-out panel), addressing four VPAT audit stories under epic AP-81. The sidebar is a single shared component, so these fixes apply on every page where it appears.

Resolves **AP-98**, **AP-100**, **AP-101**, **AP-102**. (AP-99 — focus trap / dialog semantics — is intentionally a separate follow-up PR; the audit checklist is captured on that ticket.)

## Changes by story

- **AP-98** — Tab trigger converted from `<div role="button" tabIndex={0}>` to a native `<button type="button">` with `aria-haspopup="dialog"` and `aria-expanded`. Native Enter/Space activation replaces the manual keydown handling. Added an `aria-label="Show sidebar"` fallback so the button is never unnamed when a sidebar has no title. _(WCAG 4.1.2 A)_
- **AP-100** — Header title text `white` → `$cc-charcoal` (#3f3f3f) on the #ffa350 header: **5.32:1**, clears the 4.5:1 AA threshold. _(WCAG 1.4.3 AA)_
- **AP-101** — Close-icon fill `white` → `$cc-charcoal` (#3f3f3f): **5.32:1** default, and the icon now keeps its charcoal fill through `:hover` (8.77:1) and `:active` (6.64:1) instead of fading to light orange. _(WCAG 1.4.11 AA)_
- **AP-102** — Close ("x") control converted from `<div role="button">` to a native `<button type="button">` with `aria-label="Close"`. _(WCAG 4.1.2 A)_

Both converted buttons also get a theme-independent focus indicator (solid `$cc-charcoal` outline, matching `.version-info` in `app.scss`) — they sit on the orange theme, exactly the case where the themed box-shadow ring can drop below contrast, and the charcoal outline survives forced-colors mode. _(WCAG 2.4.7 AA)_

Native button chrome is reset in SCSS (`border`/`background`/`padding`, leaving `outline` intact) so the controls render identically to the former divs. `data-cy` hooks are preserved, so Cypress selectors/interactions are unaffected.

## Testing

- Added unit tests (test-first) covering the new semantic buttons, `aria-haspopup`/`aria-expanded`, the close button's accessible name, and the title-fallback name.
- Full Jest suite green; lint clean.
- All contrast ratios verified with a WCAG contrast analyzer.
- A read-only accessibility audit of the branch confirmed all four stories' acceptance criteria are met (0 critical / 0 high findings).

## Out of scope (follow-ups)

- Tab arrow-icon contrast (1.65:1, pre-existing) — to be ticketed under AP-81.
- AP-99: expose the panel as a real `role="dialog"` with focus trap / return-focus — separate PR; checklist captured on the ticket.

### DIFF
--- a/src/components/page-sidebar/sidebar-tab.tsx
+++ b/src/components/page-sidebar/sidebar-tab.tsx
@@ -18,22 +18,21 @@ export class SidebarTab extends React.PureComponent<IProps>{
 
   render() {
     return (
-      <div className="sidebar-tab" onClick={this.handleSidebarShow} onKeyDown={this.handleSidebarShow}
-           data-cy="sidebar-tab" tabIndex={0} role="button" aria-expanded={this.props.sidebarOpen}>
+      // A native <button> gives the trigger its semantics and Enter/Space activation for free.
+      // aria-haspopup="dialog" announces that it opens the sidebar panel; aria-expanded reflects
+      // the open state. The visible "tab-name" text supplies the button's accessible name.
+      <button type="button" className="sidebar-tab" onClick={this.handleSidebarShow}
+           data-cy="sidebar-tab" aria-haspopup="dialog" aria-expanded={this.props.sidebarOpen}>
         <div className={`icon ${this.props.sidebarOpen ? "open" : ""}`}>
           <IconArrow aria-hidden="true" focusable="false" />
         </div>
         <div className="tab-name" data-cy="sidebar-tab-title">{this.props.title}</div>
-      </div>
+      </button>
     );
   }
 
-  private handleSidebarShow = (e: React.MouseEvent | React.KeyboardEvent) => {
+  private handleSidebarShow = (e: React.MouseEvent<HTMLButtonElement>) => {
     if (accessibilityClick(e)) {
-      // role="button" div: stop Space from also scrolling the page on keyboard activation
-      if (e.type === "keydown") {
-        e.preventDefault();
-      }
       this.props.handleShowSidebarContent(this.props.index, !this.props.sidebarOpen);
     }
   }
--- a/src/components/page-sidebar/sidebar-tab.tsx
+++ b/src/components/page-sidebar/sidebar-tab.tsx
@@ -27,10 +27,12 @@ export class SidebarTab extends React.PureComponent<IProps>{
       <button type="button" className="sidebar-tab" onClick={this.handleSidebarShow}
            data-cy="sidebar-tab" aria-haspopup="dialog" aria-expanded={this.props.sidebarOpen}
            aria-label={hasTitle ? undefined : "Show sidebar"}>
-        <div className={`icon ${this.props.sidebarOpen ? "open" : ""}`}>
+        {/* A <button> may only contain phrasing content, so these are <span>s, not <div>s;
+            the parent's display:flex blockifies them, so layout is unchanged. */}
+        <span className={`icon ${this.props.sidebarOpen ? "open" : ""}`}>
           <IconArrow aria-hidden="true" focusable="false" />
-        </div>
-        <div className="tab-name" data-cy="sidebar-tab-title">{this.props.title}</div>
+        </span>
+        <span className="tab-name" data-cy="sidebar-tab-title">{this.props.title}</span>
       </button>
     );
   }
--- a/src/components/page-sidebar/sidebar-tab.tsx
+++ b/src/components/page-sidebar/sidebar-tab.tsx
@@ -17,12 +17,16 @@ export class SidebarTab extends React.PureComponent<IProps>{
   }
 
   render() {
+    // The visible "tab-name" text is the button's accessible name when a title exists; fall back
+    // to a static aria-label only when there is none, so the button is never unnamed.
+    const hasTitle = !!this.props.title?.trim();
     return (
       // A native <button> gives the trigger its semantics and Enter/Space activation for free.
       // aria-haspopup="dialog" announces that it opens the sidebar panel; aria-expanded reflects
-      // the open state. The visible "tab-name" text supplies the button's accessible name.
+      // the open state.
       <button type="button" className="sidebar-tab" onClick={this.handleSidebarShow}
-           data-cy="sidebar-tab" aria-haspopup="dialog" aria-expanded={this.props.sidebarOpen}>
+           data-cy="sidebar-tab" aria-haspopup="dialog" aria-expanded={this.props.sidebarOpen}
+           aria-label={hasTitle ? undefined : "Show sidebar"}>
         <div className={`icon ${this.props.sidebarOpen ? "open" : ""}`}>
           <IconArrow aria-hidden="true" focusable="false" />
         </div>
--- a/src/components/page-sidebar/sidebar.test.tsx
+++ b/src/components/page-sidebar/sidebar.test.tsx
@@ -31,6 +31,21 @@ describe("SidebarTab component", () => {
     expect(wrapper.find('[data-cy="sidebar-title"]').text()).toContain(title);
     expect(wrapper.find('[data-cy="sidebar-content"]').length).toBe(1);
   });
+  it("renders the close control as a semantic button with an accessible name", () => {
+    const stubFunction = () => {
+      // do nothing.
+    };
+    const wrapper = shallow(<SidebarPanel
+      title={"Did you know?"}
+      index={0}
+      content={"content"}
+      show={true}
+      handleCloseSidebarContent={stubFunction} />);
+    const closeButton = wrapper.find('button[data-cy="sidebar-close-button"]');
+    expect(closeButton.length).toBe(1);
+    expect(closeButton.prop("type")).toBe("button");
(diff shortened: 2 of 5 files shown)


src/components/page-sidebar/sidebar-tab.tsx
```diff
@@ -17,23 +17,26 @@ export class SidebarTab extends React.PureComponent<IProps>{
...
+    const hasTitle = !!this.props.title?.trim();
     return (
-      <div className="sidebar-tab" onClick={this.handleSidebarShow} onKeyDown={this.handleSidebarShow}
-           data-cy="sidebar-tab" tabIndex={0} role="button" aria-expanded={this.props.sidebarOpen}>
+      // A native <button> gives the trigger its semantics and Enter/Space activation for free.
+      // aria-haspopup="dialog" announces that it opens the sidebar panel; aria-expanded reflects
+      // the open state.
+      <button type="button" className="sidebar-tab" onClick={this.handleSidebarShow}
+           data-cy="sidebar-tab" aria-haspopup="dialog" aria-expanded={this.props.sidebarOpen}
+           aria-label={hasTitle ? undefined : "Show sidebar"}>
```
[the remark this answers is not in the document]
kswenson: Intentional, and we discussed this exact mismatch. `aria-haspopup="dialog"` is an explicit acceptance criterion for AP-98, and the matching `role="dialog"` (plus aria-modal / focus trap / return-focus) lands in AP-99, which is in this same release (2.16.0). The trigger and the real dialog semantics ship together, so the interim mismatch never reaches production. Keeping `"dialog"` here avoids advertising a different popup type now only to change it right back in AP-99. The full AP-99 checklist is captured on that ticket.

src/components/page-sidebar/sidebar.test.tsx
```diff
@@ -16,6 +16,54 @@ describe("SidebarTab component", () => {
...
+    const wrapper = shallow(<SidebarTab
+      title={"Did you know?"}
+      handleShowSidebarContent={stubFunction}
+      index={0}
+      sidebarOpen={false} />);
+    const button = wrapper.find('button[data-cy="sidebar-tab"]');
+    expect(button.length).toBe(1);
+    expect(button.prop("type")).toBe("button");
+    expect(button.prop("aria-haspopup")).toBe("dialog");
+    expect(button.prop("aria-expanded")).toBe(false);
```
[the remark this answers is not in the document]
kswenson: We are keeping `aria-haspopup="dialog"` (see the reply on the tab-trigger thread — AP-99 adds the matching role in this same release), so this assertion is intentionally verifying the intended ARIA contract and stays as-is.

src/components/page-sidebar/sidebar-tab.tsx
```diff
@@ -17,23 +17,26 @@ export class SidebarTab extends React.PureComponent<IProps>{
...
         <div className={`icon ${this.props.sidebarOpen ? "open" : ""}`}>
           <IconArrow aria-hidden="true" focusable="false" />
         </div>
         <div className="tab-name" data-cy="sidebar-tab-title">{this.props.title}</div>
```
emcelroy: A `<button>` should not contain `<div>`s. It should only include "phrasing content". (See [this page's Technical Summary > Permitted Content section](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#technical_summary).) 

I think you could simply replace the `<div>`s with `<span>`s, but there might be more appropriate [phrasing content elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#phrasing_content) to use.

src/components/page-sidebar/sidebar-tab.tsx
```diff
@@ -17,23 +17,26 @@ export class SidebarTab extends React.PureComponent<IProps>{
...
+    // to a static aria-label only when there is none, so the button is never unnamed.
+    const hasTitle = !!this.props.title?.trim();
     return (
-      <div className="sidebar-tab" onClick={this.handleSidebarShow} onKeyDown={this.handleSidebarShow}
-           data-cy="sidebar-tab" tabIndex={0} role="button" aria-expanded={this.props.sidebarOpen}>
+      // A native <button> gives the trigger its semantics and Enter/Space activation for free.
+      // aria-haspopup="dialog" announces that it opens the sidebar panel; aria-expanded reflects
+      // the open state.
+      <button type="button" className="sidebar-tab" onClick={this.handleSidebarShow}
+           data-cy="sidebar-tab" aria-haspopup="dialog" aria-expanded={this.props.sidebarOpen}
```
emcelroy, line 28: Claude caught this: 

Don’t announce a dialog unless the panel is one. 

`aria-haspopup="dialog"` tells assistive tech this opens a dialog, but the paired `SidebarPanel` in this PR still renders as a plain `<div>` with no dialog role or focus-management contract. A screen reader user is told "opens a dialog," tabs forward expecting a dialog with a focus contract, and instead lands in an ordinary div. That's a legitimate WCAG 4.1.2 (Name, Role, Value) concern.

This panel is a disclosure, not a dialog. It's a non-modal help/info region that expands beside the content — the button already carries `aria-expanded={sidebarOpen}`, which is the disclosure pattern. `aria-haspopup` is really meant for menus/listboxes/dialogs that pop over content, and pairing it with `aria-expanded` sends a muddled signal. Promoting the panel to a true `role="dialog"` would mean committing to modal behavior (focus trap, Esc, focus return, backdrop) that this sliding sidebar doesn't actually have and probably shouldn't have.

Drop `aria-haspopup="dialog"` — and lean fully into disclosure semantics by wiring the button to the panel:

- Give the panel an id (e.g. `sidebar-panel-${this.props.index}`).
- Add `aria-controls={sidebar-panel-${this.props.index}}` to the button.
- Keep `aria-expanded`.

emcelroy (changes requested):
Looking good but there are a few issues that should be fixed. I left some comments on `sidebar-tab.tsx` with details, but also, the "Did you know?" tab isn't placed quite right in this new version. There's a gap between it and the viewport's right edge that isn't there on production. The padding doesn't seem to quite match either.

This is how it looks on [production](https://activity-player.concord.org/index.html?page=page_2000):
Image

Here's how it looks on [this branch](https://activity-player.concord.org/branch/did-you-know-a11y/index.html?page=page_2000).
Image

src/components/page-sidebar/sidebar-tab.tsx
kswenson: Fixed in 6f55780 — `.icon` and `.tab-name` are now `<span>`s. Since the button is `display: flex`, both children are blockified, so the visual layout is unchanged while the content model is now phrasing-only as required.

src/components/page-sidebar/sidebar-tab.tsx
kswenson: Apologies — I should have flagged in the PR description that this is intentional and resolved by the immediate follow-up, so it shouldn't block this PR.

You and Claude are right that, *as it stands in this PR alone*, `aria-haspopup="dialog"` advertises a contract (`role="dialog"` + focus management) that the panel doesn't yet honor. That contract lands in the very next PR, **#572 (AP-99)**, which is part of this same release (2.16.0). #572 gives `SidebarPanel` `role="dialog"`, `aria-modal={true}`, a focus trap, focus-on-open (lands on the heading), focus-return-to-trigger on close, and Esc-to-close. The trigger's `aria-haspopup="dialog"` and the real dialog semantics ship together, so the interim mismatch never reaches production.

So this is a deliberate disclosure-vs-dialog decision rather than an oversight: we're committing to the modal dialog pattern for this sidebar (that's what AP-99 implements), which is why I'd prefer to keep `aria-haspopup="dialog"` here rather than switch to the disclosure pattern (`aria-controls` + `aria-expanded` only) now and revert it in #572. Happy to discuss if you think the sidebar should remain a non-modal disclosure instead — that's really a question for the AP-99 design, and I'd rather settle it on #572 than diverge the two PRs.

kswenson:
Good catch on the tab placement — fixed in 6f55780.

Root cause: a native `<button>` defaults to `box-sizing: border-box`, whereas the old `<div>` was `content-box` (there's no global `* { box-sizing }` reset in this codebase). With border-box, the 5px padding was being subtracted from the 75px width, so the tab rendered ~10px narrower and pulled away from the viewport's right edge — and the padding looked off for the same reason. The button reset now includes `box-sizing: content-box`, restoring the exact pre-existing geometry.

The remaining inline threads are addressed too: the `<div>`→`<span>` content-model fix is in the same commit, and I've replied on the `aria-haspopup="dialog"` thread (short version: it's intentional and the matching `role="dialog"`/focus-trap lands in #572 / AP-99 in this same release — that one shouldn't block this PR, sorry for not making that clear up front).

emcelroy (approved):
Looks good 👍

PR #573, opened 2026-06-30
AP-122: Scope legacy global heading rules into a coherent, intentional system

kswenson:
## Background — how we got here

The global `.app h1` / `.app h2` element rules in `app.scss` were never designed as a heading system. They were originally introduced as styling for the **completion page**, and over time were quietly extended to cover headings in other contexts (intro/activity-summary, error screens, authored rich text) — accreting responsibilities one screen at a time. The result was a pair of element-level selectors that styled *every* `<h1>`/`<h2>` under `.app` by tag name, with no coherent intent behind which headings they should or shouldn't touch.

**AP-91** then introduced an actual, coherent set of semantic headings — wrapping page titles, section titles, and question titles in real `<h1>`/`<h2>` elements. But because the global rules matched by element type, every one of those new semantic headings was now *also* matched by the legacy completion-page styling, and had to fight it. Local, intentional styles were silently overridden by the global element rules, which produced visible regressions (notably the question header and the "Teacher Feedback" heading colors).

**AP-120** (#570, the branch this PR is stacked on) was a targeted fix for those color regressions, but it explicitly called out the global rules as a latent trap — any future `div`→heading conversion would hit the same wall — and deferred retiring/scoping them to this follow-up.

This PR is that follow-up: it untangles the two jobs the global rules had been conflating, so heading styling becomes intentional and local rather than a global element-type side effect.

## What this PR does

The global rules were really doing two unrelated things, so this splits them:

1. **Element-level defaults for authored rich-text content.** `renderHTML` output (text blocks, hints, footers, descriptions, related/next-step content) can contain bare `<h1>`/`<h2>` tags with no classes to hook — those legitimately need element-level defaults. The defaults are now **scoped to the authored-content containers only**, at the same `(0,1,1)` specificity the global rules had, so authored headings render identically.

2. **Explicit styles for the app's own chrome headings.** Every heading that was previously free-riding on the global rule now styles itself:
   - Headings that had *no* self-styling and would have broken (completion "Summary of Work", activity-summary title, error / idle-warning) get full, explicit styles.
   - Headings that already set their own color/size/weight but borrowed only `line-height` keep that one property explicitly.

To avoid trading one kind of duplication for another, the shared looks are expressed as named SCSS mixins in `vars.scss` — `app-primary-heading`, `app-secondary-heading`, `app-activity-title` — so the styling is DRY and the *intent* (this is the primary page heading; this is an activity title) is named, rather than coupled implicitly by element type.

After this PR there are **no `.app h1` / `.app h2` element rules** left. Every heading's appearance is set deliberately and locally.

## Notable decision — "Teacher Feedback" heading weight

While scoping the rules I found the intro "Teacher Feedback" heading never set its own `font-weight`. As an `<h1>` (pre-AP-91) it rendered at **900**; AP-91's conversion to `<h2>` silently dropped it to the **700** it inherited from the global `.app h2`, and removing that rule would have dropped it again to normal weight. AP-120 restored this heading's color and size to be visually neutral vs. the original `<h1>` but didn't catch the weight. This PR completes that restoration — the heading now reuses `app-primary-heading` and is back to its original **900**.

## Verification — no visual change

The acceptance bar is "no visual change," so I verified it by driving the running app under Cypress and **diffing the computed `color` / `font-size` / `font-weight` / `line-height`** of each heading, baseline (AP-120) vs. this branch, across the intro page, activity page, text block, sidebar, sequence intro, completion page, and error screen — all identical. The mixin refactor was re-captured the same way and is byte-identical to the pre-refactor styling.

Two headings weren't reachable at runtime and rest on cascade analysis + compiled-CSS inspection: the "Teacher Feedback" heading (requires live teacher-feedback data to render) and authored `<h1>/<h2>` inside `.question-txt` (sample fixtures contain no authored headings).

## Stacking

This is stacked on **#570 (AP-120)** and is targeted at that branch so the diff shows only AP-122's changes. Once #570 merges to `master`, GitHub will retarget this PR to `master` automatically.


emcelroy (approved):
Nice 👍

Claude just flagged this:

**(Minor — stale comment, out-of-diff loose end) embeddable.scss:20-26**
The comment on `.embeddable-header-text` still says the question header is "caught by the global `.app h2 { color: $cc-charcoal }` rule" and that "the global element-level h1/h2 rules are legacy and slated for cleanup in a follow-up story." This PR is that follow-up — it deletes those rules. The `color: inherit` override is still genuinely needed (to keep white text on the teal bar), but the rationale now describes a rule that no longer exists. Worth updating the comment to avoid leaving a misleading breadcrumb.

kswenson:
Addressed in b5f56b4 — updated the `.embeddable-header-text` comment to drop the reference to the now-deleted global `.app h2` rule. It now explains the `color: inherit` is still needed because `.app` sets charcoal on its content, which would otherwise cascade onto the header and render it dark on the teal bar. Thanks for the catch! 🙏

PR #580, opened 2026-07-26
fix: retract the stale tutor-unavailable banner on recovery [AP-118]

dougmartin:
## Problem

The `status === "error"` effect in `use-chat.ts` set the error banner but never cleared it, so once shown it latched for the life of the page.

The visible symptom, hit while testing chat against staging: a conversation whose parent doc still carried `status:"error"` from an earlier failure showed the banner on its first snapshot after load. The backend then recovered and the tutor replied normally, but the stale "The tutor is currently unavailable. Please try again." row kept sitting directly underneath the tutor's own successful replies.

Practically, a student who hit one transient failure kept seeing "tutor is currently unavailable" for the rest of that page even while it was working. It only cleared on the next send or on page navigation, which are the two existing `setError(null)` sites.

## Fix

The effect now retracts the message when status leaves `"error"`.

It retracts **only the message it owns** (extracted as `kTutorUnavailable`). A `sendMessage` failure writes its own text and must survive an unrelated status transition, so it is deliberately left alone.

## Tests

Two new cases in `use-chat.test.ts`, covering both directions:

- the banner clears once a later drain succeeds (`error` → `generating`/`idle`)
- a `sendMessage` failure is **not** wiped by a subsequent status change

Full suite: 101 suites, 553 passed, 9 skipped. Lint clean.

## Notes

Cosmetic only, no behavior change to the chat transport, rules, or the backend. Found while verifying the chat tutor against staging as part of the 2.17.0 release.

### DIFF
--- a/src/components/chat/use-chat.test.ts
+++ b/src/components/chat/use-chat.test.ts
@@ -77,4 +77,44 @@ describe("useChat with DebugTransport", () => {
     act(() => { emitStatus("error"); });
     expect(result.current.error).toContain("unavailable");
   });
+
+  it("retracts the unavailable error once the conversation recovers", () => {
+    let emitStatus: (s: ChatStatus) => void = () => undefined;
+    const transport: ChatTransport = {
+      subscribe: (onTurns, onStatus) => { emitStatus = onStatus; onTurns([]); onStatus("idle"); return () => undefined; },
+      sendUserMessage: async () => undefined,
+    };
+    const { result } = renderHook(() => useChat({ transport, header: "h" }));
+
+    // A conversation whose parent doc still carries status:"error" from an earlier failure shows the
+    // banner on its first snapshot...
+    act(() => { emitStatus("error"); });
+    expect(result.current.error).toContain("unavailable");
+
+    // ...and must clear it once a later drain succeeds, rather than leaving a stale error sitting
+    // underneath the tutor's own replies for the life of the page.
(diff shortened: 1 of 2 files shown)


dougmartin:
Self-merging without review: this was found and fixed mid-release while verifying the 2.17.0 chat tutor against staging, not planned work.

Sequence for the record. Testing the tutor with an authenticated learner surfaced a stale "The tutor is currently unavailable. Please try again." banner sitting directly underneath the tutor's own successful replies. Root cause was the `status === "error"` effect in `use-chat.ts` setting the banner but never retracting it, so once shown it latched for the life of the page.

Verified on staging against the branch build (`4216668`) rather than just unit tests. The scenario was reproduced deliberately by setting `status: "error"` on a real conversation's parent doc in `report-service-dev`: the page loaded with the banner showing, the next drain succeeded, status flipped to `idle`, and the banner cleared. The same sequence on the master build 15 minutes earlier left the banner pinned. Also re-verified on a fresh authenticated conversation with no prior history, plus 19 forwarded interactive log docs, with no function errors.

Scope is four lines plus two tests, cosmetic only, no change to the transport, rules, or backend. Full suite green: 101 suites, 553 passed, lint clean, branch CI passed first try.

`v2.17.0` was tagged before this fix existed and was never promoted to production, so the tag will be deleted and re-cut at the new master HEAD.

